PR #4, opened 2024-08-01
[template] Update instructions

llvm-beanz:
There's been a bunch of confusion around this process, so I figured updating the instructions to be a bit more explicit might help.

### DIFF
--- a/proposals/templates/basic-template.md
+++ b/proposals/templates/basic-template.md
@@ -11,10 +11,19 @@
 > When filling out the template below for a new feature proposal, please do the
 > following first:
 
-> 1. exclude the "Planned Version", "PRs" and "Issues" from the header.
-> 2. Do not spend time writing the "Detailed design" until the feature has been
->    merged in the "Under Consideration" phase.
-> 3. Delete this Instructions section including the line below.
+> 1. Copy this template to proposals and name it `NNNN-<some unique name>.md`.
+>    Do not assign the proposal a number.
+> 2. Exclude the "Planned Version", "PRs" and "Issues" from the header as
+>    appropriate.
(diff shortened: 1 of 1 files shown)


coopp (approved):
Looks good to me!. --- incoming PR for adding a proper number to my recent proposal :)

proposals/templates/basic-template.md
```diff
@@ -11,10 +11,19 @@
...
-> 2. Do not spend time writing the "Detailed design" until the feature has been
->    merged in the "Under Consideration" phase.
-> 3. Delete this Instructions section including the line below.
+> 1. Copy this template to proposals and name it `NNNN-<some unique name>.md`.
+>    Do not assign the proposal a number.
+> 2. Exclude the "Planned Version", "PRs" and "Issues" from the header as
+>    appropriate.
+> 3. Fill out the "Introduction", "Motivation", and "Proposed Solution" sections
+>    of the template. Do not spend time writing the "Detailed design" until the
+>    feature has been merged in the "Under Consideration" phase.
```
damyanp, line 20: Note that there is nothing in this repo describing the different phases of proposals.

damyanp (approved):

proposals/templates/basic-template.md
llvm-beanz: Good point. Maybe the phases just don't matter for these so we should remove it from the template entirely.

damyanp: I think they may still be useful.  For example, the root signature proposal in here is not complete and there isn't consensus that it's the right way to go, so it'd be good if we could capture that somehow.

PR #12, opened 2024-08-05
Add metadata schema to root signature

python3kgae:
This commit adds a metadata schema to the root signature.
This schema is used to describe the root signature in llvm IR.

### DIFF
--- a/proposals/0002-root-signature-in-clang.md
+++ b/proposals/0002-root-signature-in-clang.md
@@ -554,9 +554,7 @@ The additional semantic rules not already covered by the grammar are listed here
 
 | Property       | Type              | Type detail                                         |
 | -------------- | ----------------- | --------------------------------------------------- |
-| RootFlag       | i32               | D3D12_ROOT_SIGNATURE_FLAG                           |
-| Parameters     | reference to list | list of RootDescriptor/RootConstant/DescriptorTable |
-| StaticSamplers | reference to list | list of StaticSamplers                              |
+| Parameters     | reference to list | list of  D3D12_ROOT_SIGNATURE_FLAG/RootDescriptor/RootConstant/DescriptorTable/StaticSamplers |
 
 * Function RootSignature pair
 
@@ -579,15 +577,12 @@ Example for same root signature as above:
 ; placeholder - update this once detailed design complete
 !dx.rootsignatures = !{!2}
 !2 = !{ptr @main, !3 }
-!3 = !{ i32 1, !4, !10 } ; RootFlags, Parameters, StaticSamplers
-!4 = !{!5, !6} ; list of RootDescriptor/RootConstant/DescriptorTable
-!5 = !{ !"RootCBV", i32 0, i32 1, i32 0, i32 0 } ; register 0, space 1, 0 = visiblity, 0 = flags
-!6 = !{ !"DescriptorTable", i32 0, !7 } ;  0 = visibility, range list !7
-!7 = !{ !8, !9} ; reference 2 descriptor ranges
-!8 = !{ !"SRV", i32 0, i32 0, i32 -1, i32 0 } ; register 0, space 0, unbounded, flags 0
-!9 = !{ !"UAV", i32 5, i32 1, i32 10, i32 0 } ; register 5, space 1, 10 descriptors, flags 0
-!10 = !{!11} ; list of StaticSamplers
-!11 = !{ !"StaticSampler", i32 1, i32 0, ... } ; register 1, space 0, (additional params omitted)
+!3 = !{ i32 1, !4, !5, !6, !7, !8 } ; RootFlags, Parameters, StaticSamplers
+!4 = !{ !"RootCBV", i32 0, i32 1, i32 0, i32 0 } ; register 0, space 1, 0 = visiblity, 0 = flags
+!5 = !{ !"DescriptorTable", i32 0, !7 } ;  0 = visibility, range list !7
+!6 = !{ !"SRV", i32 0, i32 0, i32 -1, i32 0 } ; register 0, space 0, unbounded, flags 0
+!7 = !{ !"UAV", i32 5, i32 1, i32 10, i32 0 } ; register 5, space 1, 10 descriptors, flags 0
+!8 = !{ !"StaticSampler", i32 1, i32 0, ... } ; register 1, space 0, (additional params omitted)
 ```
 
 ### Validations during DXIL generation
--- a/proposals/0002-root-signature-in-clang.md
+++ b/proposals/0002-root-signature-in-clang.md
@@ -554,7 +554,7 @@ The additional semantic rules not already covered by the grammar are listed here
 
 | Property       | Type              | Type detail                                         |
 | -------------- | ----------------- | --------------------------------------------------- |
-| Root elements  | reference to list | list of  D3D12_ROOT_SIGNATURE_FLAG/RootDescriptor/RootConstant/DescriptorTable/StaticSamplers |
+| Root elements  | reference to list | list of root elements|
 
 * Function RootSignature pair
 
--- a/proposals/0002-root-signature-in-clang.md
+++ b/proposals/0002-root-signature-in-clang.md
@@ -37,9 +37,10 @@ This change proposes adding:
 ## Motivation
 
 ### What are Root Signatures?
+
 In DirectX HLSL, resources can be associated with registers.  For example:
 
-``` c++
+```c++
 StructuredBuffer<float4> b1 : register(t0);
 StructuredBuffer<float4> b2 : register(t1);
 StructuredBuffer<float4> bn[] : register(t2);
@@ -67,6 +68,7 @@ cl->SetGraphicsRootDescriptorTable(2, baseDescriptor);
 ```
 
 A single parameter that's a descriptor table:
+
 ```c++
 "DescriptorTable(SRV(t0, numDescriptors = unbounded))"
 ```
@@ -250,7 +252,7 @@ to ensure that our solution doesn't unnecessarily tie the non-HLSL parts to it.
              (',' 'minLOD' '=' NUMBER)?
              (',' 'maxLOD' '=' NUMBER)? (',' 'space' '=' NUMBER)?
              (',' 'visibility' '=' SHADER_VISIBILITY)? ')'
-  
+
     bReg : 'b' NUMBER
 
     tReg : 't' NUMBER
@@ -384,7 +386,7 @@ parsedRootSignature = RootSignature{
 };
 ```
 
-### Root Signatures in the LLVM IR 
+### Root Signatures in the LLVM IR
 
 During frontend code generation an IR-based representation of the root signature
 is generated from the in-memory data structures stored in the AST. This is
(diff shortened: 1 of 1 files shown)


proposals/0002-root-signature-in-clang.md
```diff
@@ -402,23 +402,75 @@ registers are bound multiple times, or where there are multiple RootFlags
...
+
+
+* DescriptorRange
+
+| RangeType | NumDescriptors | BaseShaderRegister | RegisterSpace | DescriptorRangeFlags |
+|-----------|----------------|--------------------|---------------|----------------------|
+| String    | int            | int                | int           | int                  |
+
+
+* DescriptorTable
```
damyanp, line 428: DescriptorTable should have a ShaderVisibility field as well.

damyanp: Also, in the example it seems that DescriptorTable has two fields - the shader visiblity and the reference to the list of descriptor ranges.

proposals/0002-root-signature-in-clang.md
```diff
@@ -402,23 +402,75 @@ registers are bound multiple times, or where there are multiple RootFlags
...
+  
+| Filter | AddressU | AddressV | AddressW | MipLODBias | MaxAnisotropy | ComparisonFunc | BorderColor | MinLOD | MaxLOD | ShaderRegister | RegisterSpace | ShaderVisibility |
+|--------|----------|----------|----------|------------|---------------|----------------|-------------|--------|--------|----------------|---------------|------------------|
+| int    | int      | int      | int      | float      | int           | int            | int         | float  | float  | int            | int           | int              |
+
+* RootSignature
+
+| RootFlag | Parameters                                      | StaticSamplers          |
+|----------|-------------------------------------------------|-------------------------|
+| int      | list of RootDescriptor/RootConstant/DescriptorTable | list of StaticSamplers |
```
damyanp, line 444: This doesn't match the example.  I think really it should be a list of parameters, and each parameter could be a RootDescriptor, RootConstant, DescriptorTable or StaticSampler.  I don't think there's any requirement for the static samplers to come after the other parameters - the grammar doesn't enforce that, so I'm not sure why we would need to add that extra requirement to the metadata format?

proposals/0002-root-signature-in-clang.md
```diff
@@ -402,23 +402,75 @@ registers are bound multiple times, or where there are multiple RootFlags
...
+|----------|----------------|
+| Function | RootSignature  |
+
+* RootSignature table
+
+Named metadata with name "dx.rootsignatures".
+
+| Function RootSignature Pairs |
+|-------------------------------------|
+| list of Function RootSignature Pair |
```
damyanp, line 458: I think we should be more strict in the description of the metadata here.  I think when "list of Foo" appears, what it really means is "reference to a list of Foo", not that the list is somehow inlined into that metadata node.

proposals/0002-root-signature-in-clang.md
```diff
@@ -402,23 +402,75 @@ registers are bound multiple times, or where there are multiple RootFlags
...
+
+* DescriptorTable
+
+| DescriptorRanges          |
+|---------------------------|
+| list of DescriptorRanges  |
+
+* StaticSampler
+  
+| Filter | AddressU | AddressV | AddressW | MipLODBias | MaxAnisotropy | ComparisonFunc | BorderColor | MinLOD | MaxLOD | ShaderRegister | RegisterSpace | ShaderVisibility |
```
damyanp, line 436: I have to scroll horizontally to view this table.  What if they were presented vertically?

| FIeld | Type |
|------|------|
| Filter | int |
| AddressU | int |
etc.

proposals/0002-root-signature-in-clang.md
```diff
@@ -402,23 +402,75 @@ registers are bound multiple times, or where there are multiple RootFlags
...
+|--------|----------|----------|----------|------------|---------------|----------------|-------------|--------|--------|----------------|---------------|------------------|
+| int    | int      | int      | int      | float      | int           | int            | int         | float  | float  | int            | int           | int              |
```
damyanp, line 438: For these various "int" types we need to specify what the values mean.

Maybe something like:

| Field | Type | D3D12 Enum |
|----|----|----|
| Filter | i32 | D3D12_FILTER |
| MipLODBias | f32 | n/a |

proposals/0002-root-signature-in-clang.md
```diff
@@ -402,23 +402,75 @@ registers are bound multiple times, or where there are multiple RootFlags
...
 
+#### Metadata Schema
 
-> **Open Question**: what should the named metadata be?  There's options I
-> think...
+* RootConstant
+
+| ParameterType | ShaderVisibility | ShaderRegister | RegisterSpace | Num32BitValues |
+|---------------|------------------|----------------|---------------|----------------|
+| String        | int              | int            | int           | int            |
```
damyanp, line 411: Is `int` the right way to refer to these metadata value types?  Should it be `i32`?

damyanp (changes requested):

proposals/0002-root-signature-in-clang.md
python3kgae: Updated to i32.

proposals/0002-root-signature-in-clang.md
python3kgae: Change to `reference to list` and put Foo in Type detail.

proposals/0002-root-signature-in-clang.md
python3kgae: Updated the example.
The list will be null if no static samplers.

proposals/0002-root-signature-in-clang.md
python3kgae: Updated.

proposals/0002-root-signature-in-clang.md
```diff
@@ -402,23 +402,75 @@ registers are bound multiple times, or where there are multiple RootFlags
 entries, so subsequent stages should not assume that any given root signature in
 IR is valid.
 
+#### Metadata Schema
```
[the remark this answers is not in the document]
python3kgae: Moved.

proposals/0002-root-signature-in-clang.md
damyanp: I don't understand why these two lists are separated out.  They're not separate in the grammar, so why are we introducing that here?

If this is to match the d3d12 struct definitions, I don't think that's a goal here.

python3kgae: It started from 2 different list in the string version of root signature.
And will end to 2 different list in the final serialized root signature.
It will be strange to merge them into one list and split later.

damyanp: Note also that in the example StaticSampler starts with a parameter type, "StaticSampler", precisely so that it can be intermingled with the other parameters.

python3kgae: My mistake, I though Static Sampler have to be at end of the root signature which is not true.
If we want to keep things in the same order of the string version of root signature.
Then make a one list make sense.

proposals/0002-root-signature-in-clang.md
```diff
@@ -402,26 +404,6 @@ registers are bound multiple times, or where there are multiple RootFlags
...
-!directx.rootsignatures = !{!2}
-!2 = !{ptr @main, !3 }
-!3 = !{ !4, !5, !6, !7 } ; reference 4 root parameters
-!4 = !{ !"RootFlags", i32 1 } ; root flags, 1 is numeric value of flags
-!5 = !{ !"RootCBV", i32 0, i32 1, i32 0, i32 0 } ; register 0, space 1, 0 = visiblity, 0 = flags
-!6 = !{ !"StaticSampler", i32 1, i32 0, ... } ; register 1, space 0, (additional params omitted)
-!7 = !{ !"DescriptorTable", i32 0, !8, !9 } ;  0 = visibility, 2 ranges,!8 and !9
-!8 = !{ !"SRV", i32 0, i32 0, i32 -1, i32 0 } ; register 0, space 0, unbounded, flags 0
-!9 = !{ !"UAV", i32 5, i32 1, i32 10, i32 0 } ; register 5, space 1, 10 descriptors, flags 0
-```
```
damyanp, line 422: This example should stay here since it's part of proposed solutions.

python3kgae: Updated.

proposals/0002-root-signature-in-clang.md
```diff
@@ -554,9 +554,7 @@ The additional semantic rules not already covered by the grammar are listed here
 
 | Property       | Type              | Type detail                                         |
 | -------------- | ----------------- | --------------------------------------------------- |
-| RootFlag       | i32               | D3D12_ROOT_SIGNATURE_FLAG                           |
-| Parameters     | reference to list | list of RootDescriptor/RootConstant/DescriptorTable |
-| StaticSamplers | reference to list | list of StaticSamplers                              |
+| Parameters     | reference to list | list of  D3D12_ROOT_SIGNATURE_FLAG/RootDescriptor/RootConstant/DescriptorTable/StaticSamplers |
```
damyanp, line 557: The grammar refers to these collectively as "Root elements".  We may as well just use the same terminology in the metadata format.

python3kgae: Updated.

proposals/0002-root-signature-in-clang.md
damyanp: It looks like this section is still in the detailed design part rather than the proposed solution.

proposals/0002-root-signature-in-clang.md
```diff
@@ -489,29 +473,122 @@ The additional semantic rules not already covered by the grammar are listed here
...
+| ---------------- | ----------------- | ------------------------ |
+| ParameterType    | string            | "DescriptorTable"        |
+| ShaderVisibility | i32               | D3D12_SHADER_VISIBILITY  |
+| DescriptorRanges | Reference of list | list of DescriptorRanges |
+
+* StaticSampler
+
+| Property         | Type  | Type detail                |
+| ---------------- | ----- | -------------------------- |
+| Filter           | i32   | D3D12_FILTER               |
```
damyanp, line 539: This needs a ParameterType / string / "StaticSampler" row.

proposals/0002-root-signature-in-clang.md
```diff
@@ -489,29 +473,122 @@ The additional semantic rules not already covered by the grammar are listed here
...
+| MaxLOD           | float | float                      |
+| ShaderRegister   | i32   | i32                        |
+| RegisterSpace    | i32   | i32                        |
+| ShaderVisibility | i32   | D3D12_SHADER_VISIBILITY    |
+
+* RootSignature
+
+| Property       | Type              | Type detail                                         |
+| -------------- | ----------------- | --------------------------------------------------- |
+| Root elements  | reference to list | list of  D3D12_ROOT_SIGNATURE_FLAG/RootDescriptor/RootConstant/DescriptorTable/StaticSamplers |
```
damyanp, line 557:
```suggestion
| Root elements  | reference to list | list of root elements|
```

damyanp (approved):

tex3d (approved):

PR #35, opened 2024-08-13
Add hlslpm, a tool for curating issues

damyanp:
This change adds tools/hlslpm, a tool that helps to curate the wg-hlsl milestone and workstream issues.

Also adds docs/issue_tracking.md that describes how we're using issues to track the work required for HLSL Support.


docs/issue_tracking.md
```diff
@@ -0,0 +1,80 @@
...
+field being set to "Project Milestone".
+
+The current target date for hitting a milestone is in the "Target" field.
+
+The current list of milestones can be seen on the [Milestones
+view](https://github.com/orgs/llvm/projects/4/views/15). The [Roadmap
+view](https://github.com/orgs/llvm/projects/4/views/12) also provides a
+graphical view of the milestones over time.
+
+The main body for a Milestone issue is follows this format:
```
coopp, line 33: This 'is' in the sentance reads weird...  Could be just me and not having my morning caffine yet.

docs/issue_tracking.md
```diff
@@ -0,0 +1,80 @@
...
+
+* Preamble - text providing information about the milestone, what the goals are,
+  what the challenges might be, why it is an interesting milestone etc.
+* Workstreams - a list of workstreams, and the workstream milestones that
```
coopp, line 37: .. a list of workstreams, and the workstream milestones that __ contribute __ to this one..  ?
or
.. a list of workstreams, and the workstream milestones that __ contributes __ to this one... ?

docs/issue_tracking.md
dmpots: Should we create a template for milestone issues and maybe others? Sounds like several of the issue types have recommended formatting that would be easier to enforce with a template.

docs/issue_tracking.md
```diff
@@ -0,0 +1,80 @@
...
+The Milestones section starts with the header `## Milestones`. This section is the authority for what gets aggregated into the milestone issues.
+
+
+## Implementation Tasks / Bug Tracking
+
+Issues that track tasks or bug fixes that involve changing code in Clang/LLVM
+are tracked in the [llvm-project](https://github.com/llvm/llvm-project/issues)
+repo.
+
+These labels identify track issues related to HLSL support in Clang:
```
dmpots, line 70: Awkward wording "These labels identify track issues"

coopp (approved):
Looks good to me.

tools/hlslpm/README.md
```diff
@@ -0,0 +1,62 @@
+<!-- {% raw %} -->
+
+# hlslpm
+
+This tool helps manage github issues tracking the [LLVM HLSL
```
dmpots, line 5: Are we expecting to run this tool as an automated process to keep issues updated or will this be something we run manually?

tools/hlslpm/github.py
```diff
@@ -0,0 +1,144 @@
+from enum import Enum
```
dmpots, line 1: I find it helpful to have a file-level comment that describes the contents of the script at a high level. Even something as simple as "Contains a class that can be used to access github issues and associated unit tests" is a nice way to quickly understand what we should expect to see.

docs/issue_tracking.md
damyanp: Went with _contribute_.

tools/hlslpm/github.py
```diff
@@ -0,0 +1,144 @@
...
+
+    def project_items_summary(self):
+        query = read_file("gql/project_item_summary.gql")
+
+        pageInfo = {'endCursor': None, 'hasNextPage': True}
+
+        while pageInfo['hasNextPage']:
+            response = self.graphql(query, {'after': pageInfo['endCursor']})
+
+            items = maybe_get(response, 'data', 'organization',
```
dmpots, line 35: `maybe_get` looks like it could return `None`, but the code below assumes it is a dictionary.

tools/hlslpm/github.py
```diff
@@ -0,0 +1,144 @@
...
+                        node, 'content', 'resourcePath'),
+                    issue_updatedAt=to_datetime(maybe_get(node, 'content', 'updatedAt')))
+
+            pageInfo = maybe_get(items, 'pageInfo')
+
+    def get_issues(self, issues: List[Issue]):
+        query = read_file("gql/get_issue_text.gql")
+        chunk_size = 50
+
+        issue_id_chunks = [issues[i:i+chunk_size]
```
dmpots, line 55: I was thinking that `itertools` should have something for this. Looks like it does, but it was added in python 3.12

https://docs.python.org/3/library/itertools.html#itertools.batched

tools/hlslpm/github.py
```diff
@@ -0,0 +1,144 @@
...
+                        maybe_get(node, 'category', 'name')),
+                    target_date=to_date(maybe_get(node, 'target', 'date')),
+                    issue_id=maybe_get(node, 'content', 'id'),
+                    issue_resourcePath=maybe_get(
+                        node, 'content', 'resourcePath'),
+                    issue_updatedAt=to_datetime(maybe_get(node, 'content', 'updatedAt')))
+
+            pageInfo = maybe_get(items, 'pageInfo')
+
+    def get_issues(self, issues: List[Issue]):
```
dmpots, line 51: I like the use of type annotations here. I think it really helps with documentation of the function. I think it would also be useful to add the return types (using the `->` syntax) for these functions with complex results.

tools/hlslpm/github.py
```diff
@@ -0,0 +1,144 @@
...
+                issue.title = node["title"]
+                issue.body = node["body"]
+                yield issue
+
+    def set_issue_body(self, id, body):
+        query = read_file("gql/set_issue_body.gql")
+        self.graphql(query, {"id": id, "body": body})
+
+
+def maybe_get(d: dict, *args):
```
dmpots, line 73: I think this function could use a comment describing what it does and intended usage.

tools/hlslpm/github.py
```diff
@@ -0,0 +1,144 @@
...
+    return datetime.strptime(str, '%Y-%m-%d')
+
+
+def get_pat():
+    try:
+        return read_file("pat.txt").strip()
+    except:
+        print(
+            f"Couldn't read pat.txt not found - create a pat on github and store it in pat.txt next to {__file__}.")
+        sys.exit(1)
```
dmpots, line 100: `raise`ing an exception here could make the script more friendly for testing and embedding.

tools/hlslpm/github.py
```diff
@@ -0,0 +1,144 @@
...
+
+
+def read_file(filename):
```
dmpots, line 103: Naming this something like `read_relative_file` would make the intent more clear.

tools/hlslpm/github.py
```diff
@@ -0,0 +1,144 @@
...
+    with open(os.path.join(os.path.dirname(__file__), filename), 'r') as f:
+        return f.read()
+
+
+if __name__ == '__main__':
+    gh = GH()
+
+    if False:
```
dmpots, line 111: These code blocks looks like something that could potentially be turned into tests?

tools/hlslpm/issue.py
```diff
@@ -0,0 +1,347 @@
...
+        for line in section.contents:
+            body.append(line)
+            
+        if body[-1] != "":
+            body.append("")
+
+    return body
+
+class Issues:
+    all_issues: Dict[str, Issue]
```
dmpots, line 321: I'm having trouble understanding why a dictionary is used to hold the issues instead of a list. It looks like the key is `i.issue_resourcePath`, but that seems like it could just be part of the `Issue`? I'm sure there is a good reason, but I think adding some explanation about the data structure here would help.

tools/hlslpm/hlslpm.py
```diff
@@ -0,0 +1,63 @@
...
+    if args.save and args.update:
+        print("Only one of --save or --update can be specified")
+    
+
+    print("Fetching data...")
+    gh = github.GH()
+    issues = Issues(gh)
+
+
+    tracked = list(issues.milestones.values()) + list(issues.workstreams.values())
```
dmpots, line 35: I see a lot of accesses to the raw dict inside of Issues. Might be good to hide that a bit with an accessor (something like `issues.get_milestone_list()`)_ that returns a `List[Issue]` since that seems to be the most common way to access it. Protects a bit in case we want to change the underlying data structure.

tools/hlslpm/issue.py
```diff
@@ -0,0 +1,347 @@
...
+    contents: List[str]
+
+    def __init__(self):
+        self.contents = []    
+
+    def getReferenceFromTitle(self):
+        if not self.title:
+            return None
+
+        m = re.match(".*\((.+)\)", self.title)
```
dmpots, line 28: I think [named groups](https://docs.python.org/3/library/re.html#re.Match.group) make the regex more readable.

dmpots (approved):
Overall, LGTM.

docs/issue_tracking.md
damyanp: Something to consider as we go forward - I'm not sure how stable the format is at the moment while we're still fleshing out the first few workstreams.

docs/issue_tracking.md
damyanp: More just wrong I think :)  Removed "identify".

tools/hlslpm/README.md
damyanp: At the moment I'm running it manually and checking what it'll do. Will consider automating it later if it proves to be stable.  As an example, the first time I ran it I had to fix up a bunch of things because some changes were made to a milestone issue instead of a workstream issue and they'd have just been overwritten.

tools/hlslpm/github.py
damyanp: Yup, but in this case I'd be happy with the error coming from attempting to use the dictionary - this would be a bug in the code rather than an environmental issue.

tools/hlslpm/github.py
damyanp: That's good to know...if I find myself modifying this part I might try updating it.

tools/hlslpm/github.py
damyanp: Added some.

tools/hlslpm/github.py
damyanp: Added

tools/hlslpm/github.py
damyanp: As it is, if something needs the PAT and it isn't there it really cannot continue and so I wanted the script to exit with the error message generated close to that point.

tools/hlslpm/github.py
damyanp: renamed

tools/hlslpm/github.py
damyanp: Not really - they access network resources and I don't want to have tests that require specific data to be available.  These are more for local experimentation.

tools/hlslpm/issue.py
damyanp: In the case of all_issues it is because we need to be able to look them up by resourcePath, which I'd argue is self documenting it by being a Dict.

milestones and workstreams I don't think need to be a dict.

tools/hlslpm/hlslpm.py
damyanp: These are lists now.  I'm trying to avoid too much abstraction and boilerplate code in this.

tools/hlslpm/issue.py
damyanp: I agree in some cases, but I don't think any of these regexes quite get there - named groups also clutter the regex quite a bit.

tools/hlslpm/github.py
damyanp: Added some.

llvm-beanz (approved):

PR #38, opened 2024-08-16
A proposal for DXIL Scalarization

farzonl:
This covers:
- the motivations, 
- background of how Scalarization is working in DXC 
- the approach to tasking we should take.

### DIFF
--- a/proposals/NNNN-DXIL-Scalarization.md
+++ b/proposals/NNNN-DXIL-Scalarization.md
@@ -1,0 +1,218 @@
+# DXIL Scalarization
+
+* Proposal: [NNNN](NNNN-DXIL-Scalarization.md)
+* Author(s): [Farzon Lotfi](https://github.com/farzonl)
+* Sponsor: [Farzon Lotfi](https://github.com/farzonl)
+* Status: **Under Consideration**
+* Impacted Projects: Clang
+
+## Introduction
+As mentioned in [DXIL.rst](https://github.com/microsoft/DirectXShaderCompiler/blob/main/docs/DXIL.rst#vectors) 
+"HLSL vectors are scalarized" and "Matrices are lowered to vectors". Therefore,
+ we need to be able to scalarize these types in clang DirectX backend. Today 
+this is done via [`SROA_Parameter_HLSL` Module pass in `ScalarReplAggregatesHLSL.cpp`](https://github.com/microsoft/DirectXShaderCompiler/blob/main/lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp#L4263).
+
+
+## Motivation
+Without Scalarizing the data structures and call instructions we can't generate valid DXIL.
+
+## Background
+In DXC the scalarization pass looks  something like this:
+```
+SROA_Parameter_HLSL::runOnModule 
+|-> LegalizeDxilInputOutputs
+|-> RewriteBitcastWithIdenticalStructs
+|-> createFlattenedFunction  
+    |-> flattenArgument -> DoScalarReplacement
+    |-> special matrix load stores
+    |-> LegalizeDxilInputOutputs
+|-> SROAGlobalAndAllocas -> DoScalarReplacement -> RewriteForScalarRepl
+
+
+ _________________________________________RewriteForScalarRepl_________________________________________________              <------------------|
+/                    /              /              |               \               \            |              \                                |
+RewriteForConstExpr  RewriteForGEP  RewriteForLoad RewriteForStore RewriteMemIntrin RewriteCall RewriteBitCast  RewriteForAddrSpaceCast         |
+                                                                                    /          |___                                   |_______  |
+                                                                                    RewriteCallArg |-> RewriteWithFlattenedHLIntrinsicCall    |-|
+```
+
+### RewriteForConstExpr
+Recursively "flatten" a constant expression (ConstantExpr) in LLVM IR. 
+It checks if the expression is either a GEPOperator (GEP instruction) or an AddrSpaceCast
+If the expression is used by an instruction, it converts the ConstantExpr 
+into an instruction and replaces the use with this new instruction.
+
+### RewriteForGEP
+ A function designed to "rewrite" a GetElementPtr (GEP) instruction to make it 
+ relative to a new element, usually when dealing with structs, vectors, or arrays.
+ for the struct  case when a matching element is found, it simplifies the GEP 
+ by updating its base pointer and indices. If no such element exists, new GEPs 
+ are created for each element in an array or vector and replaces the old GEP 
+ accordingly. The function ensures that the new GEP has the correct type &
+ replaces all uses of the original GEP. After rewriting, the old GEP is either
+ destroyed or marked for deletion. The goal is to to break down and flatten
+ complex memory access patterns in IR.
+
+ ### RewriteForLoad
+ Three cases:
+ 1. // Replace for Vector:
+```
+    //   %res = load { 2 x i32 }* %alloc
+    // with:
+    //   %load.0 = load i32* %alloc.0
+    //   %insert.0 insertvalue { 2 x i32 } zeroinitializer, i32 %load.0, 0
+    //   %load.1 = load i32* %alloc.1
+    //   %insert = insertvalue { 2 x i32 } %insert.0, i32 %load.1, 1
+```
+ 2. // Replace for Agregate Vector\Struct Array:
+```
+      //   %res = load [2 x <2 x float>] * %alloc
+      // with:
+      //   %load.0 = load [4 x float]* %alloc.0
+      //   %insert.0 insertvalue [4 x float] zeroinitializer,i32 %load.0,0
+      //   %load.1 = load [4 x float]* %alloc.1
+      //   %insert = insertvalue [4 x float] %insert.0, i32 %load.1, 1
+      //  ...
+```
+3. // Replace for Aggregate loads:
+```
+      //   %res = load { i32, i32 }* %alloc
+      // with:
+      //   %load.0 = load i32* %alloc.0
+      //   %insert.0 insertvalue { i32, i32 } zeroinitializer, i32 %load.0,
+      //   0
+      //   %load.1 = load i32* %alloc.1
+      //   %insert = insertvalue { i32, i32 } %insert.0, i32 %load.1, 1
+      // (Also works for arrays instead of structs)
+```
+### RewriteForStore
+1. // Replace for Vector:
+```
+    //   store <2 x float> %val, <2 x float>* %alloc
+    // with:
+    //   %val.0 = extractelement { 2 x float } %val, 0
+    //   store i32 %val.0, i32* %alloc.0
+    //   %val.1 = extractelement { 2 x float } %val, 1
+    //   store i32 %val.1, i32* %alloc.1
+```
+2. // Replace for Agregate Vector\Struct Array:
+```
+      //   store [2 x <2 x i32>] %val, [2 x <2 x i32>]* %alloc, align 16
+      // with:
+      //   %val.0 = extractvalue [2 x <2 x i32>] %val, 0
+      //   %all0c.0.0 = getelementptr inbounds [2 x i32], [2 x i32]* %alloc.0,
+      //   i32 0, i32 0
+      //   %val.0.0 = extractelement <2 x i32> %243, i64 0
+      //   store i32 %val.0.0, i32* %all0c.0.0
+      //   %alloc.1.0 = getelementptr inbounds [2 x i32], [2 x i32]* %alloc.1,
+      //   i32 0, i32 0
+      //   %val.0.1 = extractelement <2 x i32> %243, i64 1
+      //   store i32 %val.0.1, i32* %alloc.1.0
+      //   %val.1 = extractvalue [2 x <2 x i32>] %val, 1
+      //   %alloc.0.0 = getelementptr inbounds [2 x i32], [2 x i32]* %alloc.0,
+      //   i32 0, i32 1
+      //   %val.1.0 = extractelement <2 x i32> %248, i64 0
+      //   store i32 %val.1.0, i32* %alloc.0.0
+      //   %all0c.1.1 = getelementptr inbounds [2 x i32], [2 x i32]* %alloc.1,
+      //   i32 0, i32 1
+      //   %val.1.1 = extractelement <2 x i32> %248, i64 1
+      //   store i32 %val.1.1, i32* %all0c.1.1
+```
+3. // Replace for Aggregate stores:
+```
+      //   store { i32, i32 } %val, { i32, i32 }* %alloc
+      // with:
+      //   %val.0 = extractvalue { i32, i32 } %val, 0
+      //   store i32 %val.0, i32* %alloc.0
+      //   %val.1 = extractvalue { i32, i32 } %val, 1
+      //   store i32 %val.1, i32* %alloc.1
+      // (Also works for arrays instead of structs)
+```
+
+### RewriteMemIntrin
+rewriting memory intrinsics (memcpy, memset, memmove) when dealing with scalarized memory. 
+The goal is to decompose the memory intrinsic operation into element-wise operations that 
+apply to each scalar element of the memory being accessed.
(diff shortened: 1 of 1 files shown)


proposals/NNNN-DXIL-Scalarization.md
```diff
@@ -0,0 +1,218 @@
...
+* Author(s): [Farzon Lotfi](https://github.com/farzonl)
+* Sponsor: [Farzon Lotfi](https://github.com/farzonl)
+* Status: **Under Consideration**
+* Impacted Projects: Clang
+
+## Introduction
+As mentioned in [DXIL.rst](https://github.com/microsoft/DirectXShaderCompiler/blob/main/docs/DXIL.rst#vectors) 
+"HLSL vectors are scalarized" and "Matrices are lowered to vectors". Therefore,
+ we need to be able to scalarize these types in clang DirectX backend. Today 
+this is done via [`SROA_Parameter_HLSL` Module pass in `ScalarReplAggregatesHLSL.cpp`](https://github.com/microsoft/DirectXShaderCompiler/blob/main/lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp#L4263).
```
damyanp, line 13: This is a link to a source file in DXC.  This is a proposal for clang, so it isn't a link to how it is done today.  Maybe more accurate to say "In DXC, this is done via ... " and then caveat saying that we need to design a solution appropriate for clang.

proposals/NNNN-DXIL-Scalarization.md
```diff
@@ -0,0 +1,218 @@
...
+```
+
+### RewriteForConstExpr
+Recursively "flatten" a constant expression (ConstantExpr) in LLVM IR. 
+It checks if the expression is either a GEPOperator (GEP instruction) or an AddrSpaceCast
+If the expression is used by an instruction, it converts the ConstantExpr 
+into an instruction and replaces the use with this new instruction.
+
+### RewriteForGEP
+ A function designed to "rewrite" a GetElementPtr (GEP) instruction to make it
```
damyanp, line 46: nit: this section seems to be indented by one space for some reason.

damyanp:
I'm uneasy with the way that this is framed as a "port some code from DXC to Clang".  I'm not convinced that this is the right approach to take.  DXC's implementation can be used to gain some inspiration for the sorts of things that might need to be done, but at the end of the day we need to design the best implementation for Clang that uses up-to-date LLVM IR and is built upon the latest clang/llvm code.

llvm-beanz (changes requested):
I don't think this is the correct approach for finding the right solution. Fundamentally DXC and Clang are going to _wildly_ diverge early in the compiler, so the intermediate optimization and legalization needs aren't going to be consistent.

I think this is an area where we need to take a first principles approach and first define what scalarization is required for valid DXIL, then we can define how we get there in LLVM.

farzonl:
> I'm uneasy with the way that this is framed as a "port some code from DXC to Clang". I'm not convinced that this is the right approach to take. DXC's implementation can be used to gain some inspiration for the sorts of things that might need to be done, but at the end of the day we need to design the best implementation for Clang that uses up-to-date LLVM IR and is built upon the latest clang/llvm code.

The intent isn't so much a port but to indicate scalarization needs to handle `CallInst`, `LoadInst`, `StoreInst`, `GetElementPtrInst`, `AddrSpaceCastInst`, `BitCastInst`,  and `MemIntrinsic`.

damyanp:
> The intent isn't so much a port...

The spec explicitly says: "The port should start with a compatible alternative to `RewriteCall`...."  :)

> but to indicate scalarization needs to handle `CallInst`, `LoadInst`, `StoreInst`, `GetElementPtrInst`, `AddrSpaceCastInst`, `BitCastInst`, and `MemIntrinsic`.

Ok - I didn't get this from the proposal at all :)

I'd suggest moving the details about what DXC does to an appendix at the end and lead with the motivation/background explaining what needs to be do with the things in this list and why, and the proposed solution explaining how you plan to go about doing it.

farzonl:
> I don't think this is the correct approach for finding the right solution. Fundamentally DXC and Clang are going to _wildly_ diverge early in the compiler, so the intermediate optimization and legalization needs aren't going to be consistent.
> 
> I think this is an area where we need to take a first principles approach and first define what scalarization is required for valid DXIL, then we can define how we get there in LLVM.

I think the background on how DXC works here might be overshadowing the proposal. So i'm going to remove the background section. What I wanted to get across is better represented in the tickets section. We need to support function arguments and globals and  when walking the module handle the following instructions:  `CallInst`, `LoadInst`, `StoreInst`, `GetElementPtrInst`, `AddrSpaceCastInst`, `BitCastInst`,  and `MemIntrinsic`.

llvm-beanz:
Do we need to support function arguments generally or just arguments to DXIL Ops (I believe the later)?

I'm 99% sure any occurrences of `AddrSpaceCastInst` are caused by DXC generating invalid IR which it then "fixes" by performing unsafe transformations on the IR.

I think some amount of these cases are handled by LLVM's SROA pass today, so any proposal for a solution should take into account the modern state of LLVM rather than the state of DXC.

farzonl:
> DXIL Ops

Just DXIL Ops, however there two different scalarization behaviors between  `RewriteCallArg` and `RewriteWithFlattenedHLIntrinsicCall` depending on the intrinsic.

> AddrSpaceCastInst

Yes doesnt look like LLVM SROA does anything for this so I call that out now in my proposal as something we won't add and removed it from the tickets.

> some amount of these cases are handled by LLVM's SROA

Yeah similarly there seems to be a pattern SROA and `SPIRVEmitIntrinsics` are using that we could do here.

llvm-beanz:
> > DXIL Ops
> 
> Just DXIL Ops, however there two different scalarization behaviors between `RewriteCallArg` and `RewriteWithFlattenedHLIntrinsicCall` depending on the intrinsic.

LLVM doesn't have "HLIntrinsics" in any form that remotely resembles what they did in DXC, so we can probably ignore that. DXIL op rewriting seems like something we should do late.

> > AddrSpaceCastInst
> 
> Yes doesnt look like LLVM SROA does anything for this so I call that out now in my proposal as something we won't add and removed it from the tickets.

Rewriting address space casts is generally illegal, because it means you're doing something bonkers.

proposals/NNNN-DXIL-Scalarization.md
```diff
@@ -0,0 +1,239 @@
...
+
+The behaviors of `RewriteForGEP`, `RewriteForAddrSpaceCast`, `RewriteForConstExpr`,
+and `RewriteMemIntrin`. Should be things we can get  via existing LLVM passes.
+
+## Areas explicitly ignored
+SROA_Parameter pass has some  matrix loads and stores that this proposal won't go into.
+It also looks to do some legalization via `LegalizeDxilInputOutputs` that won't be covered.
+
+## Tickets
+1. Create a scalarization pass that just handles the flattenArgument case to start.
```
llvm-beanz, line 59: I'd like to understand this more. Do you have an example before and after IR for a rewritten call instruction?

proposals/NNNN-DXIL-Scalarization.md
```diff
@@ -0,0 +1,239 @@
...
+It also looks to do some legalization via `LegalizeDxilInputOutputs` that won't be covered.
+
+## Tickets
+1. Create a scalarization pass that just handles the flattenArgument case to start.      
+    - To simplify things "workists" will just be HLSL intrinsics.
+    - We iterate over all intrinsics and iterate over all value uses
+    - For now the only value we will check for are `CallInst`
+    - The pass will replace intrinsics used by DXIL.td that are a multi-element 
+    call with multiple scalarized calls
+2. Expand the scalar expansion pass to handle `loads` by checking for `LoadInst`
```
llvm-beanz, line 65: How does this align with the resources plan (@bogner)? Are we going to have custom load instructions early in the compiler or late?

farzonl:
> LLVM doesn't have "HLIntrinsics" in any form that remotely resembles what they did in DXC, so we can probably ignore that. DXIL op rewriting seems like something we should do late.

Lets not get caught up on the names, i'm referring to behaviors. The behavior is to  scalarize `CallInst` that map to DXIL Ops.  Thats a pretty easy pass for us to write.

> seems like something we should do late.
late/early is something we can figure out later, the point is the pass needs to be written and after it is, it should be easy to move it around.

proposals/NNNN-DXIL-Scalarization.md
farzonl: Yeah I have two approaches one thats easy to move and one thats less easy to move around. In both cases I need some indication that we want to scalarize. I was looking into how to use the `LowerMap` in `DXILOperation.inc`. But even then I would need a flag because some intrinsics like dot2-4  shouldn't be scalarized.   `DXIL.td` will have a bool  we set on whether to scalarize or not scalarize an intrinsic.

- case 1 the pass we can't move
We modify  `lowerIntrinsic` in `llvm/lib/Target/DirectX/DXILOpLowering.cpp` to check for this flag set or not then directly replace an intrinsics with a set of DXILOps.  In this case we do scalarization as part of DXIL lowering. Advantage to doing it this way is we are limiting the intrinsics to just those that map to DXIL Ops which means we don't have to check intrinsic scalarization across all CallInst.

- case 2 we replace 1 intrinsic with more intrinsics
Doing it this way we can put this pass any where before DXIL Lowering. And the function to do this is very simple:
```cpp
static bool scalarizeIntrinsic(Function &F, CallInst *OrigIntrinsic) {
  Instruction *CurrInstruction = OrigIntrinsic;
  IRBuilder<> Builder(CurrInstruction);

  size_t NumOperands = OrigIntrinsic->getNumOperands();
  Value *Arg0 = OrigIntrinsic->getOperand(0);
  auto *VecArg = dyn_cast<FixedVectorType>(Arg0->getType());
  Value *Result = llvm::PoisonValue::get(llvm::FixedVectorType::get(
      VecArg->getElementType(), VecArg->getNumElements()));
  for (size_t i = 0; i < VecArg->getNumElements(); i++) {
    SmallVector<Value *, 4> ExtractedElements;
    for (size_t j = 0; j < NumOperands; j++) {
      Value *CurrArg = OrigIntrinsic->getOperand(j);
      if (CurrArg->getType()->isPointerTy())
        continue;
      Value *ArgIndex =
          ConstantInt::get(Type::getInt32Ty(CurrArg->getContext()), i);
      Value *ExtractedElement = Builder.CreateExtractElement(CurrArg, ArgIndex);
      ExtractedElements.push_back(ExtractedElement);
    }
    ArrayRef<Value *> ArgsAsArrayRef(ExtractedElements.data(),
                                     ExtractedElements.size());
    CallInst *CurrCall = Builder.CreateIntrinsic(
        /*ReturnType*/ VecArg->getElementType(), F.getIntrinsicID(),
        ArgsAsArrayRef);
    Result = Builder.CreateInsertElement(Result, CurrCall, i);
  }

  OrigIntrinsic->replaceAllUsesWith(Result);
  OrigIntrinsic->eraseFromParent();
  return true;
}
```

llvm-beanz:
> Lets not get caught up on the names, i'm referring to behaviors. The behavior is to scalarize `CallInst` that map to DXIL Ops. Thats a pretty easy pass for us to write.

If you don't want to get hung up on names you should be careful which ones you use and how much description you provide. If you say:

> A good starting place of behaviors to try to match is `RewriteCall`  specifically `RewriteWithFlattenedHLIntrinsicCall`As the other pieces are not needed  for compute shaders.

(which the doc says), I'm going to get real curious why we need to flatten HLIntrinsic calls since we don't have HLIntrinsics.

> I think it makes the most sense to do this operation when we still have an intrinsic and not wait until after we have lowered to a DXIL op. Thats because the intrinsic is easy to key on. The latest we could is as part of the DXIL lowering pass. But i would ideally want to make this its own pass.

Why _shouldn't_ we do this as part of the DXIL lowering pass? That seems like the logical place to me. The DXIL lowering pass consumes calls to intrinsics and rewrites them to DXIL Ops scalarizing them if needed. That seems reasonable to me.

farzonl:
> Why _shouldn't_ we do this as part of the DXIL lowering pass?

Its easy enough to do it there. My concern was that we lose flexibility if say we find a pass should run before or after. I.e it makes the discussion about late\early transformation something we can change very easily.

proposals/NNNN-DXIL-Scalarization.md
farzonl: - example for  lowerintrinsic
```llvm
; CHECK:extractelement <{{.*}} x float> %{{.*}}, i32 {{.*}}
; CHECK:call float @dx.op.unary.f32(i32 13, float %{{.*}})
; CHECK:insertelement <{{.*}} x float>
; CHECK:extractelement <{{.*}} x float> %{{.*}}, i32 {{.*}}
; CHECK:call float @dx.op.unary.f32(i32 13, float %{{.*}})
; CHECK:insertelement <{{.*}} x float>
 %elt.sin = call <2 x float> @llvm.sin.v2f32(<2 x float> %0)
```
-example as seperate pass
```llvm
; CHECK:extractelement <{{.*}} x float> %{{.*}}, i32 {{.*}}
; CHECK:call float @llvm.sin.f32(float %{{.*}})
; CHECK:insertelement <{{.*}} x float>
; CHECK:extractelement <{{.*}} x float> %{{.*}}, i32 {{.*}}
; CHECK:call float @llvm.sin.f32(float %{{.*}})
; CHECK:insertelement <{{.*}} x float>
 %elt.sin = call <2 x float> @llvm.sin.v2f32(<2 x float> %0)
```

llvm-beanz: Oaky, that's fine for an element wise operation, but what about something like `dot`?

llvm-beanz: I don't think I understand why we need the "flexibility" to move this around. It seems to me that we're unlikely to want to run a lot of optimizations between scalarizing intrinsic calls and generating DXIL since the point of preserving the vectors is to allow LLVM to optimize more effectively.

farzonl: > Oaky, that's fine for an element wise operation, but what about something like `dot`?

We did dot already:
https://github.com/llvm/llvm-project/blob/ea8bb4d633683f5cbfd82491620be3056f347a02/llvm/lib/Target/DirectX/DXILOpLowering.cpp#L86-L89
In this PR:
https://github.com/llvm/llvm-project/pull/86071/files#diff-abc4769e41f4d9022e3b88d8d752267e5d1bb26c0d4c4186f5d551c0dce76b80

farzonl: > I don't think I understand why we need the "flexibility" to move this around. It seems to me that we're unlikely to want to run a lot of optimizations between scalarizing intrinsic calls and generating DXIL since the point of preserving the vectors is to allow LLVM to optimize more effectively.

Maybe we don't I'll confirm the use case i'm worried about with you offline.

proposals/NNNN-DXIL-Scalarization.md
```diff
@@ -0,0 +1,655 @@
...
+Are largely handled for us if we use the `scalarizer` pass with the `-scalarize-load-store` and `-scalarize-variable-insert-extract` flags. And we probably don't need to handle the `AddrSpaceCastInst` case. Allocas could be handled by running SROA. So thats 6/8 cases
+just by using an existing pass. 
+
+There are a few ways we could go about this:
+1. support `PassInfoMixin` in the DirectX backend. 
+   - Need to override "buildCodeGenPipeline" for the DirectX backend.
+2. Add a Legacy Pass manager abstraction to `scalarizer`
+   - Do we need permission to do this? Do we need an RFC?
+3. Fork the `scalarizer` so we can have a legacy and `PassInfoMixin` in the DirectX Backend?
+   - We could make this work for dxil ops and not just intrinsics.
```
bogner, line 389: Couldn't we also just run the pass before the backend? Is that too early?

farzonl: It needs to run after `DXILInstructionExpansion`  which is in the backend and before `DXILLowering`. So we are currently stuck in the backend unless we move the instruction Expansion.

PR #42, opened 2024-08-19
[0006] Propose the set of resource attributes that are needed to describe texture resources

bob80905:
This PR proposes a set of attributes that will replace `HLSLResourceAttr`. This set will be sufficient to distinguish between any texture type, and a table is proposed below that defines which attributes should be present and what values the attributes should have for each texture kind. Certain buffers are also added for extra clarity.

### DIFF
--- a/proposals/0006-resource-attributes.md
+++ b/proposals/0006-resource-attributes.md
@@ -39,6 +39,11 @@ the attributes we would expect to see on the type that defines these
 texture variables. We should expect that given any combination of
 attributes and values of the attributes, we can directly infer what
 the texture type is and how to construct the associated dxil handle.
+For example, a `RWTexture2DMSArray` texture would have a definition
+in an HLSL header, and the type itself would be constructed with
+specific attributes attached to it: `hlsl::resource_class` would be 
+`SRV`, `hlsl::is_rov` would not be present, `hlsl::texture_dimension`
+would be `2DArray`, and `hlsl::texture_type` would be `MS`.
 
 | HLSL Texture Type              | hlsl::resource_class | hlsl::is_rov | hlsl::texture_dimension | hlsl::texture_type |
 | ------------------------------- | -------------------- | ------------ | ----------------------- | ------------------ |
--- a/proposals/0006-resource-attributes.md
+++ b/proposals/0006-resource-attributes.md
@@ -70,9 +70,9 @@ would be `2DArray`, and `hlsl::texture_type` would be `MS`.
 | RWTexture2DMSArray              | UAV                  | -            | 2                       | -             | yes         | -                 | yes            |
 | FeedbackTexture2D               | SRV                  | -            | 2                       | -             | -           | yes               | -              |
 | FeedbackTexture2DArray          | SRV                  | -            | 2                       | -             | -           | yes               | yes            |
-| TypedBuffer                     | SRV                  | -            | -                       | -             | -           | yes               | yes            |
-| RawBuffer                       | SRV                  | -            | -                       | -             | -           | yes               | yes            |
-| StructuredBuffer                | SRV                  | -            | -                       | -             | -           | yes               | yes            |
+| TypedBuffer                     | SRV                  | -            | -                       | -             | -           | -                 | -              |
+| RawBuffer                       | SRV                  | -            | -                       | -             | -           | -                 | -              |
+| StructuredBuffer                | SRV                  | -            | -                       | -             | -           | -                 | -              |
 
 
 ## Detailed design
--- a/proposals/0006-resource-attributes.md
+++ b/proposals/0006-resource-attributes.md
@@ -45,31 +45,35 @@ specific attributes attached to it: `hlsl::resource_class` would be
 `SRV`, `hlsl::is_rov` would not be present, `hlsl::texture_dimension`
 would be `2DArray`, and `hlsl::texture_type` would be `MS`.
 
-| HLSL Texture Type              | hlsl::resource_class | hlsl::is_rov | hlsl::texture_dimension | hlsl::texture_type |
-| ------------------------------- | -------------------- | ------------ | ----------------------- | ------------------ |
-| Texture1D                       | SRV                  | -            | 1D                      | -                  |
-| RWTexture1D                     | UAV                  | -            | 1D                      | -                  |
-| RasterizerOrderedTexture1D      | SRV                  | yes          | 1D                      | -                  |
-| Texture1DArray                  | SRV                  | -            | 1DArray                 | -                  |
-| RWTexture1DArray                | UAV                  | -            | 1DArray                 | -                  |
-| RasterizerOrderedTexture1DArray | SRV                  | yes          | 1DArray                 | -                  |
-| Texture2D                       | SRV                  | -            | 2D                      | -                  |
-| RWTexture2D                     | UAV                  | -            | 2D                      | -                  |
-| RasterizerOrderedTexture2D      | SRV                  | yes          | 2D                      | -                  |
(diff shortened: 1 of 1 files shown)


proposals/0006-resource-attributes.md
```diff
@@ -0,0 +1,25 @@
...
+| Texture2DArray                  | SRV                  | -            | 2DArray                 | -                  |
+| RWTexture2DArray                | UAV                  | -            | 2DArray                 | -                  |
+| RasterizerOrderedTexture2DArray | SRV                  | yes          | 2DArray                 | -                  |
+| Texture3D                       | SRV                  | -            | 3D                      | -                  |
+| RWTexture3D                     | UAV                  | -            | 3D                      | -                  |
+| RasterizerOrderedTexture3D      | SRV                  | yes          | 3D                      | -                  |
+| TextureCUBE                     | SRV                  | -            | CUBE                    | -                  |
+| TextureCUBEArray                | SRV                  | -            | CUBEArray               | -                  |
+| Texture2DMS                     | SRV                  | -            | 2D                      | MS                 |
```
damyanp, line 20: If we have is_rov, should we have is_multisampled and is_feedback?

damyanp:
FWIW, I think we've gone too far in the "make the proposal shorter" direction :)  It'd be good to get this filled out with more of the proposal stuff.  Maybe you were planning this already, it is a draft after all.

A couple of examples of how and where we expect the attributes to appear would be good as well.

proposals/0006-resource-attributes.md
bob80905: That'll be part of the discussion!

proposals/0006-resource-attributes.md
```diff
@@ -0,0 +1,78 @@
...
+in an HLSL header, and the type itself would be constructed with
+specific attributes attached to it: `hlsl::resource_class` would be 
+`SRV`, `hlsl::is_rov` would not be present, `hlsl::texture_dimension`
+would be `2DArray`, and `hlsl::texture_type` would be `MS`.
+
+| HLSL Texture Type              | hlsl::resource_class | hlsl::is_rov | hlsl::texture_dimension | hlsl::texture_type |
+| ------------------------------- | -------------------- | ------------ | ----------------------- | ------------------ |
+| Texture1D                       | SRV                  | -            | 1D                      | -                  |
+| RWTexture1D                     | UAV                  | -            | 1D                      | -                  |
+| RasterizerOrderedTexture1D      | SRV                  | yes          | 1D                      | -                  |
```
damyanp, line 52: Courtesy of Brian- the ROV resources are UAVs, not SRVs.

damyanp (changes requested):
Before we can accept this I would like to see some indication of how these attributes will be used.

Take, for example, `hlsl::texture_dimension` - what will this be _used_ for?

If one of the uses of it is to determine things about the `operator[]`'s that are available, how will this work for buffers?  Are they a special case?

A bit of me suspects that it might make sense to make the `texture_dimension` thing be more of a `resource_dimension` - and so buffers will be 1D resources, like 1D textures are. 

Note though: the change I'm requesting here is to fix the entry for the buffers in the table. I think fixing that would be enough to merge this PR, but we'll need some follow up work in order to get the proposal ready to be accepted.

proposals/0006-resource-attributes.md
```diff
@@ -0,0 +1,82 @@
...
+| RWTexture2DMS                   | UAV                  | -            | 2                       | -             | yes         | -                 | -              |
+| RWTexture2DMSArray              | UAV                  | -            | 2                       | -             | yes         | -                 | yes            |
+| FeedbackTexture2D               | SRV                  | -            | 2                       | -             | -           | yes               | -              |
+| FeedbackTexture2DArray          | SRV                  | -            | 2                       | -             | -           | yes               | yes            |
+| TypedBuffer                     | SRV                  | -            | -                       | -             | -           | yes               | yes            |
+| RawBuffer                       | SRV                  | -            | -                       | -             | -           | yes               | yes            |
+| StructuredBuffer                | SRV                  | -            | -                       | -             | -           | yes               | yes            |
```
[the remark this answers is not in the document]
bob80905: Ok, I've adjusted the buffer entries.

bob80905:
> Before we can accept this I would like to see some indication of how these attributes will be used.
> 
> Take, for example, `hlsl::texture_dimension` - what will this be _used_ for?
> 
> If one of the uses of it is to determine things about the `operator[]`'s that are available, how will this work for buffers? Are they a special case?
> 
> A bit of me suspects that it might make sense to make the `texture_dimension` thing be more of a `resource_dimension` - and so buffers will be 1D resources, like 1D textures are. 
> 
> Note though: the change I'm requesting here is to fix the entry for the buffers in the table. I think fixing that would be enough to merge this PR, but we'll need some follow up work in order to get the proposal ready to be accepted.


The texture dimension preserves the dimension, and may be used in the future to distinguish between the right load/store overloads to use for the handle.
Though this use isn't currently set in stone, it was briefly proposed as an idea here: 
https://github.com/llvm/llvm-project/pull/90553/commits/4be43928206617ad2b9387fbd207758a4cd24de7#diff-5d235a31f5df8cd876e698f807bb6042678a30a6e1ea6f5649e879a8914504edR167

@bogner will solidify the handle creation design, but has stated before that it's important to capture all information that could be useful in constructing the handle, and dimension seems to be useful.

damyanp:
I'm not sure that creating a bunch of attributes to store things that we think might be useful really counts as designing a feature.  How can we evaluate whether or not the design of the attribute works or not if we don't know what the attribute is for?

For example, how will the implementation of GetDimensions() make use of these attributes?  How about the various Atomic* operations?

Will these need to have to special case something for buffers, when it could be something that falls out naturally from the definition of them? If they are special cases, what attribute does it look at to figure this out?

Maybe it won't, but at the moment this proposal tells us nothing about that.

proposals/0006-resource-attributes.md
```diff
@@ -0,0 +1,82 @@
...
+| RWTexture2DMS                   | UAV                  | -            | 2                       | -             | yes         | -                 | -              |
+| RWTexture2DMSArray              | UAV                  | -            | 2                       | -             | yes         | -                 | yes            |
+| FeedbackTexture2D               | SRV                  | -            | 2                       | -             | -           | yes               | -              |
+| FeedbackTexture2DArray          | SRV                  | -            | 2                       | -             | -           | yes               | yes            |
+| TypedBuffer                     | SRV                  | -            | -                       | -             | -           | -                 | -              |
+| RawBuffer                       | SRV                  | -            | -                       | -             | -           | -                 | -              |
+| StructuredBuffer                | SRV                  | -            | -                       | -             | -           | -                 | -              |
```
tex3d, line 75: The three buffer types are indistinguishable according to these attributes, and I expect we want to cover the missing `ConstantBuffer` and `TextureBuffer` types here as well.

Here are some ideas for extending the attributes to distinguish buffer types and incorporate `ConstantBuffer`/`TextureBuffer`:

- Add `[[hlsl::row_layout]]` for `ConstantBuffer` and `TextureBuffer` to identify where special cbuffer row layout rules are used with untranslated components from the constant buffer or typed buffer.
- `[[hlsl::row_layout]]` would be combined with `[[hlsl::resource_class(CBV)]]` for ConstantBuffer or `[[hlsl::resource_class(SRV)]]` for TextureBuffer.
- Add `[[hlsl::raw_buffer]]` for `RawBuffer` and `StructuredBuffer`.
- Add `[[hlsl::struct_stride(n)]]` for `StructuredBuffer`, where `n` is the stride for the structured buffer in bytes.

llvm-beanz:
Closing as subsumed by #68

PR #47, opened 2024-08-26
tools/hlslpm: always get issue titles, and don't accept things that aren't issue references in section titles

damyanp:
Previously, this would blindly accept something like `### Foo (??)` as a section heading referring to an issue with reference `??`.  Various things fell over after this, so this change instead only accepts things that look like github-style issues references.

Also included in this change is the infrastructure added to help track down this issue - namely, grabbing more information about each issue and displaying information about that when something goes wrong.

### DIFF
--- a/tools/hlslpm/issue.py
+++ b/tools/hlslpm/issue.py
@@ -4,7 +4,7 @@ from dataclasses import dataclass, field
 from datetime import datetime, date
 from enum import Enum
 import re
-from typing import Dict, List, Optional, Union
+from typing import Dict, List, Optional, Self, Union
 import json
 
 class Category(Enum):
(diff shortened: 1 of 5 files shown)


tools/hlslpm/issue.py
```diff
@@ -27,7 +27,7 @@ def getReferenceFromTitle(self):
         if not self.title:
             return None
 
-        m = re.match(".*\((.+)\)", self.title)
+        m = re.match(r".*\((.*#\d+)\)", self.title)
```
dmpots, line 30: Something I tend to forget about is that they added assignment expressions in python 3.8 (https://docs.python.org/3/reference/simple_stmts.html#grammar-token-python-grammar-assignment_stmt).

Not sure it would be better here, but thought I'd throw it out in case it was new to you.
```
if m := re.match(...):
  # m is valid here
```

dmpots (approved):
LGTM

tools/hlslpm/issue.py
damyanp: Thanks, that was new to me.

llvm-beanz (approved):

PR #62, opened 2024-09-05
A proposal for DXIL Function Scalarization

farzonl:
This covers the motivations, background of how Scalarization is working in DXC
 and the approach to tasking we should take.
This is also and updated proposal to what was decided in team meeting

It does not cover Data scalarization.

### DIFF
--- a/proposals/NNNN-DXIL-Scalarization.md
+++ b/proposals/NNNN-DXIL-Scalarization.md
@@ -10,9 +10,9 @@
 
 we need to be able to scalarize data structures, call instructions, and vector
 operations like math ops, logical ops, bitcasts, loads, and stores. The goal of
- this proposal is to present a solution for scalarizing vector operation and
- call instructions via LLVM's pass manager. This proposal will not cover data
- layout transformations.
+this proposal is to present a solution for scalarizing vector operation and
+call instructions via LLVM's pass manager. This proposal will not cover data
+layout transformations.
 
 ## Motivation
 
@@ -37,19 +37,19 @@ that end it makes the most sense to use the Scalarizer pass.
 
 * [Scalarizer.cpp](https://github.com/llvm/llvm-project/blob/main/llvm/lib/Transforms/Scalar/Scalarizer.cpp)
 
-The `Scalarizer` pass is a `FunctionPass`. As such it won't transform global 
-data structures. The `scalarizer` pass has two flags relevant to our use case: 
-`-scalarize-load-store` and `-scalarize-variable-insert-extract`. These flags 
-will handle the `BitCastInst`,`CallInst`, `LoadInst` and `StoreInst`, and 
+The `Scalarizer` pass is a `FunctionPass`. As such it won't transform global
+data structures. The `scalarizer` pass has two flags relevant to our use case:
+`-scalarize-load-store` and `-scalarize-variable-insert-extract`. These flags
+will handle the `BitCastInst`,`CallInst`, `LoadInst` and `StoreInst`, and
 `GetElementPtrInst` cases.
 
-The `AddrSpaceCastInst` case is unique to DXC. DXC was using `AddrSpaceCastInst` 
-to fix up bad codgen to avoid undefined behavior or generating ilegal DXIL. 
+The `AddrSpaceCastInst` case is unique to DXC. DXC was using `AddrSpaceCastInst`
+to fix up bad codgen to avoid undefined behavior or generating ilegal DXIL.
 Any remaining `AddrSpaceCastInst` should be handled by the O1 optimization
 pipeline.
 
 `MemIntrinsic` is the odd case out. Clang does not currently emit memcpy for
-the same cases as DXC. In DXC operations like memcpy and memset most easily 
+the same cases as DXC. In DXC operations like memcpy and memset most easily
 happen in global scope.
 
 In my observation these cases get converted into cbuffer. Since we don't have
@@ -66,11 +66,11 @@ The team determined it was best for this pass to run in the DirectX backend.
 That allows The DirectX backend to be agnostic to the frontend.
 The team also determined the pass should run as late as possible. This has two
 benefits. First, code size: if scalarization happens too early then things like
-the `-combiner-alias-analysis` pass limits are reached. For this particular 
-pass and potentially others if the limits are reached it does not perfom a 
-transformation. That means there is the potential for less efficent code 
-generation if we scalarize to soon as some ooptimizations that benefit from 
-less IR wont run. Second, scalarization for DXIL should be considered a 
+the `-combiner-alias-analysis` pass limits are reached. For this particular
+pass and potentially others if the limits are reached it does not perfom a
+transformation. That means there is the potential for less efficent code
+generation if we scalarize to soon as some ooptimizations that benefit from
+less IR wont run. Second, scalarization for DXIL should be considered a
 legalization step. To that end it should happen at or right before
 `DXILOpLowering`.
 
@@ -80,6 +80,7 @@ it only converts the scalarized llvm intrinsics meaning there is no way for it
 to know about DXIL OPs.
 
 ### Proposal: Ways to add the Scalarizer pass
+
 The Scalarizer pass has been converted to the new pass manager (new PM). That
 is a problem for us because the new PM is only experimental in the backend.
 The Direct X backend is still on the legacy PM. Further, new PM isn't
--- a/proposals/NNNN-DXIL-Scalarization.md
+++ b/proposals/NNNN-DXIL-Scalarization.md
@@ -1,0 +1,90 @@
+# DXIL Scalarization
+
+* Proposal: [NNNN](NNNN-DXIL-Scalarization.md)
+* Author(s): [Farzon Lotfi](https://github.com/farzonl)
+* Sponsor: [Farzon Lotfi](https://github.com/farzonl)
+* Status: **Under Consideration**
+* Impacted Projects: Clang
+
+## Introduction
+As mentioned in [DXIL.rst](https://github.com/microsoft/DirectXShaderCompiler/blob/main/docs/DXIL.rst#vectors) 
+"HLSL vectors are scalarized" and "Matrices are lowered to vectors". Therefore,
+ we need to be able to scalarize datastructures, call instructions, and vector 
+ operations like mathamatical, logical and bitcasts. The goal of this proposal 
+ is to present a solution for LLVM's pass manager.
+
+## Motivation
+Without Scalarizing the data structures and call instructions we can't generate
+ valid DXIL.
+
+ ## Background
+ In DXC Scalarization and SROA are merged in unfortnate ways. Most of that code
+ can be found here: [`SROA_Parameter_HLSL` Module pass in `ScalarReplAggregatesHLSL.cpp`](https://github.com/microsoft/DirectXShaderCompiler/blob/main/lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp#L4263). 
+The specific instructions DXC operates on are `CallInst`, `LoadInst`, 
+`StoreInst`, `GetElementPtrInst`, `AddrSpaceCastInst`, `BitCastInst`, and 
+`MemIntrinsic`. Not covered in this proposal is the handling of `Allocas`.
+That will be addressed in a later proposal covering DXIL Legalization.
+
+## Proposal
+Our goal should be legal DIXL with behavioral similarity to DXC. To that end it
+ makes the most sense to use the Scalarizer pass.
+- [Scalarizer.cpp](https://github.com/llvm/llvm-project/blob/main/llvm/lib/Transforms/Scalar/Scalarizer.cpp)
+
+The `scalarizer` pass with the `-scalarize-load-store` and 
+`-scalarize-variable-insert-extract` flags  will handle the `BitCastInst`, 
+`CallInst`, `LoadInst` and `StoreInst`, and `GetElementPtrInst` cases. The 
+`AddrSpaceCastInst` case isn't relevant for upstream LLVM.
+
+`MemIntrinsic` is the odd case out. We don't currently emit memcpy for
+the same cases we do in DXC.
+In many cases operations like memcpy and memset seem to happen in global scope.
+See Example:
+- https://hlsl.godbolt.org/z/8vx1Yee7b
+
+In my observation these cases get converted into cbuffer. Since we don't have
+cbuffer support yet we can likely hold off on this cases.
+
+
+The team debated when how early or late this pass should run and three ways to 
+onboard this pass. There was also a discussion on  the scalarizer passes only 
+working on llvm intrinsics and how to allow for DirectX Intrinsics.
+
+### Proposal: When to run pass
+The team determined it was best for this pass to run in the DirectX backend.
+That allows The DirectX backend to be agnostic to the frontend.
+The team also determined the pass should run as late as possible. This has two
+benefits. First code size. if scalarization happens too early things like the 
+`-combiner-alias-analysis` pass limits are reached. More optimizations will be
+possible as analysis passes will benefit from less IR. Second, scalarization 
+for DXIL should be considered a legalization step. To that end it should happen
+at or right before `DXILOpLowering`.
+
+`DXILOpLowering` is also the last place for a functional reason. The scalarizer
+pass only operates on llvm intrinsics that are `TriviallyVectorizable`. Further
+it only converts the scalarized llvm intrinsics meaning there is no way for it
+to know abot DXIL OPs.
+
+### Proposal: Ways to add the Scalarizer Pass
+There are a few ways we could go about this:
+1. support `PassInfoMixin` in the DirectX backend. 
+   - Need to override "buildCodeGenPipeline" for the DirectX backend.
+   - Pro: implementation is easy enough: https://github.com/llvm/llvm-project/compare/main...farzonl:llvm-project:create-codegen-pass-builder
+   - Con: The new PM isn't hooked up in clang yet via `BackendUtil.cpp`'s
+     `AddEmitPasses`. the `TM->addPassesToEmitFile` is all legacy PM.
+2. Add the Legacy Pass manager abstraction for `scalarizer` back
+   - We would be undoing the work the community did to port this pass.
+   - Pro: works seemlessly: https://github.com/llvm/llvm-project/compare/main...farzonl:llvm-project:add-legacy-scalarizer-for-dxil
+3. Fork the `scalarizer` so we can have a legacy and `PassInfoMixin` in the 
+   DirectX Backend?
+   - We could make this work for dxil ops and not just intrinsics.
+   - Con: Mantaining passes like this defeats some of the benefits of the
+     modernization effort.
+
+The team decided the best path forward was to add the legacy pass manager back.
+
+### Proposal: Add Support for DirectX intrinsics
+`isTriviallyScalariable(Intrinsic::ID ID)` will be renamed to `isTriviallyScalarizable`.
+We will add a Module parameter so that we can lookup `M.getTargetTriple();`. 
+We will then have a condition on `Triple::ArchType ==  == Triple::dxil` that 
+will call a static function with all the DirectX intrinsics that need to be
+scalarized specified.
--- a/proposals/NNNN-DXIL-Scalarization.md
+++ b/proposals/NNNN-DXIL-Scalarization.md
@@ -37,11 +37,16 @@ that end it makes the most sense to use the Scalarizer pass.
 
 * [Scalarizer.cpp](https://github.com/llvm/llvm-project/blob/main/llvm/lib/Transforms/Scalar/Scalarizer.cpp)
 
-The `Scalarizer` pass is a `FunctionPass`. As such it won't transform global
-data structures. The `scalarizer` pass has two flags relevant to our use case:
-`-scalarize-load-store` and `-scalarize-variable-insert-extract`. These flags
-will handle the `BitCastInst`,`CallInst`, `LoadInst` and `StoreInst`, and
-`GetElementPtrInst` cases.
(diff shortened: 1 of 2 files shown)


proposals/NNNN-DXIL-Scalarization.md
```diff
@@ -0,0 +1,90 @@
...
+ operations like mathamatical, logical and bitcasts. The goal of this proposal 
+ is to present a solution for LLVM's pass manager.
+
+## Motivation
+Without Scalarizing the data structures and call instructions we can't generate
+ valid DXIL.
+
+ ## Background
+ In DXC Scalarization and SROA are merged in unfortnate ways. Most of that code
+ can be found here: [`SROA_Parameter_HLSL` Module pass in `ScalarReplAggregatesHLSL.cpp`](https://github.com/microsoft/DirectXShaderCompiler/blob/main/lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp#L4263).
```
damyanp, line 22: nit: some inconsistent indentation going on around this part of the file

proposals/NNNN-DXIL-Scalarization.md
```diff
@@ -0,0 +1,90 @@
...
+ we need to be able to scalarize datastructures, call instructions, and vector 
+ operations like mathamatical, logical and bitcasts. The goal of this proposal 
+ is to present a solution for LLVM's pass manager.
+
+## Motivation
+Without Scalarizing the data structures and call instructions we can't generate
+ valid DXIL.
+
+ ## Background
+ In DXC Scalarization and SROA are merged in unfortnate ways. Most of that code
```
damyanp, line 21:
```suggestion
 In DXC, Scalarization and SROA are merged in unfortunate ways. Most of that code
```

proposals/NNNN-DXIL-Scalarization.md
```diff
@@ -0,0 +1,90 @@
...
+ can be found here: [`SROA_Parameter_HLSL` Module pass in `ScalarReplAggregatesHLSL.cpp`](https://github.com/microsoft/DirectXShaderCompiler/blob/main/lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp#L4263). 
+The specific instructions DXC operates on are `CallInst`, `LoadInst`, 
+`StoreInst`, `GetElementPtrInst`, `AddrSpaceCastInst`, `BitCastInst`, and 
+`MemIntrinsic`. Not covered in this proposal is the handling of `Allocas`.
+That will be addressed in a later proposal covering DXIL Legalization.
+
+## Proposal
+Our goal should be legal DIXL with behavioral similarity to DXC. To that end it
```
damyanp, line 29:
```suggestion
Our goal should be to emit legal DXIL with behavioral similarity to DXC. To that end it
```

proposals/NNNN-DXIL-Scalarization.md
```diff
@@ -0,0 +1,90 @@
...
+ makes the most sense to use the Scalarizer pass.
+- [Scalarizer.cpp](https://github.com/llvm/llvm-project/blob/main/llvm/lib/Transforms/Scalar/Scalarizer.cpp)
+
+The `scalarizer` pass with the `-scalarize-load-store` and 
+`-scalarize-variable-insert-extract` flags  will handle the `BitCastInst`, 
+`CallInst`, `LoadInst` and `StoreInst`, and `GetElementPtrInst` cases. The 
+`AddrSpaceCastInst` case isn't relevant for upstream LLVM.
```
damyanp, line 36: > The `AddrSpaceCastInst` case isn't relevant for upstream LLVM.

Why isn't it relevant?  Presumably this is something you had to figure out so it'd be nice to get that discovery written down somewhere.

proposals/NNNN-DXIL-Scalarization.md
```diff
@@ -0,0 +1,90 @@
...
+
+`MemIntrinsic` is the odd case out. We don't currently emit memcpy for
+the same cases we do in DXC.
+In many cases operations like memcpy and memset seem to happen in global scope.
+See Example:
+- https://hlsl.godbolt.org/z/8vx1Yee7b
```
damyanp, line 42: I'm not entirely sure what I'm meant to be looking at on this godbolt link. Are the clang ones likely to remain stable (eg, after you've done this work will the clang links no longer make the point they're trying to make now?)

proposals/NNNN-DXIL-Scalarization.md
```diff
@@ -0,0 +1,90 @@
...
+- [Scalarizer.cpp](https://github.com/llvm/llvm-project/blob/main/llvm/lib/Transforms/Scalar/Scalarizer.cpp)
+
+The `scalarizer` pass with the `-scalarize-load-store` and 
+`-scalarize-variable-insert-extract` flags  will handle the `BitCastInst`, 
+`CallInst`, `LoadInst` and `StoreInst`, and `GetElementPtrInst` cases. The 
+`AddrSpaceCastInst` case isn't relevant for upstream LLVM.
+
+`MemIntrinsic` is the odd case out. We don't currently emit memcpy for
+the same cases we do in DXC.
+In many cases operations like memcpy and memset seem to happen in global scope.
```
damyanp, line 40: > In many cases operations like memcpy and memset seem to happen in global scope.

In DXC or in clang? Or both? For the godbolt link below I only see memcpy in the dxc -fcgl version.

proposals/NNNN-DXIL-Scalarization.md
```diff
@@ -0,0 +1,90 @@
...
+See Example:
+- https://hlsl.godbolt.org/z/8vx1Yee7b
+
+In my observation these cases get converted into cbuffer. Since we don't have
+cbuffer support yet we can likely hold off on this cases.
```
damyanp, line 45: Aren't we also building cbuffer support right now though?

proposals/NNNN-DXIL-Scalarization.md
```diff
@@ -0,0 +1,90 @@
...
+
+
+The team debated when how early or late this pass should run and three ways to
```
damyanp, line 48:
```suggestion
The team debated how early or late this pass should run and considered(?) three ways to 
```

proposals/NNNN-DXIL-Scalarization.md
```diff
@@ -0,0 +1,90 @@
...
+See Example:
+- https://hlsl.godbolt.org/z/8vx1Yee7b
+
+In my observation these cases get converted into cbuffer. Since we don't have
+cbuffer support yet we can likely hold off on this cases.
+
+
+The team debated when how early or late this pass should run and three ways to 
+onboard this pass. There was also a discussion on  the scalarizer passes only 
+working on llvm intrinsics and how to allow for DirectX Intrinsics.
```
damyanp, line 50:
```suggestion
onboard this pass. There was also a discussion on how the scalarizer passes only 
work on llvm intrinsics and how we would extend it to allow for DirectX Intrinsics.
```
?

proposals/NNNN-DXIL-Scalarization.md
```diff
@@ -0,0 +1,90 @@
...
+
+### Proposal: When to run pass
+The team determined it was best for this pass to run in the DirectX backend.
+That allows The DirectX backend to be agnostic to the frontend.
+The team also determined the pass should run as late as possible. This has two
+benefits. First code size. if scalarization happens too early things like the
```
damyanp, line 56:
```suggestion
benefits. First code size: if scalarization happens too early then things like the 
```

proposals/NNNN-DXIL-Scalarization.md
```diff
@@ -0,0 +1,90 @@
...
+onboard this pass. There was also a discussion on  the scalarizer passes only 
+working on llvm intrinsics and how to allow for DirectX Intrinsics.
+
+### Proposal: When to run pass
+The team determined it was best for this pass to run in the DirectX backend.
+That allows The DirectX backend to be agnostic to the frontend.
+The team also determined the pass should run as late as possible. This has two
+benefits. First code size. if scalarization happens too early things like the 
+`-combiner-alias-analysis` pass limits are reached. More optimizations will be
+possible as analysis passes will benefit from less IR. Second, scalarization
```
damyanp, line 58: > More optimizations will be possible as analysis passes will benefit from less IR.

Is this saying that the analysis passes will be able to perform better analysis if there is less IR?  Or do the pass limits mentioned previously mean that some passes are disabled?

proposals/NNNN-DXIL-Scalarization.md
```diff
@@ -0,0 +1,90 @@
...
+The team also determined the pass should run as late as possible. This has two
+benefits. First code size. if scalarization happens too early things like the 
+`-combiner-alias-analysis` pass limits are reached. More optimizations will be
+possible as analysis passes will benefit from less IR. Second, scalarization 
+for DXIL should be considered a legalization step. To that end it should happen
+at or right before `DXILOpLowering`.
+
+`DXILOpLowering` is also the last place for a functional reason. The scalarizer
+pass only operates on llvm intrinsics that are `TriviallyVectorizable`. Further
+it only converts the scalarized llvm intrinsics meaning there is no way for it
```
damyanp, line 64:
```suggestion
it only converts to(?) scalarized llvm intrinsics meaning there is no way for it
```

proposals/NNNN-DXIL-Scalarization.md
```diff
@@ -0,0 +1,90 @@
...
+to know abot DXIL OPs.
```
damyanp, line 65:
```suggestion
to know about DXIL OPs.
```

proposals/NNNN-DXIL-Scalarization.md
```diff
@@ -0,0 +1,90 @@
...
+
+### Proposal: Ways to add the Scalarizer Pass
+There are a few ways we could go about this:
```
damyanp, line 68: I think that this section is describing some potential solutions to a problem, but the problem hasn't been explained previously. IIRC, something to do with the legacy pass manager?

proposals/NNNN-DXIL-Scalarization.md
```diff
@@ -0,0 +1,90 @@
...
+3. Fork the `scalarizer` so we can have a legacy and `PassInfoMixin` in the 
+   DirectX Backend?
+   - We could make this work for dxil ops and not just intrinsics.
+   - Con: Mantaining passes like this defeats some of the benefits of the
+     modernization effort.
+
+The team decided the best path forward was to add the legacy pass manager back.
+
+### Proposal: Add Support for DirectX intrinsics
+`isTriviallyScalariable(Intrinsic::ID ID)` will be renamed to `isTriviallyScalarizable`.
```
damyanp, line 86: > `isTriviallyScalariable(Intrinsic::ID ID)` will be renamed to `isTriviallyScalarizable`.

Is this just us fixing a typo? Or is "scalariable" a real word and renaming it in some way is changing the meaning?

damyanp (commented):
Some comments mostly calling out typos and bits I had trouble following.

proposals/NNNN-DXIL-Scalarization.md
```diff
@@ -0,0 +1,90 @@
...
+`StoreInst`, `GetElementPtrInst`, `AddrSpaceCastInst`, `BitCastInst`, and 
+`MemIntrinsic`. Not covered in this proposal is the handling of `Allocas`.
+That will be addressed in a later proposal covering DXIL Legalization.
+
+## Proposal
+Our goal should be legal DIXL with behavioral similarity to DXC. To that end it
+ makes the most sense to use the Scalarizer pass.
+- [Scalarizer.cpp](https://github.com/llvm/llvm-project/blob/main/llvm/lib/Transforms/Scalar/Scalarizer.cpp)
+
+The `scalarizer` pass with the `-scalarize-load-store` and
```
python3kgae, line 33: What is the plan for static global variables with vector or vector array type and  groupshared variables with vector array type?

farzonl: Those are handled by the same scalarizer pass with the load store flag. Here is a groushared example
```hlsl
groupshared float3 sharedData[2]; 
export void fn2() { 
    sharedData[0] = float3(1.0f, 2.0f, 3.0f); 
    sharedData[1] = float3(2.0f, 4.0f, 6.0f); 
} 
```
generates the following IR
```llvm
@"sharedData" = local_unnamed_addr addrspace(3) global [2 x <3 x float>] zeroinitializer, align 16 
define void @fn2 () local_unnamed_addr { 
    store <3 x float> <float 1.000000e+00, float 2.000000e+00, float 3.000000e+00>, ptr addrspace(3) @"sharedData", align 16 
    store <3 x float> <float 2.000000e+00, float 4.000000e+00, float 6.000000e+00>, ptr addrspace(3)   getelementptr inbounds (i8, ptr addrspace(3) @"sharedData", i32 16), align 16 
    ret void
 }
```
```powershell
opt.exe -S -passes=scalarizer -scalarize-load-store  scalarize_store.ll 
```
```llvm
@"sharedData" = local_unnamed_addr addrspace(3) global [2 x <3 x float>] zeroinitializer, align 16 
define void @"?fn2@@YAXXZ"() local_unnamed_addr { 
store float 1.000000e+00, ptr addrspace(3) @"sharedData", align 16 
store float 2.000000e+00, ptr addrspace(3) getelementptr (float, ptr addrspace(3) @"sharedData", i32 1), align 4 
store float 3.000000e+00, ptr addrspace(3) getelementptr (float, ptr addrspace(3) @"sharedData", i32 2), align 8 
store float 2.000000e+00, ptr addrspace(3) getelementptr inbounds (i8, ptr addrspace(3) @"sharedData", i32 16), align 16 
store float 4.000000e+00, ptr addrspace(3) getelementptr (float, ptr addrspace(3) getelementptr inbounds (i8, ptr addrspace(3) @"sharedData", i32 16), i32 1), align 4 
store float 6.000000e+00, ptr addrspace(3) getelementptr (float, ptr addrspace(3) getelementptr inbounds (i8, ptr addrspace(3) @"sharedData", i32 16), i32 2), align 8 
ret void 
} 
```

python3kgae: This type of @"sharedData" is still [2 x <3 x float>] after opt.
We need it to be [2 x [3 x float]] for groupshared.

For static global, we could split it to 3 [2 x float] globals if possible.

farzonl: My previous proposal covered global scope, but this one doesn't. i'll make that more explicit. This is a `FunctionPass` so it won't change the layout of global scope just how we access that data.  The last piece that DXC does that this pass doesn't is to encode `sharedData` as a 6 element 1d array. All the store operations are correct.

```llvm
!20 = !DIGlobalVariable(name: "sharedData", linkageName: "\01?sharedData@@3PAV?$vector@M$02@@A.v.1dim", scope: !0, file: !1, line: 1, type: !21, isLocal: false, isDefinition: true, variable: [6 x float] addrspace(3)* @"\01?sharedData@@3PAV?$vector@M$02@@A.v.1dim")
```

python3kgae: So we'll have module pass to cover global variables?
Flat multiple dim array into 1d array should be a separate pass, not part of scalarization.

proposals/NNNN-DXIL-Scalarization.md
farzonl: DXC was using `AddrSpaceCastInst` to fix up bad codgen to avoid undefined behavior or DXIL generation that wouldn't be legal. This case should be handled for us by existing clang optimization passes available in `-O1`.

proposals/NNNN-DXIL-Scalarization.md
farzonl: I can remove the godbolt link. Yes after we have done the work the godbolt like might no longer be valid as the clang portion is on trunk.

 The point I was trying to make is that we should wait to handle the `MemIntrinsic` cases until we have cbuffer support. If I were to figure it out today it would be a direct translation. But I imagine there might be some intermediary we want because the other buffer cases happen after DXIL lowering and we are doing scalarization before.

proposals/NNNN-DXIL-Scalarization.md
farzonl: In DXC.  I'll specify DXC in the paragraph. I just didn't generate the other cases, but the code that inserts `memcpy` aslo handles the `memset` and `memmove`  emits. It looks like this:
```cpp
// Finally, insert the meminst for this element.
    if (isa<MemSetInst>(MI)) {
      Builder.CreateMemSet(EltPtr, MI->getArgOperand(1), EltSize,
                           MI->isVolatile());
    } else {
      assert(isa<MemTransferInst>(MI));
      Value *Dst = SROADest ? EltPtr : OtherElt; // Dest ptr
      Value *Src = SROADest ? OtherElt : EltPtr; // Src ptr

      if (isa<MemCpyInst>(MI))
        Builder.CreateMemCpy(Dst, Src, EltSize, OtherEltAlign,
                             MI->isVolatile());
      else
        Builder.CreateMemMove(Dst, Src, EltSize, OtherEltAlign,
                              MI->isVolatile());
    }
```
So DXC inserts the Mem intrinsics as part of `RewriteMemIntrin`

proposals/NNNN-DXIL-Scalarization.md
farzonl: We are but this proposal is to onboard a generic scalarizer pass that is not going to have anything unique to Direct X. There is also some out of order problems between the two work streams. from talking with Justin that work is happening after DXIL lowering and scalarization is happening before.

 To me this seems like a case we can handle separately in another proposal. We already need to have our own special DXIL pass for global arrays to convert `[2 x <3 x float>]` to `[2 x [3 x float]]`.

proposals/NNNN-DXIL-Scalarization.md
```diff
@@ -0,0 +1,90 @@
+# DXIL Scalarization
+
+* Proposal: [NNNN](NNNN-DXIL-Scalarization.md)
+* Author(s): [Farzon Lotfi](https://github.com/farzonl)
+* Sponsor: [Farzon Lotfi](https://github.com/farzonl)
+* Status: **Under Consideration**
+* Impacted Projects: Clang
```
python3kgae, line 7: LLVM here?

proposals/NNNN-DXIL-Scalarization.md
farzonl: There are some passes that have hard limits on code size `combiner-alias-analysis` was one of them. In fact, load store scalarization isn't on by default because it can affect that particular pass.

proposals/NNNN-DXIL-Scalarization.md
python3kgae: We may need to update the description for this PR to clarify the onboarding of the generic scalarizer pass.
The current description is very broad.

proposals/NNNN-DXIL-Scalarization.md
farzonl: in our past proposals in HLSL specs we have used Clang to indicate the upstream LLVM project and DXC to indicate DXC, and Clang and DXC to indicate both. I'm Keeping that same distinction, but open to changing it.

proposals/NNNN-DXIL-Scalarization.md
farzonl: There will have to be another pass to handle data layout like this. Thats already something I'm considering because of memory intrinsics.

proposals/NNNN-DXIL-Scalarization.md
farzonl: I checked a few spell checkers `Scalariable` isn't a word. `Scalarizable` isn't a word either, but it matches closer to the naming of the pass. My feeling is they mixed `Scalarize` with `Scalable` and dropped the `z`.

proposals/NNNN-DXIL-Scalarization.md
farzonl: no the is correct here.

proposals/NNNN-DXIL-Scalarization.md
farzonl: fixed

proposals/NNNN-DXIL-Scalarization.md
farzonl: your comment came before I updated the proposal. I think my updates address this concern but please confirm.

proposals/NNNN-DXIL-Scalarization.md
damyanp: So the goal of renaming it is just to clarify what it does, not actually any kind of functional change?

farzonl: Correct.

python3kgae (approved):

proposals/NNNN-DXIL-Scalarization.md
damyanp: > Further it only converts the scalarized llvm intrinsics

If this is correct, then I'm afraid I'm not sure I understand what it means.  I would have thought that the scalarizer converts the vectorized intrinics to scalar ones. So I guess I've got a gap in my understanding here.

damyanp (approved):

proposals/NNNN-DXIL-Scalarization.md
farzonl: So when i say llvm intrinsics i mean intrinsics defined in intrinsics.td. So im trying to make a distinction between the intrinsics that are exposed to all backends in intrinsics.td and the direct x intrinsics. The distinction being that if we want this to work with our backend we need a way of exposing the direct x intrinsics to the scalarizer pass.

proposals/NNNN-DXIL-Scalarization.md
llvm-beanz: I think we should probably just remove the `Impacted Projects` line from the template in this repo. Nothing goes into this repo that isn't targeting LLVM/Clang.

llvm-beanz (approved):

PR #69, opened 2024-09-13
[0011] Resource element type validation

bob80905:
This spec describes how the compiler will validate resource element types.
This spec is needed before the implementation, which will eventually solve https://github.com/llvm/llvm-project/issues/75676

### DIFF
--- a/proposals/0010-resource-element-type-validation.md
+++ b/proposals/0010-resource-element-type-validation.md
@@ -16,10 +16,10 @@ In this code, the RET is `float`, and the resource type is `RWBuffer`.
 All resources can be placed in two categories, raw buffers and typed buffers. 
 Below is a description of all resources and their corresponding categories
 * raw buffers
-  * [Append|Consume|RW]StructuredBuffer
+  * [Append|Consume|RW|RasterizerOrdered]StructuredBuffer
   * [RW]ByteAddressBuffer
 * typed buffers
-  * [RW]Buffer
+  * [RW|RasterizerOrdered]Buffer
   * [Feedback]Texture*
 
 There is a distinct set of rules that define valid RETs for raw buffer resources, 
@@ -65,7 +65,7 @@ them when constructing the AST in Sema.
 A new built-in, `__builtin_hlsl_is_line_vector_layout_compatible`, will be 
 added in order to express the extra typed buffer constraint. This builtin
 will be added to each AST node that requires that constraint. The builtin is 
-described below.Standard clang diagnostics for unsatisfied constraints will be 
+described below. Standard clang diagnostics for unsatisfied constraints will be 
 used to report any invalid element types. Concepts required will differ depending
 on whether the resource is a typed buffer or a raw buffer. Until concepts are 
 formally supported by HLSL, the concepts and constraints will be expressed 
--- a/proposals/0010-resource-element-type-validation.md
+++ b/proposals/0010-resource-element-type-validation.md
@@ -6,50 +6,41 @@
 * Issues: [#75676](https://github.com/llvm/llvm-project/issues/75676)
 
 ## Introduction
-Resources are often used in HLSL, with various resource element types (RETs).
+Resources are often used in HLSL, with various resource element types.
 
 For example:
 ```
 RWBuffer<float> rwbuf: register(u0);
 ```
-In this code, the RET is `float`, and the resource type is `RWBuffer`.
-All resources can be placed in two categories, raw buffers and typed buffers. 
-Below is a description of all resources and their corresponding categories
-* raw buffers
-  * [Append|Consume|RW|RasterizerOrdered]StructuredBuffer
-  * [RW]ByteAddressBuffer
+In this code, the element type is `float`, and the resource type is `RWBuffer`.
+`RWBuffer`, along with some other buffers and textures, fall under the "typed buffer"
+category. Below is a description of resources that are considered "typed buffers".
 * typed buffers
   * [RW|RasterizerOrdered]Buffer
   * [Feedback]Texture*
 
-There is a distinct set of rules that define valid RETs for raw buffer resources, 
-and a separate set of rules that define valid RETs for typed buffer resources.
+There is a distinct set of rules that define valid element types for typed buffer resources.
 
-RETs for typed buffer resources:
+Element types for typed buffer resources:
 * Are not intangible (e.g., isn't a resource type)
 * Must be vectors or scalars of arithmetic types, not bools nor enums nor arrays
-* The type should be line-vector layout compatible (homogenous, at most 4 elements, and at most 128 bits in size) 
+* Should be a scalar or homogenous vector of a floating-point or integer type, with a maximum of 4 components after translating 64-bit components into pairs of uint32_t components
 
-RETs for raw buffer variants are much less constrained:
-* Are not intangible (e.g., isn't a resource type)
-* All constituent types must be arithmetic types or bools or enums
-
-Resource types are never allowed as RETs (i.e., `RWBuffer<int>` as an RET).
-If someone writes `RWBuffer<MyCustomType>` and MyCustomType is not a 
-valid RET, there should be infrastructure to reject this RET and emit a message 
-explaining why it was rejected as an RET.
+Resource types are never allowed as element types (i.e., `RWBuffer<int>` as an element type).
+If someone writes `RWBuffer<MyCustomType>` and MyCustomType is not a valid element type, 
+there should be infrastructure to reject this element type and emit a message explaining 
+why it was rejected as an element type.
 
 ## Motivation
-Currently, there is an allow list of valid RETs. It must be modified with respect 
-to this spec. Anything that is not a valid RET will be rejected. The allow list isn't
-broad enough, because there is no case where the RET is user-defined for a raw buffer.
-Ideally, a user should be able to determine how any user-defined structure is invalid 
-as an RET. Some system should be in place to more completely enforce the rules for 
-valid and invalid RETs, as well as provide useful information on why they are invalid.
+Currently, there is an allow list of valid element types. It must be modified with respect 
+to this spec. Anything that is not a valid element type will be rejected. The allow list isn't
+broad enough, because user-defined types aren't allowed.
+Ideally, a user should be able to determine exactly how any user-defined type is invalid 
+as an element type. Some system should be in place to more completely enforce the rules for 
+valid and invalid element types, as well as provide useful information on why they are invalid.
 
-For example, when targeting DXIL IR, `RWBuffer<double4> b : register(u4);` will emit
-an error in DXC, but will not in clang-dxc, despite the fact that `double4` is an 
-invalid RET for typed buffers.
+For example, `RWBuffer<double4> b : register(u4);` will emit an error in DXC, but will not in 
+clang-dxc, despite the fact that `double4` is an invalid element type for typed buffers.
 
 ## Proposed solution
 
@@ -57,19 +48,18 @@ The proposed solution is to modify the declaration of each resource declared in
 `clang\lib\Sema\HLSLExternalSemaSource.cpp` and insert into each representative
 AST node a concept. The AST node will be created as if the C++20 `concept` keyword
 was parsed and applied to the declaration. The concept will be used to validate the
-given RET, and will emit errors when the given RET is invalid. Although concepts are
-not currently supported in HLSL, we expect support to be added at some point in the
-future. Meanwhile, because LLVM does support concepts, we can make use of
-them when constructing the AST in Sema.
+given element type, and will emit errors when the given element type is invalid. 
+Although concepts are not currently supported in HLSL, we expect support to be 
+added at some point in the future. Meanwhile, because LLVM does support concepts, 
+we can make use of them when constructing the AST in Sema.
 
-A new built-in, `__builtin_hlsl_is_line_vector_layout_compatible`, will be 
-added in order to express the extra typed buffer constraint. This builtin
-will be added to each AST node that requires that constraint. The builtin is 
-described below. Standard clang diagnostics for unsatisfied constraints will be 
-used to report any invalid element types. Concepts required will differ depending
-on whether the resource is a typed buffer or a raw buffer. Until concepts are 
-formally supported by HLSL, the concepts and constraints will be expressed 
-directly in the AST via the HLSL external sema source.
+A new built-in, `__builtin_hlsl_typed_element_compatible`, will be added in order
+to fully express the typed buffer constraint. This builtin will be placed within a
+concept constraint expression that is added to each AST node representing a typed
+buffer resource. The builtin is described below. Standard clang diagnostics for
+unsatisfied constraints will be used to report any invalid element types. Until 
+concepts are formally supported by HLSL, the concepts and constraints will be 
+expressed only in the AST via the HLSL external sema source.
 
 ## Detailed design
 
@@ -77,27 +67,26 @@ In `clang\lib\Sema\HLSLExternalSemaSource.cpp`, `RWBuffer` is defined, along wit
 `RasterizerOrderedBuffer` and `StructuredBuffer`. It is at this point that the 
 concept would be incorporated into these resource declarations. A concept representing
 the relevant constraints will be applied to each resource declaration. If a concept
-is not true for the given RET, a corresponding error message will be emitted.
+is not true for the given element type, a corresponding error message will be emitted.
 
 The list of builtins to be used as type traits that will be available for
 concept definition are described below:
 | type trait | Description|
 |-|-|
-| `__is_intangible` | An RET should be an arithmetic type, bool, enum, or a vector or matrix or UDT containing such types. This is equivalent to validating that the RET is not intangible. This will error when given an incomplete type. |
-| `__builtin_hlsl_is_line_vector_layout_compatible` | A typed buffer RET should never have two different subelement types. Line vector layout compatible also requires at most 4 elements, and a total size of at most 16 bytes. The builtin will also disallow the RET if any of its constituent types are enums or bools. |
+| `!__builtin_hlsl_is_intangible ` | An element type should be an arithmetic type, bool, enum, or a vector or matrix or UDT containing such types. This is equivalent to validating that the element type is not intangible. This will error when given an incomplete type. |
+| `__builtin_hlsl_typed_element_compatible ` | A typed buffer element type should never have two different subelement types. Line vector layout compatible also requires at most 4 elements, and a total size of at most 16 bytes. The builtin will also disallow the element type if any of its constituent types are enums or bools. |
 
-For raw buffers, only `!__is_intangible` needs to be true. 
-For typed buffers, `__builtin_hlsl_is_line_vector_layout_compatible` 
-also needs to be true. This builtin will be used to ensure homogeneity. 
-It will use `BuildFlattenedTypeList` to retrieve a small vector of the subelement types.
-From this subvector, the first element will be compared to all elements in the vector,
-and any mismatches will return false. Typed buffer RETs will 
-also need to have at most 4 subelements, and the total size in bytes cannot exceed 16,
-which will also be verified by `__builtin_hlsl_is_line_vector_layout_compatible`.
-Finally, the builtin will validate that there are no bools or enums present in any 
-component of the type.
+For typed buffers, `__builtin_hlsl_typed_element_compatible` and 
+`!__builtin_hlsl_is_intangible` needs to be true. `__builtin_hlsl_typed_element_compatible`
+will be used to ensure homogeneity. It will use `BuildFlattenedTypeList` to retrieve a small
+vector of the subelement types. From this subvector, the first element will be compared to 
+all elements in the vector, and any mismatches will return false. Typed buffer element types
+will also need to have at most 4 subelements, after splitting 64-bit subelements into two 32-bit
+subelements, which will also be verified by `__builtin_hlsl_is_line_vector_layout_compatible`.
+Finally, the `__builtin_hlsl_typed_element_compatible` will validate that there are no bools or
+enums present in any component of the type.
 
-### Examples of RET validation results:
+### Examples of Element Type validation results:
 ```
 struct oneInt {
     int i;
@@ -152,12 +141,12 @@ struct weirdStruct {
   intVec iv;
 };
 
-RWBuffer<double2> r0; // valid - RET fits in 4 32-bit quantities
+RWBuffer<double2> r0; // valid - element type fits in 4 32-bit quantities
 RWBuffer<int> r1; // valid
 RWBuffer<float> r2; // valid
 RWBuffer<float4> r3; // valid
-RWBuffer<notComplete> r4; // invalid - the RET isn't complete, the definition is missing. 
-// the type trait that would catch this is the negation of `__is_intangible`
+RWBuffer<notComplete> r4; // invalid - the element type isn't complete, the definition is missing. 
+// the type trait that would catch this is the negation of `__builtin_hlsl_is_intangible`
 RWBuffer<oneInt> r5; // valid - all leaf types are valid primitive types, and homogenous
 RWBuffer<oneFloat> r6; // valid
 RWBuffer<twoInt> r7; // valid
@@ -173,11 +162,11 @@ RWBuffer<EightHalves> r12; // invalid, > 4 elements, DXC fails
 StructuredBuffer<EightHalves> r12Structured; // valid
 RWBuffer<oneIntWithVec> r13; // valid
 RWBuffer<weirdStruct> r14; // valid
-RWBuffer<RWBuffer<int> > r15; // invalid - the RET has a handle with unknown size, thus it is an intangible RET.
-// the type trait that would catch this is the negation of `__is_intangible`
+RWBuffer<RWBuffer<int> > r15; // invalid - the element type has a handle with unknown size, 
+// thus it is an intangible element type. The type trait that would catch this is the negation of `__builtin_hlsl_is_intangible`
 ```
 
-Below is a sample C++ implementation of the `RWBuffer` resource type, which is a typed buffer variant.
+Below is a sample C++ implementation of the `RWBuffer` resource type.
 This code would exist within an hlsl header, but concepts are not implemented in HLSL. Instead, the AST node
 associated with RWBuffers is constructed as if this code was read and parsed by the compiler.
 ```
@@ -193,13 +182,6 @@ template<typename element_type> requires !__is_intangible(element_type) && is_va
 struct RWBuffer {
     element_type Val;
 };
-
-// doesn't need __builtin_hlsl_is_line_vector_layout_compatible, because this is a raw buffer
-// also, raw buffers allow bools and enums as constituent types
-template<typename T> requires !__is_intangible(T)
-struct StructuredBuffer {
-    T Val;
-};
 }
 
 ```
@@ -209,9 +191,9 @@ We could instead implement a diagnostic function that checks each of these conce
 one place, either in Sema or CodeGen, but this would prevent us from defining a single header where 
 all resource information is localized.
 
-Another alternative considered was creating a builtin called `__is_valid_resource_element_type`, to
-check all possible valid resource element types, rather than just checking that the RET is not intangible.
-This is unneeded because all primitive non-intangible types are valid RETs.
+Another alternative considered was creating a builtin called `__builtin_hlsl_is_valid_resource_element_type`, to
+check all possible valid resource element types, rather than just checking that the element type is not intangible.
+This is unneeded because all primitive non-intangible types are valid element types.
 
 ## Acknowledgments (Optional)
 * Damyan Pepper
--- a/proposals/0010-resource-element-type-validation.md
+++ b/proposals/0010-resource-element-type-validation.md
@@ -12,11 +12,13 @@ For example:
 ```
 RWBuffer<float> rwbuf: register(u0);
 ```
-In this code, the RET is `float`, and the resource type is `RWBuffer`. The
-resource type is not a `RawBuffer` variant, and so there is a distinct set
-of rules that define valid RETs for this resource type.
+In this code, the RET is `float`, and the resource type is `RWBuffer`.
+There are two types of buffers, `RawBuffer` and `TypedBuffer`. `RWBuffer`
+is a `TypedBuffer` variant, and `StructuredBuffer` is a `RawBuffer` variant.
+There is a distinct set of rules that define valid RETs for `RawBuffer` types, 
+and a separate set of rules that define valid RETs for `StructuredBuffer` types.
 
-RETs for non-`RawBuffer` variants may include:
+RETs for `TypedBuffer` variants may include:
 * basic types: 
   * 16- and 32-bit int and uint
   * half and float
@@ -26,12 +28,12 @@ RETs for non-`RawBuffer` variants may include:
 * user defined types (structs / classes), as long as:
   * all fields in the struct have the same type
   * there are at most 4 sub elements, and each sub element is at most 32 bits
-  
-Resource types are not allowed as RETs (i.e., `RWBuffer<int>` as an RET).
 
-RETs for raw buffer resources are much less constrained:
+RETs for `RawBuffer` variants are much less constrained:
 * it must be a complete type
 * cannot contain a handle or resource type
+
+Resource types are never allowed as RETs (i.e., `RWBuffer<int>` as an RET).
 
 If someone writes `RWBuffer<MyCustomType>` and MyCustomType is not a 
 valid RET, there should be infrastructure to reject this RET and emit a message 
@@ -54,8 +56,8 @@ but will not in clang-dxc, despite the fact that `double4` is an invalid RET.
 The proposed solution is to use some type_traits defined in the std library, create
 some custom type_traits that aren't defined there, and join them together to define a 
 set of conceptual constraints for any RET that is used. These conceptual constraints
-will be applied to every non-`RawBuffer` resource type that is defined, so that all
-non-`RawBuffer` HLSL resources have the same rules about which RETs are valid. 
+will be applied to every `TypedBuffer` resource type that is defined, so that all
+`TypedBuffer` HLSL resources have the same rules about which RETs are valid. 
 Validation will occur upon resource type instantiation. Additionally, certain 
 resource types are `RawBuffer` variants, such as `StructuredBuffer`. Such resource 
 types will have a `[[hlsl::raw_buffer]]` attribute in the attributed type. These
@@ -68,11 +70,11 @@ loosen constraints on viable RETs. Specifically, `__is_homogenous` and
 In `clang\lib\Sema\HLSLExternalSemaSource.cpp`, `RWBuffer` is defined, along with 
 `RasterizerOrderedBuffer` and `StructuredBuffer`. It is at this point that the 
 `type_traits` should be incorporated into these resource declarations. All of the
-non-`RawBuffer` `type_traits` will be applied on each non-`RawBuffer` HLSL resource
+`TypedBuffer` `type_traits` will be applied on each `TypedBuffer` HLSL resource
 type. For every `type_trait` that is not true for the given RET, an associated error
 message will be emitted. 
 
-The list of type_traits that define a valid non-`RawBuffer` RET are described below:
+The list of type_traits that will be available for use are described below:
 | type_trait | Description|
 |-|-|
 | `__is_complete_type` | An RET should either be a complete type, or a user defined type that has been completely defined. |
@@ -80,7 +82,8 @@ The list of type_traits that define a valid non-`RawBuffer` RET are described be
 | `__is_homogenous` | RETs may be basic types (including vector or matrix), but if they are aggregate types, then all underlying basic types should be the same type. |
 | `__is_at_most_four_elements_and_at_most_thirty_two_bits_each` | RETs should not have more than 4 elements, and each element may not exceed 32 bits in size. |
 
-Only `__is_complete_type` and `__is_intangible_type` are needed for `RawBuffer` RETs.
+All `type_traits` above must be true for `TypedBuffer`s, but only `__is_complete_type` 
+and `__is_intangible_type` are needed for `RawBuffer` RETs.
 
 * Examples:
 ```
@@ -112,7 +115,7 @@ RWBuffer<oneInt> r5; // valid - all fields are valid primitive types
 RWBuffer<a> r6; // valid - all leaf types are valid primitive types, and homogenous
 
 // diagnostic: "resource element type 'b' has incomplete definition"
-RWBuffer<b> r7;// invalid - the RET isn't complete, the definition is missing. 
+RWBuffer<b> r7; // invalid - the RET isn't complete, the definition is missing. 
 // the type_trait that would catch this is `__is_complete_type`
 
 // diagnostic: "resource element type 'c' has non-homogenous aggregate type"
--- a/proposals/0010-resource-element-type-validation.md
+++ b/proposals/0010-resource-element-type-validation.md
@@ -13,8 +13,8 @@ For example:
 RWBuffer<float> rwbuf: register(u0);
 ```
 In this code, the RET is `float`, and the resource type is `RWBuffer`.
-There are two types of buffers, raw buffers and typed buffers. Below is a 
-list of all buffers and their corresponding variants
+All resources can be placed in two categories, raw buffers and typed buffers. 
+Below is a description of all resources and their corresponding categories
 * raw buffers
   * [Append|Consume|RW]StructuredBuffer
   * [RW]ByteAddressBuffer
@@ -22,32 +22,19 @@ list of all buffers and their corresponding variants
   * [RW]Buffer
   * [Feedback]Texture*
 
-There is a distinct set of rules that define valid RETs for raw buffer types, 
-and a separate set of rules that define valid RETs for typed buffer types.
-These rules also depend on the target IR, SPIR-V or DXIL.
+There is a distinct set of rules that define valid RETs for raw buffer resources, 
+and a separate set of rules that define valid RETs for typed buffer resources.
 
-RETs for typed buffer variants with a DXIL target IR may include:
-* basic types: 
-  * 16, 32, and 64-bit int and uint
-  * half, float, and double
-* vectors and matrices 
-  * containing 4 elements or fewer
-  * total size may not exceed 128 bits
-* user defined types (structs / classes), as long as:
-  * all fields in the struct have the same type
-  * there are at most 4 sub elements
-  * total size may not exceed 128 bits
+RETs for typed buffer resources:
+* Are not intangible (e.g., isn't a resource type)
+* Must be vectors or scalars of arithmetic types, not bools nor enums nor arrays
+* The type should be line-vector layout compatible (homogenous, at most 4 elements, and at most 128 bits in size) 
 
 RETs for raw buffer variants are much less constrained:
-* it must be a complete type
-* cannot contain a handle (such as resource types)
+* Are not intangible (e.g., isn't a resource type)
+* All constituent types must be arithmetic types or bools or enums
 
 Resource types are never allowed as RETs (i.e., `RWBuffer<int>` as an RET).
-Texture resources conform to the rules for typed buffers.
-If the target is SPIR-V, then only the set of rules for raw buffers apply. Typed buffers
-like `RWBuffer` may have RETs that exceed 128 bits, for example, if the target 
-IR is SPIR-V. The typed buffer rules above are only enforced when the IR target is DXIL.
-
 If someone writes `RWBuffer<MyCustomType>` and MyCustomType is not a 
 valid RET, there should be infrastructure to reject this RET and emit a message 
 explaining why it was rejected as an RET.
@@ -55,7 +42,7 @@ explaining why it was rejected as an RET.
 ## Motivation
 Currently, there is an allow list of valid RETs. It must be modified with respect 
 to this spec. Anything that is not a valid RET will be rejected. The allow list isn't
-broad enough, because it doesn't include the case where the RET is user-defined. 
+broad enough, because there is no case where the RET is user-defined for a raw buffer.
 Ideally, a user should be able to determine how any user-defined structure is invalid 
 as an RET. Some system should be in place to more completely enforce the rules for 
 valid and invalid RETs, as well as provide useful information on why they are invalid.
@@ -66,119 +53,152 @@ invalid RET for typed buffers.
 
 ## Proposed solution
 
-The proposed solution is to use C++20 concepts to validate the element type.
-A new built-in, `__builtin_hlsl_is_line_vector_layout_compatible`, will be added in 
-order to express constraints that can't currently be expressed in pure HLSL code.
-Standard clang diagnostics for unsatisfied constraints will be used to report any 
-invalid element types. Concepts required will differ depending on whether the 
-resource is a typed buffer or a raw buffer. Until concepts are formally supported 
-by HLSL, the concepts and constraints will be expressed directly in the AST via 
-the HLSL external sema source.
+The proposed solution is to modify the declaration of each resource declared in 
+`clang\lib\Sema\HLSLExternalSemaSource.cpp` and insert into each representative
+AST node a concept. The AST node will be created as if the C++20 `concept` keyword
+was parsed and applied to the declaration. The concept will be used to validate the
+given RET, and will emit errors when the given RET is invalid.
+
+Specifically, a new built-in, `__builtin_hlsl_is_line_vector_layout_compatible`, 
+will be added in order to express the extra typed buffer constraint. This builtin
+will be added to each AST node that requires that constraint. The builtin is 
+described below.Standard clang diagnostics for unsatisfied constraints will be 
+used to report any invalid element types. Concepts required will differ depending
+on whether the resource is a typed buffer or a raw buffer. Until concepts are 
+formally supported by HLSL, the concepts and constraints will be expressed 
+directly in the AST via the HLSL external sema source.
 
 ## Detailed design
 
 In `clang\lib\Sema\HLSLExternalSemaSource.cpp`, `RWBuffer` is defined, along with 
 `RasterizerOrderedBuffer` and `StructuredBuffer`. It is at this point that the 
-type traits should be incorporated into these resource declarations. A concept 
-containing the relevant type traits will be applied to each resource declaration,
-and there wil be sufficient context to determine the target IR.
-If DXIL is given, then all of the typed buffer type traits will be applied on each
-typed buffer HLSL resource type. Otherwise, the raw buffer type traits will be 
-applied to each resource type. If a type trait is not true for the given 
-RET, a corresponding error message will be emitted.
+concept would be incorporated into these resource declarations. A concept representing
+the relevant constraints will be applied to each resource declaration. If a concept
+is not true for the given RET, a corresponding error message will be emitted.
 
-The list of type traits that will be available for use are described below:
+The list of builtins to be used as type traits that will be available for
+concept definition are described below:
 | type trait | Description|
 |-|-|
-| `__is_complete_type` | An RET should either be a complete type, or a user defined type that has been completely defined. |
-| `__is_intangible` | An RET should be an arithmetic type, or a bool, or a vector or matrix or UDT containing such types. This is equivalent to validating that the RET is not intangible. |
-| `__builtin_hlsl_is_line_vector_layout_compatible` | A typed buffer RET with the DXIL IR target should never have two different subelement types. Line vector layout compatible also requires at most 4 elements, and a total size of at most 16 bytes |
+| `__is_intangible` | An RET should be an arithmetic type, bool, enum, or a vector or matrix or UDT containing such types. This is equivalent to validating that the RET is not intangible. This will error when given an incomplete type. |
+| `__builtin_hlsl_is_line_vector_layout_compatible` | A typed buffer RET with the DXIL IR target should never have two different subelement types. Line vector layout compatible also requires at most 4 elements, and a total size of at most 16 bytes. |
 
-For the SPIR-V IR target, only `__is_complete_type` and `!__is_intangible` 
-need to be true. When the target IR is DXIL, and the resource is a typed buffer variant,
-`__builtin_hlsl_is_line_vector_layout_compatible` will be used to ensure homogeneity. 
+For raw buffers, only `!__is_intangible` needs to be true. 
+For typed buffers, `__builtin_hlsl_is_line_vector_layout_compatible` 
+also needs to be true. This builtin will be used to ensure homogeneity. 
 It will use `BuildFlattenedTypeList` to retrieve a small vector of the subelement types.
 From this subvector, the first element will be compared to all elements in the vector,
-and any mismatches will return false. Typed buffer RETs with the DXIL IR target will 
+and any mismatches will return false. Typed buffer RETs will 
 also need to have at most 4 subelements, and the total size in bytes cannot exceed 16,
 which will also be verified by `__builtin_hlsl_is_line_vector_layout_compatible`.
+Finally, there will be an additional check that there are no bools or enums present
+in any component of the type.
 
-* Examples:
+* Examples of RET validation results:
 ```
-// targeting DXIL
+
 struct oneInt {
-	int i;
+    int i;
 };
+
 struct twoInt {
    int aa;
    int ab;
 };
-struct a {
-   oneInt bx;
-   int i;
+
+struct threeInts {
+  oneInt o;
+  twoInt t;
 };
-struct b;
-struct c {
-  oneInt ca;
-  float1 cb;
+
+struct oneFloat {
+    float f;
 };
-struct d {
+struct notComplete;
+struct depthDiff {
+  int i;
+  oneInt o;
+  oneFloat f;
+};
+
+struct notHomogenous{     
+  int i;
+  float f;
+};
+
+struct EightElements {
   twoInt x[2];
   twoInt y[2];
 };
+
+struct EightHalves {
+half x[8]; 
+};
+
+struct intVec {
+  int2 i;
+};
+
+struct oneIntWithVec {
+  int i;
+  oneInt i2;
+  int2 i3;
+};
+
+struct weirdStruct {
+  int i;
+  intVec iv;
+};
+
 RWBuffer<double2> r0; // valid - RET fits in 4 32-bit quantities
 RWBuffer<int> r1; // valid
 RWBuffer<float> r2; // valid
 RWBuffer<float4> r3; // valid
-RWBuffer<oneInt> r4; // valid
-RWBuffer<oneInt> r5; // valid - all fields are valid primitive types
-RWBuffer<a> r6; // valid - all leaf types are valid primitive types, and homogenous
-
-RWBuffer<b> r7; // invalid - the RET isn't complete, the definition is missing. 
-// the type trait that would catch this is `__is_complete_type`
-
-RWBuffer<c> r8; // invalid - struct `oneInt` has int types, and this is not homogenous with the float1 contained in `c`. 
-// the type trait that would catch this is __builtin_hlsl_is_line_vector_layout_compatible
-
-StructuredBuffer<c> r8Structured; // valid
-
-RWBuffer<d> r9; // invalid - the struct d exceeds 16 bytes.
+RWBuffer<notComplete> r4; // invalid - the RET isn't complete, the definition is missing. 
+// the type trait that would catch this is the negation of `__is_intangible`
+RWBuffer<oneInt> r5; // valid - all leaf types are valid primitive types, and homogenous
+RWBuffer<oneFloat> r6; // valid
+RWBuffer<twoInt> r7; // valid
+RWBuffer<threeInts> r8; // valid
+RWBuffer<notHomogenous> r9; // invalid, all template type components must have the same type, DXC fails
+StructuredBuffer<notHomogenous> r9Structured; // valid
+RWBuffer<depthDiff> r10; // invalid, all template type components must have the same type, DXC fails
+RWBuffer<EightElements> r11; // invalid, > 4 elements and > 16 bytes, DXC fails 
 // This would be caught by __builtin_hlsl_is_line_vector_layout_compatible
-
-StructuredBuffer<d> r9Structured; // valid
-
-RWBuffer<RWBuffer<int> > r10; // invalid - the RET has a handle with unknown size, thus it is an intangible RET.
-// the type trait that would catch this is `!__is_intangible`
-
-struct EightHalves { half x[8] };  // sizeof(EightHalves) == 16
-RWBuffer<EightHalves> b; // invalid - EightHalves has 8 subelements, which exceeds the limit of 4.
+StructuredBuffer<EightElements> r9Structured; // valid
+RWBuffer<EightHalves> r12; // invalid, > 4 elements, DXC fails
 // This would be caught by __builtin_hlsl_is_line_vector_layout_compatible
+StructuredBuffer<EightHalves> r12Structured; // valid
+RWBuffer<oneIntWithVec> r13; // valid
+RWBuffer<weirdStruct> r14; // valid
+RWBuffer<RWBuffer<int> > r15; // invalid - the RET has a handle with unknown size, thus it is an intangible RET.
+// the type trait that would catch this is the negation of `__is_intangible`
 ```
 
 Below is a sample C++ implementation of the `RWBuffer` resource type, which is a typed buffer variant.
-This code would exist within `HLSLExternalSemaSource.cpp`, as a substitute for the existing definition
-of `RWBuffer`
+This code would exist within an hlsl header, but concepts are not implemented in HLSL. However, the AST node
+associated with RWBuffers is constructed as if this code was read and parsed by the compiler.
 ```
 #include <type_traits>
 
 namespace hlsl {
 
-const bool is_spirv_target = getASTContext().getTargetInfo().getTriple().isSPIRV();
 
 template<typename T>
-concept is_valid_RET_for_typed_buffer = is_spirv_target || __builtin_hlsl_is_line_vector_layout_compatible(T)
 
-template<typename T> requires is_valid_RET_for_typed_buffer<T>
- && __is_complete_type(T) && !__is_intangible(T)
+template<typename T> requires __builtin_hlsl_is_line_vector_layout_compatible(T)
+ && !__is_intangible(T) && !isa<bool> && !isa<enum>
 struct RWBuffer {
     T Val;
 };
 
-// doesn't need is_valid_RET_for_typed_buffer, because this is a raw buffer
-template<typename T> requires __is_complete_type(T) && !__is_intangible(T)
+// doesn't need __builtin_hlsl_is_line_vector_layout_compatible, because this is a raw buffer
+// also, raw buffers allow bools and enums as constituent types
+template<typename T> requires !__is_intangible(T)
 struct StructuredBuffer {
     T Val;
 };
+}
 
 ```
 
@@ -196,4 +216,6 @@ Damyan Pepper
 Chris Bieneman
 Greg Roth
 Sarah Spall
+Tex Riddell
+Justin Bogner
 <!-- {% endraw %} -->
--- a/proposals/0010-resource-element-type-validation.md
+++ b/proposals/0010-resource-element-type-validation.md
@@ -57,10 +57,13 @@ The proposed solution is to modify the declaration of each resource declared in
 `clang\lib\Sema\HLSLExternalSemaSource.cpp` and insert into each representative
 AST node a concept. The AST node will be created as if the C++20 `concept` keyword
 was parsed and applied to the declaration. The concept will be used to validate the
-given RET, and will emit errors when the given RET is invalid.
+given RET, and will emit errors when the given RET is invalid. Although concepts are
+not currently supported in HLSL, we expect support to be added at some point in the
+future. Meanwhile, because LLVM does support concepts, we can make use of
+them when constructing the AST in Sema.
 
-Specifically, a new built-in, `__builtin_hlsl_is_line_vector_layout_compatible`, 
-will be added in order to express the extra typed buffer constraint. This builtin
+A new built-in, `__builtin_hlsl_is_line_vector_layout_compatible`, will be 
+added in order to express the extra typed buffer constraint. This builtin
 will be added to each AST node that requires that constraint. The builtin is 
 described below.Standard clang diagnostics for unsatisfied constraints will be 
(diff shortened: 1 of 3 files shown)


proposals/0010-validating-resource-container-elements.md
```diff
@@ -0,0 +1,142 @@
...
+For example:
+```
+RWBuffer<float> rwbuf: register(u0);
+```
+In this code, the RET is `float`, and the resource type is `RWBuffer`. The
+resource type is not a `RawBuffer` variant, and so there is a distinct set
+of rules that define valid RETs for this resource type.
+
+RETs for non-`RawBuffer` variants may include basic types (ints and uints of sizes 16
+and 32, as well as half, and float). Structs that contain fields of these basic
```
pow3clk, line 23: bool, uint64_t and double are currently allowed as well https://godbolt.org/z/eb7Yh73W6

proposals/0010-validating-resource-container-elements.md
```diff
@@ -0,0 +1,142 @@
...
+types (where all fields in the struct have the same type) may also be RETs. 
+Structs that either have structs as fields or arrays of structs as fields may also be
+allowed, as long as everything can fit in 4 32-bit quantities. Additionally, resource 
+types are not allowed within an RET, even if the underlying resource type has a 
+primitive RET.
+
+RETs for `RawBuffer` variants are much less constrained, the only rule is that the RET
+may not be an incomplete type (a handle type or a resource type).
```
pow3clk, line 31: Can HLSL authors interact with handle types directly?

proposals/0010-validating-resource-container-elements.md
```diff
@@ -0,0 +1,142 @@
...
+
+If someone writes `RWBuffer<MyCustomType>` and MyCustomType is not a 
+valid RET, there there should be infrastructure to reject this RET and emit a message 
+explaining why it was rejected as an RET.
+
+## Motivation
+
+Currently, in `clang\lib\CodeGen\CGHLSLRuntime.cpp`, under `calculateElementType` 
+there is a whitelist of valid RETs.
```
pow3clk, line 40: "allow list" or similar is a preferable term to whitelist.

Any reference to code seems more at home in detailed design. I think the same information can be provided by saying what the current clang compiler will and won't accept.

proposals/0010-validating-resource-container-elements.md
```diff
@@ -0,0 +1,142 @@
...
+RWBuffer<float> rwbuf: register(u0);
+```
+In this code, the RET is `float`, and the resource type is `RWBuffer`. The
+resource type is not a `RawBuffer` variant, and so there is a distinct set
+of rules that define valid RETs for this resource type.
+
+RETs for non-`RawBuffer` variants may include basic types (ints and uints of sizes 16
+and 32, as well as half, and float). Structs that contain fields of these basic
+types (where all fields in the struct have the same type) may also be RETs. 
+Structs that either have structs as fields or arrays of structs as fields may also be
```
pow3clk, line 25: In addition, vectors and matrices of 4 elements or fewer are allowed both in and out of structs. I recognize that the matrix support is not part of this proposal, but it might be worth mentioning as part of the introduction potentially with a parenthetical or something that says it will be addressed separately.

proposals/0010-validating-resource-container-elements.md
```diff
@@ -0,0 +1,142 @@
...
+
+If someone writes `RWBuffer<MyCustomType>` and MyCustomType is not a 
+valid RET, there there should be infrastructure to reject this RET and emit a message 
+explaining why it was rejected as an RET.
+
+## Motivation
+
+Currently, in `clang\lib\CodeGen\CGHLSLRuntime.cpp`, under `calculateElementType` 
+there is a whitelist of valid RETs. 
+Anything that is not an int, uint, nor a floating-point type, will be rejected.
```
pow3clk, line 41: those types within a vector are also currently supported

proposals/0010-validating-resource-container-elements.md
```diff
@@ -0,0 +1,142 @@
...
+
+The proposed solution is to use some type_traits defined in the std library, create
+some custom type_traits that aren't defined there, and join them together to define a 
+set of conceptual constraints for any RET that is used. These conceptual constraints
+will be applied to every non-`RawBuffer` resource type that is defined, so that all
+non-`RawBuffer` HLSL resources have the same rules about which RETs are valid. 
+Validation will occur upon resource type instantiation. Additionally, certain 
+resource types are `RawBuffer` variants, such as `StructuredBuffer`. Such resource 
+types will have a `[[hlsl::raw_buffer]]` attribute in the attributed type. If this is
+detected, the rules for valid RETs will be loosened.
```
pow3clk, line 61: I get from this that there will be one set of type_traits for all that will be the most restrictive, so the typed buffer requirements and then the structured buffers will have some kind of exceptions? Is that simple than giving each its own set of type_traits or have I misunderstood?

From the below, it sounds like there will be two sets of type_traits, but this paragraph makes it sound like there will only be one set of type_traits that will be applied to "any RET".

proposals/0010-validating-resource-container-elements.md
```diff
@@ -0,0 +1,142 @@
...
+there is a whitelist of valid RETs. 
+Anything that is not an int, uint, nor a floating-point type, will be rejected.
+The whitelist isn't broad enough, because it doesn't include the case where the RET 
+is user-defined. Ideally, a user should be able to determine how any user-defined 
+structure is invalid as an RET. Some system should be in place to more completely 
+enforce the rules for valid and invalid RETs, as well as provide useful information
+on why they are invalid.
+
+For example, `RWBuffer<bool> b : register(u4);` will not emit any error in DXC, 
+but will in clang-dxc, despite the fact that `bool` is a valid RET.
```
pow3clk, line 49: Per the link above, bool itself seems to work. Bool vectors do not work in any respect, but that's another matter: https://godbolt.org/z/eb7Yh73W6

proposals/0010-validating-resource-container-elements.md
```diff
@@ -0,0 +1,142 @@
...
+## Detailed design
+
+In `clang\lib\Sema\HLSLExternalSemaSource.cpp`, `RWBuffer` is defined, along with 
+`RasterizerOrderedBuffer` and `StructuredBuffer`. It is at this point that the 
+`type_traits` should be incorporated into these resource declarations. All of the
+non-`RawBuffer` `type_traits` will be applied on each non-`RawBuffer` HLSL resource
+type. For every `type_trait` that is not true for the given RET, an associated error
+message will be emitted.
+
+The list of type_traits that define a valid non-`RawBuffer` RET are descsribed below:
```
pow3clk, line 72: typo
```suggestion
The list of type_traits that define a valid non-`RawBuffer` RET are described below:

```

proposals/0010-validating-resource-container-elements.md
```diff
@@ -0,0 +1,142 @@
...
+| type_trait | Description|
+|-|-|
+| `__is_complete_type` | An RET should either be a complete type, or a user defined type that has been completely defined. |
+| `__is_intangible_type` | An RET should not contain any handles with unknown sizes, i.e., should not be intangible. So, we should assert this type_trait is false. |
+| `__is_homogenous_aggregate` | RETs may be basic types, but if they are aggregate types, then all underlying basic types should be the same type. |
```
pow3clk, line 77: In LLVM parlance, I think "aggregate type" applies to structs and arrays, but not vectors nor matrices. https://github.com/llvm/llvm-project/blob/a1d64626ba16f5128530ac771c6e641b1155184f/llvm/include/llvm/IR/Type.h#L291-L293

So I think we need another term for this.

proposals/0010-validating-resource-container-elements.md
```diff
@@ -0,0 +1,142 @@
...
+| `__is_contained_in_four_groups_of_thirty_two_bits` | RETs should fit in four 32-bit quantities |
```
pow3clk, line 78: I think there are two criteria here that are munged into one. That's how DXC implies it is working as well and it was only relevant when we got sub 32-bit types, but now that we have, there are two separate requirements here. 

* The total number of components must be four or fewer. That's how DXC behaves and we might consider changing it, but I don't think it makes sense to do so now. 
*  The total size must be less than that of 4x32 bits. I think this wording is more confusing than clarifying. I would prefer to state it as 128 bits or if you prefer, 16 bytes. 

The reason for separating these is that 8 uint16s will fit in the size, but is more than 4 components and is rejected by DXC.

proposals/0010-validating-resource-container-elements.md
```diff
@@ -0,0 +1,142 @@
...
+  float4 cb;
+};
+struct e {
+  int a;
+  int b;
+};
+struct f {
+  e x[2];
+  e y[2];
+};
```
pow3clk, line 107: This would be a bit easier to follow if, instead of arbitrary struct names, they described what they contained. So `struct x` could be `struct oneInt` and `struct a` could be `struct twoInt` etc

proposals/0010-validating-resource-container-elements.md
```diff
@@ -0,0 +1,142 @@
...
+RWBuffer<a> r5; // valid - all fields are valid primitive types
+RWBuffer<b> r6; // valid - all fields (the struct) has valid primitive types for all its fields
+
+// diagnostic: "resource element type 'c' has incomplete definition"
+RWBuffer<c> r7;// invalid - the RET isn't complete, the definition is missing. 
+// the type_trait that would catch this is `__is_complete_type`
+
+// diagnostic: "resource element type 'd' has non-homogenous aggregate type"
+RWBuffer<d> r8; // invalid - struct `a` has int types, and this is not homogenous with the float4 contained in `c`. 
+// the type_trait that would catch this is `__is_homogenous_aggregate`
```
pow3clk, line 121: It also has 4 floats and 2 ints, which is more than 4 components and also won't fit in 16 bytes. Might be better to keep the examples at one violation each.

proposals/0010-validating-resource-container-elements.md
```diff
@@ -0,0 +1,142 @@
+* Proposal: [0010](0010-validating-resource-container-elements.md)
```
damyanp, line 1: FWIW: I prefer the title of the PR "Resource element type validation" to the filename "validating resource container elements"

damyanp: Also, I think that this proposal is only about element types for RWBuffer?  Is it going to cover rules for any other resource types?  If not, then the title should probably reflect the limited scope.

proposals/0010-validating-resource-container-elements.md
```diff
@@ -0,0 +1,142 @@
...
+
+## Introduction
+Resources are often used in HLSL, with various resource element types (RETs).
+For example:
+```
+RWBuffer<float> rwbuf: register(u0);
+```
+In this code, the RET is `float`, and the resource type is `RWBuffer`. The
+resource type is not a `RawBuffer` variant, and so there is a distinct set
+of rules that define valid RETs for this resource type.
```
damyanp, line 20: >  The resource type is not a `RawBuffer` variant

I'm a bit lost here.  I think that there's some context about RawBuffer's that I need to have in order to understand this...but...actually, I can't find any documentation online for `RawBuffer`.  Am I missing something?

Maybe this should be `ByteAddressBuffer`?

And possibly the context here is that ByteAddressBuffers can only read/write `uint`s, whereas `RWBuffer` needs to have a specified element type?

damyanp (commented):
Only read through a bit of the introduction so far.

pow3clk (commented):
The descriptions are good and the examples are very helpful!

proposals/0010-validating-resource-container-elements.md
bob80905: > Also, I think that this proposal is only about element types for RWBuffer? Is it going to cover rules for any other resource types? If not, then the title should probably reflect the limited scope.

I intend for this proposal to cover rules for all resource element types used in any resource context.

proposals/0010-validating-resource-container-elements.md
bob80905: I'll add some context in the spec, but rawbuffers are essentially a set of buffers that defer to the shader in interpreting / converting elements in the buffer to be intelligible. Thus, there are fewer constraints on their element types, they don't need to fit in 4 32-bit groups for example. `StructuredBuffer` is an example of a rawbuffer.

proposals/0010-validating-resource-container-elements.md
```diff
@@ -0,0 +1,142 @@
...
+* Status: **Under Consideration**
+* Impacted Project(s): (LLVM)
+
+*During the review process, add the following fields as needed:*
+
+* PRs: [#NNNN](https://github.com/microsoft/DirectXShaderCompiler/pull/NNNN)
+* Issues: [#75676](https://github.com/llvm/llvm-project/issues/75676)
+
+## Introduction
+Resources are often used in HLSL, with various resource element types (RETs).
```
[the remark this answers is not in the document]
bob80905: Yes, completely new as far as I know. Just thought I'd use an acronym since that phrase is littered everywhere in this spec.

proposals/0010-validating-resource-container-elements.md
bob80905: If I understand correctly from my conversation with @tex3d, though the `uint64_t` and `double` cases compile, they will cause undefined behavior at runtime, and we do not want to allow any types that are larger than 32 bits.
Even if 2 64-bit types can technically fit in 4 32-bit fields, it's a hard limit that there may not be more than 4 elements, and each element may not be more than 32 bits.
However, I'm unsure if we can allow `bool`, I think we can.

proposals/0010-validating-resource-container-elements.md
bob80905: It should technically be possible.
I wrote a test in `clang\test\SemaHLSL\resource_binding_attr_error.hlsl` where `__hlsl_resource_t`, the handle type, was directly declared, so it is spellable.
But I don't think we intend users to interact with them.

proposals/0010-validating-resource-container-elements.md
bob80905: Yes, what I intend to communicate is that that `RawBuffer` set of type traits will be applied to those resource types on declaration in HLSLExternalSemaSource.cpp, and the non-`RawBuffer` type traits will be applied to their respective resource types. I'll tweak the wording.

proposals/0010-validating-resource-container-elements.md
bob80905: Hmmm, there must have been a fix over the past several months for there to no longer be any errors. I'll update this statment.

proposals/0010-validating-resource-container-elements.md
bob80905: To your second point, I think the constraint is slightly stronger, each element may not exceed 32 bits. I alluded to the reason earlier. 
Yes, the wording here can be made clearer, I'll adjust.

proposals/0010-validating-resource-container-elements.md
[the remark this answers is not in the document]
damyanp: FWIW, when I see RET I think of this: https://www.felixcloutier.com/x86/ret

proposals/0010-validating-resource-container-elements.md
damyanp: > I intend for this proposal to cover rules for all resource element types used in any resource context.

In this PR?  Currently it only seems to cover buffers and the word "texture" doesn't appear in the document at all.  If the rules for buffers are the same for textures then that's fine, but it'd be good to call that out.

proposals/0010-validating-resource-container-elements.md
damyanp: The thing that's really confusing me here is seeing `RawBuffer` formatted like that - it really suggests that this is the name of something that appears in HLSL code - like `RWBuffer` is.

proposals/0010-resource-element-type-validation.md
```diff
@@ -0,0 +1,136 @@
...
+valid RET, there there should be infrastructure to reject this RET and emit a message 
+explaining why it was rejected as an RET.
+
+## Motivation
+Currently, there is an allow list of valid RETs. It must be modified with respect 
+to this spec. Anything that is not an int, uint, nor a floating-point type, or vectors 
+or matrices containing the aforementioned types, will be rejected. The allow list isn't
+broad enough, because it doesn't include the case where the RET is user-defined. 
+Ideally, a user should be able to determine how any user-defined structure is invalid 
+as an RET. Some system should be in place to more completely enforce the rules for
```
damyanp, line 41: > a user should be able to determine how any user-defined structure is invalid as an RET

Is this a statement about the type of diagnostics we need to emit, or something else?

proposals/0010-resource-element-type-validation.md
```diff
@@ -0,0 +1,136 @@
...
+Resources are often used in HLSL, with various resource element types (RETs).
+For example:
+```
+RWBuffer<float> rwbuf: register(u0);
+```
+In this code, the RET is `float`, and the resource type is `RWBuffer`. The
+resource type is not a `RawBuffer` variant, and so there is a distinct set
+of rules that define valid RETs for this resource type.
+
+RETs for non-`RawBuffer` variants may include basic types (ints and uints of sizes 16
```
damyanp, line 19: Maybe a list of resource types and what category they fall into would be helpful here?

I'm wondering if DXIL.rst needs to be updated to not refer to "RawBuffer" as an HLSL resource type.  Maybe it means ByteAddressBuffer?

DXIL.rst talks about "typed buffers".  Is that the right term for a "not raw buffer"?

proposals/0010-resource-element-type-validation.md
```diff
@@ -0,0 +1,136 @@
...
+and 32, as well as half, float, vectors, and matrices of 4 elements or fewer).
```
damyanp, line 20: > vectors, and matrices 

does the element type matter for vectors and matrices?

proposals/0010-resource-element-type-validation.md
```diff
@@ -0,0 +1,136 @@
...
+Structs that contain fields of these basic types (where all fields in the struct have
+the same type) may also be RETs. 
+Structs that either have structs as fields or arrays of structs as fields may also be
+allowed, as long as there are at most 4 sub elements, and each sub element is at most
+32 bits. Additionally, resource types are not allowed within an RET, even if the 
+underlying resource type has a primitive RET (i.e., `RWBuffer<int>` as an RET).
```
damyanp, line 26: I find these big blocks of prose really hard to read.

```suggestion

RETs for non-`RawBuffer` variants may include:
* basic types: 
  * 16- and 32-bit int and uint
  * half and float
  * vectors
  * matrices containing 4 elements or fewer
* user defined types (structs / classes), as long as:
  * all fields in the struct have the same type
  * there are at most 4 sub elements, and each sub element is at most 32 bits
  
Resource types are not allowed within an RET, even if the underlying resource
type has a primitive RET (i.e., `RWBuffer<int>` as an RET).
```

damyanp, line 26: > even if the underlying resource type has a primitive RET (i.e., `RWBuffer<int>` as an RET).

What is a "primitive RET"?  I think all this sentence has to say is "Resource types are not allowed."  

If it really is important for some reason to say something about the RET of a resource that someone is attempting to use as a RET then maybe "Resource types are not allowed, regardless of that resource's RET."  But I don't think it is.

proposals/0010-resource-element-type-validation.md
```diff
@@ -0,0 +1,136 @@
...
+
+RETs for `RawBuffer` variants are much less constrained, the only rule is that the RET
+may not be an incomplete type (a handle type or a resource type).
```
damyanp, line 29:
```suggestion
RETs for raw buffer resources are much less constrained:
* it must be a complete type
* cannot contain a handle or resource type
```

Bullets to make it easier to read.  And added "cannot contain" to try and capture a struct that has another struct field that contains a resource or handle.

Are we avoiding saying "intangible" here?

proposals/0010-resource-element-type-validation.md
damyanp: I feel that there must be simpler way of explaining this...isn't it just "max of 4 elements and max of 128 bits"?

Maybe some examples could be used to illustrate how that rule applies, but I think there's just one rule?

proposals/0010-resource-element-type-validation.md
damyanp, line 20: DXC accepts double: https://godbolt.org/z/Y19TanY8r

Is that a bug in DXC? It looks like it knows to do two BufferLoad's and to make a double from them.

damyanp (commented):
Reviewed Introduction and Motivation.

proposals/0010-resource-element-type-validation.md
bob80905: Yes, this is a statement highlighting the lack of detailed diagnostics emitted when an invalid UDT is used as an RET.
We want to emit detailed diagnostics, but primarily, emit *some* diagnostic, in the UDT as RET case. It gives motivation for this spec.

proposals/0010-resource-element-type-validation.md
bob80905: Used bulleted list instead.
If I were to use "max of 4 elements and max of 128 bits", a user may think 2 64-bit values would work. This would not work, it would cause undefined behavior at runtime. We must specify that each element may not exceed 32 bits.

bob80905: Simplified.

proposals/0010-validating-resource-container-elements.md
damyanp: Support for 64-bit values seems like a totally reasonable thing to do.  Is there something fundamental about them that mean we can't support them properly, or is this just that DXC is buggy?

What about in SPIR-V today?  Looks like SPIR-V supports `int64_t2`.

proposals/0010-resource-element-type-validation.md
bob80905: Simplified. We could use intangible, I just didn't end up saying it.

proposals/0010-resource-element-type-validation.md
bob80905: @tex3d can confirm, but IIUC, yes it's a bug. This would cause undefined behavior at runtime, and we should disallow it.

proposals/0010-validating-resource-container-elements.md
bob80905: It's something fundamental (If I recall correctly, I think it had to do something with the hardware converter / DXGI format being capped at 32 bits in D3D, and that is not changeable).

proposals/0010-resource-element-type-validation.md
bob80905: I think the full list would be way too long and would overpower this spec. There's a lot of resource types. 
It's strange, I think there are some types in that file that refer to a set of actual spellable types, but of course, RawBuffer isn't actually spellable. I don't think it's a typo.
Perhaps typed buffers is the right term, I'll double check.

proposals/0010-resource-element-type-validation.md
bob80905: I'll add that the element type may not exceed 32 bits.

proposals/0010-validating-resource-container-elements.md
tex3d: I don't think I said anything about them not working or causing undefined behavior.  The undefined behavior case was tied to matching unorm or snorm in HLSL with the DXGI resource type for typed UAVs.

They definitely are supported on DXC (by decomposing to 32-bit components), and FXC the same way IIRC (for double).

tex3d: Perhaps what was thought of was the fact that you can't use these with Sample operations (only Load/Store), since it's equivalent to a typed resource with an int32 component type, which doesn't support those operations.

In fact, if you try it, you get a validation error (we only catch it there), as it attempts to use a 64-bit overload for the sample operation, and define an invalid `%dx.types.ResRet.f64` type.

proposals/0010-resource-element-type-validation.md
```diff
@@ -0,0 +1,160 @@
...
+## Introduction
+Resources are often used in HLSL, with various resource element types (RETs).
+For example:
+```
+RWBuffer<float> rwbuf: register(u0);
+```
+In this code, the RET is `float`, and the resource type is `RWBuffer`.
+There are two types of buffers, RawBuffers and TypedBuffers. `RWBuffer`
+is a TypedBuffer variant, and `StructuredBuffer` is a RawBuffer variant.
+There is a distinct set of rules that define valid RETs for RawBuffer types,
```
spall, line 18: Since StructuredBuffers are RawBuffers I assume you meant to compare RawBuffer to TypedBuffers here?

proposals/0010-resource-element-type-validation.md
```diff
@@ -0,0 +1,160 @@
...
+Resources are often used in HLSL, with various resource element types (RETs).
+For example:
+```
+RWBuffer<float> rwbuf: register(u0);
+```
+In this code, the RET is `float`, and the resource type is `RWBuffer`.
+There are two types of buffers, RawBuffers and TypedBuffers. `RWBuffer`
+is a TypedBuffer variant, and `StructuredBuffer` is a RawBuffer variant.
+There is a distinct set of rules that define valid RETs for RawBuffer types, 
+and a separate set of rules that define valid RETs for `StructuredBuffer` types.
```
spall, line 19:
```suggestion
and a separate set of rules that define valid RETs for TypedBuffer types.

```

proposals/0010-resource-element-type-validation.md
```diff
@@ -12,24 +12,37 @@ For example:
...
-the same type) may also be RETs. 
-Structs that either have structs as fields or arrays of structs as fields may also be
-allowed, as long as there are at most 4 sub elements, and each sub element is at most
-32 bits. Additionally, resource types are not allowed within an RET, even if the 
-underlying resource type has a primitive RET (i.e., `RWBuffer<int>` as an RET).
-
-RETs for `RawBuffer` variants are much less constrained, the only rule is that the RET
-may not be an incomplete type (a handle type or a resource type).
+In this code, the RET is `float`, and the resource type is `RWBuffer`.
+There are two types of buffers, RawBuffers and TypedBuffers. `RWBuffer`
```
damyanp, line 16:
```suggestion
There are two types of buffers, raw buffers and typed buffers. `RWBuffer`

```

(etc.)

proposals/0010-resource-element-type-validation.md
```diff
@@ -0,0 +1,160 @@
...
+
+## Introduction
+Resources are often used in HLSL, with various resource element types (RETs).
+For example:
+```
+RWBuffer<float> rwbuf: register(u0);
+```
+In this code, the RET is `float`, and the resource type is `RWBuffer`.
+There are two types of buffers, RawBuffers and TypedBuffers. `RWBuffer`
+is a TypedBuffer variant, and `StructuredBuffer` is a RawBuffer variant.
```
damyanp, line 17: Given a random buffer type, how do I determine if it is raw or typed?  I really think just having a list of the HLSL data types and whether they're raw or typed would help here.  I don't even think it is long, but obfuscating it in prose doesn't help.

* `[Append|Consume|RW]StructuredBuffer` - raw
* `[RW]Buffer` - typed
* `[RW]ByteAddressBuffer` - raw?
* `[Feedback]Texture*` - typed (I assume Feedback textures are typed)
*

proposals/0010-resource-element-type-validation.md
damyanp: Given a random buffer type, how do I determine if it is raw or typed?  I really think just having a list of the HLSL data types and whether they're raw or typed would help here.  I don't even think it is long, but obfuscating it in prose doesn't help.

* `[Append|Consume|RW]StructuredBuffer` - raw
* `[RW]Buffer` - typed
* `[RW]ByteAddressBuffer` - raw?
* `[Feedback]Texture*` - typed (I assume Feedback textures are typed)

proposals/0010-resource-element-type-validation.md
damyanp: I'm going to mark this as resolved and cover it in a new conversation.

proposals/0010-resource-element-type-validation.md
damyanp: Bulletted list looks much better! I'm going to mark this as resolved and cover the 64-bit value support in a new conversation.

proposals/0010-resource-element-type-validation.md
damyanp: I'm going to mark this as resolved and cover it in a new conversation.

proposals/0010-resource-element-type-validation.md
```diff
@@ -0,0 +1,160 @@
...
+  * containing 4 elements or fewer
+  * total size may not exceed 128 bits
+* user defined types (structs / classes), as long as:
+  * all fields in the struct have the same type
+  * there are at most 4 sub elements
+  * total size may not exceed 128 bits
+
+RETs for RawBuffer variants are much less constrained:
+* it must be a complete type
+* cannot contain a handle or resource type
```
damyanp, line 36: What is the difference between a resource type and a type that contains a handle?  Aren't they the same thing?

```suggestion
* cannot contain a handle (such as resource types)

```

damyanp, line 36: Is this list complete? Are there going to be intangible types that aren't handles?

proposals/0010-resource-element-type-validation.md
```diff
@@ -0,0 +1,160 @@
...
+Validation will occur upon resource type instantiation. Additionally, certain 
+resource types are RawBuffer variants, such as `StructuredBuffer`. These
+resource types will have a different set of type-traits applied, which will
+loosen constraints on viable RETs.
+
+## Detailed design
+
+In `clang\lib\Sema\HLSLExternalSemaSource.cpp`, `RWBuffer` is defined, along with 
+`RasterizerOrderedBuffer` and `StructuredBuffer`. It is at this point that the 
+`type_traits` should be incorporated into these resource declarations. The
```
damyanp, line 77: I don't understand what the plan is for how these type traits are going to be used?  Just having a type trait doesn't automatically make diagnostics get emitted.

Are you planning to add static_asserts to the resource class definitions?  If so, I'd really like to see some examples of how this will be done.

proposals/0010-resource-element-type-validation.md
```diff
@@ -0,0 +1,160 @@
...
+resource types are RawBuffer variants, such as `StructuredBuffer`. These
+resource types will have a different set of type-traits applied, which will
+loosen constraints on viable RETs.
+
+## Detailed design
+
+In `clang\lib\Sema\HLSLExternalSemaSource.cpp`, `RWBuffer` is defined, along with 
+`RasterizerOrderedBuffer` and `StructuredBuffer`. It is at this point that the 
+`type_traits` should be incorporated into these resource declarations. The 
+preprocessor will take responsibility for selecting the right set of type_traits
```
damyanp, line 78: > preprocessor

Is this referring to the C preprocessor or something else?

damyanp (commented):
I'd like to understand more about how you propose using the type_traits.

In addition, I think the document throughout needs to be updated to indicate that double and uint64_t are valid element types.

proposals/0010-resource-element-type-validation.md
```diff
@@ -0,0 +1,160 @@
...
+to check, depending on the target IR. If DXIL is given, then all of the TypedBuffer
+`type_traits` will be applied on each TypedBuffer HLSL resource type. Otherwise, the
+RawBuffer type_traits will be applied to each resource type. For every `type_trait`
+that is not true for the given RET, an associated error message will be emitted.
+
+The list of type_traits that will be available for use are described below:
+| type_trait | Description|
+|-|-|
+| `__is_complete_type` | An RET should either be a complete type, or a user defined type that has been completely defined. |
+| `__is_resource_element_type` | An RET should be an arithmetic type, or a bool, or a vector or matrix or UDT containing such types. |
```
llvm-beanz, line 88: What is different between this and `!__builtin_hlsl_is_intangible`?

proposals/0010-resource-element-type-validation.md
```diff
@@ -0,0 +1,160 @@
...
+| `__builtin_is_homogenous` | A TypedBuffer RET with the DXIL IR target should never have two different subelement types. |
+| `__builtin_is_contained_in_four_groups_of_thirty_two_bits` | A TypedBuffer RET with the DXIL IR target should not have more than 4 elements, and the total size of the RET may not exceed 128 bits. |
```
llvm-beanz, line 90: I feel like we could simplify this down to something like `__builtin_hlsl_is_line_vector_layout_compatible` to mark that a type is layout compatible with a [line-width](https://github.com/microsoft/hlsl-specs/pull/321/files#diff-29269dd46a5c0ae5cce91d3af91fce8d966c04887b38fe30ec54cb37cb28c5ceR300) vector. That would handle requiring homogeneousness, element count, and total size.

proposals/0010-resource-element-type-validation.md
bob80905: I am unsure about our plan for the implementation of RayQuery, it will be intangible but will not be a resource. So, it may not have a handle. It might be feasible that there will exist intangible types that aren't handles.

proposals/0010-resource-element-type-validation.md
bob80905: For the purposes of the single static assert. I don't think that just because an RET is not intangible, it is always allowed as an RET. Testing the affirmative seems to be better for long term. For example, what if HLSL supports strings in the future? We'd want to disallow strings as RETs.

proposals/0010-resource-element-type-validation.md
bob80905: Originally, yes, but I think I can get access to what the target IR is in another way, so I'll rewrite this.

proposals/0010-resource-element-type-validation.md
damyanp: So would it be better to say it cannot contain an intangible type?

proposals/0010-resource-element-type-validation.md
llvm-beanz: I'm not sure I agree about strings, but I'm also not sure how HLSL strings will be implemented, but my guess is that they'll be constant only so an offset and a length is the most I expect to represent the "value" of a string.

Setting aside strings since they aren't designed. As described, `__is_resource_element_type` is equivalent to `!__builtin_hlsl_is_intangible`, so I think we should have one trait not two because it simplifies the language rules.

bob80905: Talked with Chris offline, we both agree that `an arithmetic type, or a bool, or a vector or matrix or UDT containing such types` is exactly equivalent to `!__builtin_hlsl_is_intangible`, so I'll use that existing builtin.

proposals/0010-resource-element-type-validation.md
bob80905: Talked to Chris offline, we agreed `__builtin_hlsl_is_line_vector_layout_compatible` could be simplified by using the size_of builtin and other template tricks to verify vector length and total size in bytes.

proposals/0010-resource-element-type-validation.md
```diff
@@ -31,74 +38,74 @@ RETs for TypedBuffer variants may include:
...
+will be applied to every typed buffer resource type that is defined, so that all
+typed buffer HLSL resources have the same rules about which RETs are valid. 
 Validation will occur upon resource type instantiation. Additionally, certain 
-resource types are RawBuffer variants, such as `StructuredBuffer`. These
-resource types will have a different set of type-traits applied, which will
-loosen constraints on viable RETs.
+resource types are raw buffer variants, such as `StructuredBuffer`. These resource 
+types will have a different set of type-traits applied, which will loosen constraints
+on viable RETs. The type traits will be validated as `concepts`, which is a feature 
+introduced in C++20 to statically validate a type.
```
damyanp, line 78: Are we really planning on adding concepts to HLSL?

proposals/0010-resource-element-type-validation.md
```diff
@@ -184,7 +184,7 @@ struct vector_type_info<T __attribute__((ext_vector_type(N)))> {
   constexpr static unsigned Size = N;
 };
 
-bool is_spirv_target = ...;
+const bool is_spirv_target = getASTContext().getTargetInfo().getTriple().isSPIRV();
```
damyanp, line 187: I'm confused - is this code sample HLSL code, or part of the Clang implementation?

proposals/0010-resource-element-type-validation.md
```diff
@@ -0,0 +1,225 @@
...
+RWBuffer<float4> r3; // valid
+RWBuffer<oneInt> r4; // valid
+RWBuffer<oneInt> r5; // valid - all fields are valid primitive types
+RWBuffer<a> r6; // valid - all leaf types are valid primitive types, and homogenous
+
+// diagnostic: "resource element type 'b' has incomplete definition"
+RWBuffer<b> r7; // invalid - the RET isn't complete, the definition is missing. 
+// the type_trait that would catch this is `__is_complete_type`
+
+// diagnostic: "resource element type 'c' has non-homogenous aggregate type"
```
damyanp, line 145: With the proposed concepts-based implementation do you think you'll be able to get these exact diagnostics?

proposals/0010-resource-element-type-validation.md
bob80905: Further discussion is necessary, though it seems easy to add to HLSL. However, while resource definitions are in HLSLExternalSemaSource.cpp, a C++ file, we can use concepts along with the resource definitions. If we plan to move resource definitions to HLSL files, we will either need to add concepts to HLSL, or implement RET validation some other way. The spec as it stands now specifies a design while resources are still defined in a C++ file though.

proposals/0010-resource-element-type-validation.md
bob80905: It would be part of the clang implementation, C++ code within HLSLExternalSemaSource.cpp. I'll make that clearer.

proposals/0010-resource-element-type-validation.md
bob80905: No, I don't think the diagnostics are any longer controllable.
For example, if an RET is a vector of 8 doubles, we'd expect these diagnostics:

```
<source>:41:5: error: constraints not satisfied for class template 'RWBuffer' [with T = double __attribute__((ext_vector_type(8)))]
   41 |     RWBuffer<vector<double, 8>> Buf;
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~
<source>:29:32: note: because '!is_vector_type<double __attribute__((ext_vector_type(8)))>::value' evaluated to false
   29 | template<typename T> requires (!is_vector_type<T>::value) || (is_spirv_target || is_valid_line_vector<T>) && __is_complete_type(T)
      |                                ^
<source>:29:63: note: and 'is_spirv_target' evaluated to false
   29 | template<typename T> requires (!is_vector_type<T>::value) || (is_spirv_target || is_valid_line_vector<T>) && __is_complete_type(T)
      |                                                               ^
<source>:29:82: note: and 'double __attribute__((ext_vector_type(8)))' (vector of 8 'double' values) does not satisfy 'is_valid_line_vector'
   29 | template<typename T> requires (!is_vector_type<T>::value) || (is_spirv_target || is_valid_line_vector<T>) && __is_complete_type(T)
      |                                                                                  ^
<source>:27:60: note: because 'sizeof(double __attribute__((ext_vector_type(8)))) <= 16' (64 <= 16) evaluated to false
   27 | concept is_valid_line_vector = is_vector_type<T>::value && sizeof(T) <= 16 && vector_type_info<T>::Size <= 4;
      | 
```

I'll remove the expected diagnostics.

proposals/0010-resource-element-type-validation.md
damyanp: Apart from line 185, this looks like what we'd write in hlsl.h, but because it isn't all currently spellable we'd have to implement it in HLSLExternalSemaSource.cpp by constructing AST nodes.

proposals/0010-resource-element-type-validation.md
```diff
@@ -0,0 +1,223 @@
...
+RWBuffer<b> r7; // invalid - the RET isn't complete, the definition is missing. 
+// the type_trait that would catch this is `__is_complete_type`
+
+RWBuffer<c> r8; // invalid - struct `oneInt` has int types, and this is not homogenous with the float1 contained in `c`. 
+// the type_trait that would catch this is `__builtin_is_homogenous`
+
+StructuredBuffer<c> r8Structured; // valid
+
+RWBuffer<d> r9; // invalid - the struct f cannot be grouped into 4 32-bit quantities.
+// no type_trait would catch this, but it would be caught by a concept failure, using the sizeof builtin.
```
damyanp, line 150: I think that this is invalid just because `sizeof(d) > 16`, nothing to do with whether or not it can be grouped into 4 32-bit values.

proposals/0010-resource-element-type-validation.md
```diff
@@ -0,0 +1,223 @@
...
+
+StructuredBuffer<d> r9Structured; // valid
+
+RWBuffer<RWBuffer<int> > r10; // invalid - the RET has a handle with unknown size, thus it is an intangible RET.
+// the type trait that would catch this is `!__is_intangible`
+```
```
damyanp, line 156: How about:

```
struct EightHalves { half x[8] };  // sizeof(EightHalves) == 16
RWBuffer<EightHalves> b;
```

I think this should be rejected because it ends up with 8 components, but I don't see any trait proposed that handles element count.

proposals/0010-resource-element-type-validation.md
```diff
@@ -0,0 +1,223 @@
...
+typed buffer HLSL resource type. Otherwise, the raw buffer type_traits will be 
+applied to each resource type. If a `type_trait` is not true for the given 
+RET, a corresponding error message will be emitted.
+
+The list of type_traits that will be available for use are described below:
+| type_trait | Description|
+|-|-|
+| `__is_complete_type` | An RET should either be a complete type, or a user defined type that has been completely defined. |
+| `__is_intangible` | An RET should be an arithmetic type, or a bool, or a vector or matrix or UDT containing such types. This is equivalent to validating that the RET is not intangible. |
+| `__builtin_is_homogenous` | A typed buffer RET with the DXIL IR target should never have two different subelement types. |
```
damyanp, line 97: ```__builtin_is_homogenous``` will have an interesting definition. For example, for it to be useful here, it'll need to return true here:

```
struct Foo {
    int2 x;
};

struct Type {
    Foo f;
    int2 x;
};
```

However, with a very generalized name like this I think that might be confusing for other scenarios.

proposals/0010-resource-element-type-validation.md
```diff
@@ -0,0 +1,223 @@
...
+The proposed solution is to use some type_traits defined in the std library, create
+some custom builtins to use as type_traits, and join them together to define a 
+set of conceptual constraints for any RET that is used. These conceptual constraints
+will be applied to every typed buffer resource type that is defined, so that all
+typed buffer HLSL resources have the same rules about which RETs are valid. 
+Validation will occur upon resource type instantiation. Additionally, certain 
+resource types are raw buffer variants, such as `StructuredBuffer`. These resource 
+types will have a different set of type-traits applied, which will loosen constraints
+on viable RETs. The type traits will be validated as `concepts`, which is a feature 
+introduced in C++20 to statically validate a type.
```
damyanp, line 78: I think this paragraph can be simplified somewhat, I suspect it has evolved into it's current form.

```suggestion
The proposed solution is to use C++20 concepts to validate the element type.
A new built-in, `__builtin_is_homogenous`, will be added in order to express constraints
that can't currently be expressed in pure HLSL code. Standard clang diagnostics for 
unsatisfied constrints will be used to report any invalid element types.

Until concepts are formally supported by HLSL, the concepts and constraints 
will be expressed directly in the AST via the HLSL external sema source.
```

proposals/0010-resource-element-type-validation.md
```diff
@@ -0,0 +1,223 @@
...
+
+## Detailed design
+
+In `clang\lib\Sema\HLSLExternalSemaSource.cpp`, `RWBuffer` is defined, along with 
+`RasterizerOrderedBuffer` and `StructuredBuffer`. It is at this point that the 
+`type_traits` should be incorporated into these resource declarations. A concept
```
damyanp, line 84: Apart from the header, there isn't really a thing in the code called `type_traits` so I don't think we should be marking it as code in the markdown.

In almost all cases I think this should be "type traits".

proposals/0010-resource-element-type-validation.md
damyanp: We should have some text in here saying that we're going to rely on the standard diagnostics emitted when constraints aren't satisfied.

proposals/0010-resource-element-type-validation.md
bob80905: There's no trait that handles element count or total RET size, both of those constraints will be verified by template magic and the `sizeof` builtin. I've added that example to further clarify.

proposals/0010-resource-element-type-validation.md
bob80905: Yes, its definition will use a function that's already been defined, that collects all the leaf types and puts them in a small vector.  The first element in this small vector will be compared to each subsequent element, and any mismatches will result in false, true otherwise. 
Other names that could be considered: __builtin_is_leaf_type_homogenous, __builtin_is_sub_element_homogenous?

proposals/0010-resource-element-type-validation.md
bob80905: Well, magic was a way for me to summarize this line of code in the spec:
`concept is_valid_line_vector = sizeof(T) <= 16 && vector_type_info<T>::Size <= 4;`
This is specifically the plan to implement the element count check and the total size in bytes check.
We get the templated type, T, and run sizeof on it, making sure it's at most 16. Also, if the given templated type is a vector, the Size variable extracts the vector element count, and we make sure that it's at most 4.
`is_valid_line_vector` will only be checked when instantiating RWBuffer when the RET is detected to be a vector type.

damyanp: In the EightHalves example, EightHalves isn't a vector, it is a struct.

Also, how about:

```
struct EightHalvesAnotherWay {
   half a[4];
   half b[3];
   half c;
};
```

bob80905: My sample does seem to fail with UDTs, I've made a big update.

proposals/0010-resource-element-type-validation.md
llvm-beanz: We could define such a thing in the HLSL headers by doing something like:

```c
#if __is_target_arch(spirv)
constexpr bool is_spirv_target = true;
#else
constexpr bool is_spirv_target = false;
#endif
```

In the External Sema Source we can just generate the constraints differently based on the target.

proposals/0010-resource-element-type-validation.md
```diff
@@ -0,0 +1,199 @@
...
+and there wil be sufficient context to determine the target IR.
+If DXIL is given, then all of the typed buffer type traits will be applied on each
+typed buffer HLSL resource type. Otherwise, the raw buffer type traits will be 
+applied to each resource type. If a type trait is not true for the given 
+RET, a corresponding error message will be emitted.
+
+The list of type traits that will be available for use are described below:
+| type trait | Description|
+|-|-|
+| `__is_complete_type` | An RET should either be a complete type, or a user defined type that has been completely defined. |
```
llvm-beanz, line 93: We may not actually need `__is_complete_type`. I know that the type needs to be complete, but `__is_intangible` will error if the type is incomplete.

proposals/0010-resource-element-type-validation.md
```diff
@@ -0,0 +1,221 @@
...
+
+## Proposed solution
+
+The proposed solution is to modify the declaration of each resource declared in 
+`clang\lib\Sema\HLSLExternalSemaSource.cpp` and insert into each representative
+AST node a concept. The AST node will be created as if the C++20 `concept` keyword
+was parsed and applied to the declaration. The concept will be used to validate the
+given RET, and will emit errors when the given RET is invalid.
+
+Specifically, a new built-in, `__builtin_hlsl_is_line_vector_layout_compatible`,
```
damyanp, line 62:
```suggestion
A new built-in, `__builtin_hlsl_is_line_vector_layout_compatible`, 

```

proposals/0010-resource-element-type-validation.md
```diff
@@ -0,0 +1,221 @@
...
+an error in DXC, but will not in clang-dxc, despite the fact that `double4` is an 
+invalid RET for typed buffers.
+
+## Proposed solution
+
+The proposed solution is to modify the declaration of each resource declared in 
+`clang\lib\Sema\HLSLExternalSemaSource.cpp` and insert into each representative
+AST node a concept. The AST node will be created as if the C++20 `concept` keyword
+was parsed and applied to the declaration. The concept will be used to validate the
+given RET, and will emit errors when the given RET is invalid.
```
damyanp, line 60: Wonder if something along these lines would help capture some of the discussions that led up to this decision?

> Requirements such as this are well expressed using the C++20 concepts. These aren't currently supported
> in HLSL, but we expect support to be added at some point in the future. However, LLVM does support
> concepts and it is possible to make use of them even if they can't be spelled in HLSL source code. The way
> to do this is to construct the appropriate AST nodes in Sema. In fact, this is already done for other
> non-spellable aspects of resources in `HLSLExternalSemaSource.cpp`, and so we have a natural place to
> add these concepts.

proposals/0010-resource-element-type-validation.md
```diff
@@ -0,0 +1,221 @@
...
+// This would be caught by __builtin_hlsl_is_line_vector_layout_compatible
+StructuredBuffer<EightHalves> r12Structured; // valid
+RWBuffer<oneIntWithVec> r13; // valid
+RWBuffer<weirdStruct> r14; // valid
+RWBuffer<RWBuffer<int> > r15; // invalid - the RET has a handle with unknown size, thus it is an intangible RET.
+// the type trait that would catch this is the negation of `__is_intangible`
+```
+
+Below is a sample C++ implementation of the `RWBuffer` resource type, which is a typed buffer variant.
+This code would exist within an hlsl header, but concepts are not implemented in HLSL. However, the AST node
```
damyanp, line 179:
```suggestion
This code would exist within an hlsl header, but concepts are not implemented in HLSL. Instead, the AST node

```

proposals/0010-resource-element-type-validation.md
```diff
@@ -0,0 +1,221 @@
...
+associated with RWBuffers is constructed as if this code was read and parsed by the compiler.
+```
+#include <type_traits>
+
+namespace hlsl {
+
+
+template<typename T>
```
damyanp, line 187: Looks like something got edited away here.

I expect you'll want to think of this as something like:
```c++
template<typename RET>
concept TypedResourceElementType = 
    __builtin_hlsl_is_line_vector_layout_compatible<RET>() &&
    !std::is_enum_v<RET> && !std::is_same_v<RET, bool>;


template<typename RET>
    requires TypedResourceElementType<RET>
struct Buffer 
{
};
```

(https://godbolt.org/z/MdfYhYqh9)

proposals/0010-resource-element-type-validation.md
```diff
@@ -0,0 +1,221 @@
...
+check all possible valid resource element types, rather than just checking that the RET is not intangible.
+This is unneeded because all primitive non-intangible types are valid RETs.
+
+## Acknowledgments (Optional)
+Damyan Pepper
+Chris Bieneman
+Greg Roth
+Sarah Spall
+Tex Riddell
+Justin Bogner
```
damyanp, line 220:
```suggestion
* Damyan Pepper
* Chris Bieneman
* Greg Roth
* Sarah Spall
* Tex Riddell
* Justin Bogner
```

damyanp (approved):
Some editorial notes, but otherwise LGTM.

proposals/0010-validating-resource-container-elements.md
tex3d: > since it's equivalent to a typed resource with an int32 component type, which doesn't support those operations.

Circling back to this conversation, I recall that this isn't correct anymore - there is more to add.

We added [integer sampling support in SM 6.7](https://microsoft.github.io/DirectX-Specs/d3d/HLSL_SM_6_7_Advanced_Texture_Ops.html#integer-sampling).  This relaxed checks in the front end that disallowed integer types.  So integer types are actually allowed in SM 6.7 on (but not for comparison sampling).

Unfortunately, when 64-bit types are used with sample operations, those resource return types won't be translated into two i32 values per 64-bit value.  That's what leads to the invalid operation type for these caught by validation.  These types could be supported using integer sampling, if we had broken them up as we do for load operations.

I think this means that when using a 64-bit component type in a typed resource element, we should consider this as being implicitly translated to two 32-bit uints per 64-bit component for the actual type used.  Then the rules are applied according to the translated element type.

There is one thing bugging me: double component type seemingly *should* trigger an error in `IsValidObjectElement`, but it doesn't.  See the code here:
https://github.com/microsoft/DirectXShaderCompiler/blob/a023a95f73618e8ca1791147d3722a3e1ea7608f/tools/clang/lib/Sema/SemaHLSL.cpp#L6058

proposals/0010-resource-element-type-validation.md
```diff
@@ -0,0 +1,223 @@
...
+  * [RW|RasterizerOrdered]Buffer
+  * [Feedback]Texture*
+
+There is a distinct set of rules that define valid RETs for raw buffer resources, 
+and a separate set of rules that define valid RETs for typed buffer resources.
+
+RETs for typed buffer resources:
+* Are not intangible (e.g., isn't a resource type)
+* Must be vectors or scalars of arithmetic types, not bools nor enums nor arrays
+* The type should be line-vector layout compatible (homogenous, at most 4 elements, and at most 128 bits in size)
```
tex3d, line 31: I have always had a problem with this terminology.
- It's inventing new terminology not connected to any prior terminology used for the concept to which it applies: "line-vector"?  It's trying to sound more general than it actually is.
- If you were to think about what it could/should be referring to, you could justifiably think it's about what fits into one constant buffer row (line?) in the legacy (current) constant buffer layout.  But that would be a mistake, since the rules are different.  (for example, you can put 8 native 16-bit values or a struct containing non-homogenous types in one row)
- It implies that this type is used in the (memory) layout of the resource itself.  It's not: the element stored and the layout is dependent on runtime factors: the DXGI resource type and device-dependent factors.  The element type in HLSL for a typed resource is just how you want to interpret the values loaded from that up-to-4-component typed resource element.
- The description sounds like it would support non-homogenous types (a struct with several different types that fit should work, as it would for constant buffers)
- This 128-bit restriction sounds equivalent, but hides the true meaning: each 64-bit component in HLSL is translated to 2-32-bit components for the resource element type, with a bitcast for translation.  This has implications for the actual element type used, compatible DXGI type mapping, and allowed operations.  This has a limit of 2 64-bit components because those 2 components expand into a 4 component element type, not because it's 128 bits in size.
- We could probably create specializations that handle 64-bit type translation directly to/from the translated element type in our HLSL header, rather than adding special-case code to the compiler.  In this scenario, the real element type for the resource does not use a 64-bit type as the component type, but uses the translated element type with 32-bit uint components instead.

I would rather a more specific term be used for the more specific typed resource element constraint.

How about "typed-element compatible" and `__builtin_hlsl_typed_element_compatible`?
- A "typed-element" flows from existing terminology: typed resources are textures and buffers with a DXGI type applying to the resource view descriptor at runtime.  These support translation upon load/store to the HLSL element type used in a shader (depending on the types at either end and the operation being performed).
- The element must be a scalar or vector of an integer or floating point type with a maximum of 4 components.  All loads and stores are performed at element granularity, which is a 4 component vector (or struct) in the shader IR, whether or not all those components are read/defined.
- If a 64-bit component type is used in HLSL, each 64-bit value maps to 2-32-bit `uint` components which will be reinterpreted together as the 64-bit value in HLSL.  This has implications for supported operations, and the maximum number of 64-bit components allowed.  `double2` or `[u]int64_t2` becomes `uint4`, which is the maximum number of components, and because the real component used is integer, it triggers those operation restrictions as well.

These above rules and description would apply to this "typed-element compatible"/`__builtin_hlsl_typed_element_compatible` constraint.

The short description could be:
- slightly modified: "at most 4 homogenous scalars, and at most 128 bits in total size" (avoiding using "element" for scalar/component since the whole thing is the element type)
- or "at most 4 homogenous scalars after translation of 64-bit scalars into pairs of uint32_t scalars".  This makes the implied translation clear right in the short constraint description.
- or subsume the scalar or vector and component type rule:
  "scalar or vector of a floating-point or integer type, with a maximum of 4 components after translating 64-bit components into pairs of `uint32_t` components"

There some additional constraints on the element type relating to the way the resource is used (methods called), but these would need to be enforced separately on the method calls:
- If comparison sample (`SampleCmp*`) methods are used, the element type in HLSL must be a scalar floating point type of at most 32-bits.
- If other `Sample` methods are used, integer components are only allowed in the element type when the shader model is 6.7 or greater.  This is triggered by the use of double components because they break down into 2 `uint32_t` components.
There are additional runtime restrictions for legal type combinations between the shader and the DXGI type, but these cannot be enforced by the compiler when they depend on runtime state.  For instance: typed UAVs using a unorm or snorm DXGI type need to have a matching unorm float or snorm float component type in the HLSL resource element declaration.

proposals/0010-resource-element-type-validation.md
```diff
@@ -0,0 +1,223 @@
+* Proposal: [0010](0010-resource-element-type-validation.md)
+* Author(s): [Joshua Batista](https://github.com/bob80905)
+* Sponsor: Joshua Batista
+* Status: **Under Consideration**
+* Impacted Project(s): (LLVM)
+* Issues: [#75676](https://github.com/llvm/llvm-project/issues/75676)
+
+## Introduction
+Resources are often used in HLSL, with various resource element types (RETs).
```
tex3d, line 9: I'm not a fan of the new acronym "RET", since it collides so readily with an instruction for return from procedure.

I think I'd prefer "resource element type" or "element type" or "resource element", or just "element" for short where the context is sufficient.

tex3d: One thought is to use the term "typed resource element" (maybe "TRE" for short, but I'm not sure it's necessary).  That would be a specific thing with specific constraints, avoiding the collision with the `StructuredBuffer` element type or `ByteAddressBuffer` templated `Load`/`Store` element type where the constraints are different.

proposals/0010-resource-element-type-validation.md
```diff
@@ -0,0 +1,223 @@
...
+
+## Introduction
+Resources are often used in HLSL, with various resource element types (RETs).
+
+For example:
+```
+RWBuffer<float> rwbuf: register(u0);
+```
+In this code, the RET is `float`, and the resource type is `RWBuffer`.
+All resources can be placed in two categories, raw buffers and typed buffers.
```
tex3d, line 16: This document would be much clearer if it focused only on HLSL "typed resource element" type constraints.

Raw buffers don't have the same sort of element type, at least not in the same sense.  In fact, strictly speaking "raw buffers" are ByteAddressBuffers, which don't have an element type at all (are treated as raw bits).  You can either use the ordinary Load[2-4]/Store methods which operate on uint1-4 granularity, or use the templated versions which allow you to specify your current interpretation of bits directly for an operation.  That's still not the "element type" of the resource.

For structured buffers, these do have a structure type, also known as an element type, but again, not in the same sense as for typed resources.

I think it's more confusing to try to describe both side-by-side with the same element terminology throughout the same document.

The constraints applied to the raw/structured buffer accesses are actually more broadly applicable than a "resource element type".  For instance, the same rules apply for raytracing or mesh payload structures, ray tracing attribute structures, groupshared variable types, ConstantBuffer/TextureBuffer template types, and really any structure that can be used to load or store host memory outside the typed resource case.

tex3d: Not to mention that this does not describe "all resources" as stated.

proposals/0010-resource-element-type-validation.md
damyanp: The naming used here is intended to be consistent with https://github.com/microsoft/hlsl-specs/pull/321. Please comment on that PR if you think we need to adjust the naming there.

proposals/0010-resource-element-type-validation.md
bob80905: I see how typed resource element is more specific, though I'd like to address element types in this spec specifically. How does TRET (typed resource element types) sound? There are many references to this concept so I think an abbreviation is worthwhile.

proposals/0010-validating-resource-container-elements.md
hekota: Technically the `__hlsl_resource_t` type is spellable, but there is not much one can do with it. It needs to be decorated with resource type attributes to signify what kind of resource it is representing, and it will never get automatic binding unless it is embedded in a struct that looks like a resource class (has the handle as a first field named "h").

proposals/0010-resource-element-type-validation.md
```diff
@@ -0,0 +1,223 @@
...
+`RasterizerOrderedBuffer` and `StructuredBuffer`. It is at this point that the 
+concept would be incorporated into these resource declarations. A concept representing
+the relevant constraints will be applied to each resource declaration. If a concept
+is not true for the given RET, a corresponding error message will be emitted.
+
+The list of builtins to be used as type traits that will be available for
+concept definition are described below:
+| type trait | Description|
+|-|-|
+| `__is_intangible` | An RET should be an arithmetic type, bool, enum, or a vector or matrix or UDT containing such types. This is equivalent to validating that the RET is not intangible. This will error when given an incomplete type. |
```
hekota, line 86: The full name is `__builtin_hlsl_is_intangible` and I would probably use list it here with a negation `!__builtin_hlsl_is_intangible`, otherwise it does not match the description on the right.

proposals/0010-resource-element-type-validation.md
hekota: I'm not a fan of the RET acronym either. I agree with Tex that in many places it can be shortened to "element type", "resource element", or just "element" or "type".

proposals/0010-resource-element-type-validation.md
```diff
@@ -0,0 +1,223 @@
...
+* typed buffers
+  * [RW|RasterizerOrdered]Buffer
+  * [Feedback]Texture*
+
+There is a distinct set of rules that define valid RETs for raw buffer resources, 
+and a separate set of rules that define valid RETs for typed buffer resources.
+
+RETs for typed buffer resources:
+* Are not intangible (e.g., isn't a resource type)
+* Must be vectors or scalars of arithmetic types, not bools nor enums nor arrays
```
hekota, line 30: From what I read in this PR comments `bool`s are allowed for typed buffers.

proposals/0010-resource-element-type-validation.md
bob80905: Ok, I'll change the spec so that it makes no mention of raw buffers.

proposals/0010-resource-element-type-validation.md
bob80905: Ok, I'll remove acronyms.

proposals/0010-resource-element-type-validation.md
bob80905: They aren't. Although is_intangible will not return false in the case a bool is detected, the other builtin, which is currently named is line vector layout compatible, will return false if a bool is detected

proposals/0010-resource-element-type-validation.md
damyanp: Is this really necessary?  Do we need to further churn the spec for this?

Maybe "All resources" could be rewritten as "All buffer resources" if there's concern about that, but I don't know if we need to do another big reworking of this document.

proposals/0010-resource-element-type-validation.md
```diff
@@ -6,98 +6,87 @@
...
+
+For typed buffers, `__builtin_hlsl_typed_element_compatible` and 
+`!__builtin_hlsl_is_intangible` needs to be true. `__builtin_hlsl_typed_element_compatible`
+will be used to ensure homogeneity. It will use `BuildFlattenedTypeList` to retrieve a small
+vector of the subelement types. From this subvector, the first element will be compared to 
+all elements in the vector, and any mismatches will return false. Typed buffer element types
+will also need to have at most 4 subelements, after splitting 64-bit subelements into two 32-bit
+subelements, which will also be verified by `__builtin_hlsl_is_line_vector_layout_compatible`.
+Finally, the `__builtin_hlsl_typed_element_compatible` will validate that there are no bools or
+enums present in any component of the type.
```
hekota, line 87: Implementation detail - this should not be included in the spec.

proposals/0010-resource-element-type-validation.md
```diff
@@ -6,98 +6,87 @@
...
-From this subvector, the first element will be compared to all elements in the vector,
-and any mismatches will return false. Typed buffer RETs will 
-also need to have at most 4 subelements, and the total size in bytes cannot exceed 16,
-which will also be verified by `__builtin_hlsl_is_line_vector_layout_compatible`.
-Finally, the builtin will validate that there are no bools or enums present in any 
-component of the type.
-
-### Examples of RET validation results:
+| `!__builtin_hlsl_is_intangible ` | An element type should be an arithmetic type, bool, enum, or a vector or matrix or UDT containing such types. This is equivalent to validating that the element type is not intangible. This will error when given an incomplete type. |
+| `__builtin_hlsl_typed_element_compatible ` | A typed buffer element type should never have two different subelement types. Line vector layout compatible also requires at most 4 elements, and a total size of at most 16 bytes. The builtin will also disallow the element type if any of its constituent types are enums or bools. |
```
hekota, line 77: This still uses the previous term "Line vector layout compatible".

proposals/0010-resource-element-type-validation.md
hekota: Thanks Josh! In my opinion it looks much cleaner now.

proposals/0010-resource-element-type-validation.md
hekota: Yeah, I think the raw buffers should stay in. Their element types need to be validated by `!__builtin_hlsl_is_intangible`.

hekota (approved):
LGTM!

proposals/0010-resource-element-type-validation.md
```diff
@@ -0,0 +1,217 @@
...
+Below is a description of resources that are considered "typed buffers" vs "raw buffers".
+* typed buffers
+  * [RW|RasterizerOrdered]Buffer
+  * [Feedback]Texture*
+* raw buffers
+  * [Append|Consume|RW|RasterizerOrdered]StructuredBuffer
+  * [RW]ByteAddressBuffer
+
+There is a distinct set of rules that define valid element types for typed buffer resources
+and valid element types for raw buffer resources.
```
tex3d, line 27: Nit: I think the language of "element types" applied to "raw buffer resources" is a little problematic.  StructuredBuffer has an element (or struct) type.  ByteAddressBuffer does not have an "element type", but has templated Load/Store operations that take a type.

Perhaps being more explicit could just be the answer:
```suggestion
and valid element types for `*StructuredBuffer` resources and
`*ByteAddressBuffer` Load/Store operations.
```

proposals/0010-resource-element-type-validation.md
```diff
@@ -0,0 +1,217 @@
...
+
+Element types for typed buffer resources:
+* Are not intangible (e.g., isn't a resource type)
+* Must be vectors or scalars of arithmetic types, not bools nor enums nor arrays
+* Should be a scalar or homogenous vector of a floating-point or integer type, with a maximum of 4 components after translating 64-bit components into pairs of uint32_t components
+Element types for raw buffer resources:
+* Are not intangible (e.g., isn't a resource type)
+* All constituent types must be arithmetic types or bools or enums
```
tex3d, line 35: Other than the language nit of "Element types for raw buffer resources", is there anything that's tangible which isn't allowed?  If not I don't think we need the next point.

If we still need this point, does "arithmetic types or bools or enums" sound inclusive of aggregates of such types?  It's not clear to me that it is.

proposals/0010-resource-element-type-validation.md
```diff
@@ -0,0 +1,217 @@
...
+
+Resource types are never allowed as element types (i.e., `RWBuffer<int>` as an element type).
+If someone writes `RWBuffer<MyCustomType>` and MyCustomType is not a valid element type, 
+there should be infrastructure to reject this element type and emit a message explaining 
+why it was rejected as an element type.
```
tex3d, line 40: Isn't the intangible property sufficient, plus saying that user defined types are made intangible when containing intangible types.

proposals/0010-resource-element-type-validation.md
bob80905: I believe we are disallowing arrays. The second point doesn't sound inclusive of this, I'll fix this.

bob80905: I don't think there's anything that's tangible which isn't allowed. I've removed the second point.

proposals/0010-resource-element-type-validation.md
bob80905: This is indeed repeating the same information, but I think it's a simpler, alternative way to convey more information, that draws the reader to a conclusion they may not have come to.

proposals/0010-resource-element-type-validation.md
```diff
@@ -0,0 +1,217 @@
+* Proposal: [0010](0010-resource-element-type-validation.md)
```
llvm-beanz, line 1: nit: I think we already have a proposal 10, so make sure to update the number before merging.

proposals/0010-resource-element-type-validation.md
llvm-beanz: The point of this is to provide a roadmap for implementation, not documentation for users. Simplifying the language and not repeating things should be the goal.

We should be able to say that cannot be intangible types. We don't even need to go into details about what an intangible type is, we can just link to the language spec for that information.

proposals/0010-resource-element-type-validation.md
```diff
@@ -0,0 +1,217 @@
...
+was parsed and applied to the declaration. The concept will be used to validate the
+given element type, and will emit errors when the given element type is invalid. 
+Although concepts are not currently supported in HLSL, we expect support to be 
+added at some point in the future. Meanwhile, because LLVM does support concepts, 
+we can make use of them when constructing the AST in Sema.
+
+Two builtins will be used to validate typed buffer element types. Any resource 
+element type may not be intangible, so the negation of `__builtin_hlsl_is_intangible`
+will be used for both typed and raw buffer resources.
+A new built-in, `__builtin_hlsl_typed_element_compatible`, will be added in order
```
llvm-beanz, line 67: Not to bike shed, but this builtin name seems odd to me because it doesn't really describe what this is an element of. Maybe `__builtin_hlsl_typed_buffer_element_compatible` or `__builtin_hlsl_typed_resource_element_compatible` to add the missing context?

llvm-beanz (approved):
Mostly looks good to me. A few small comments.

proposals/0010-resource-element-type-validation.md
bob80905: Sounds good, I'll go with typed resource.

PR #71, opened 2024-09-18
Add proposal architecture document for compiling shaders out of process

coopp:
This is an initial commit with an out of process design for compiling shaders.  This design will be exposed via a low-level c api that high level api implementations can use to compile.


proposals/NNNN-outofproc-compiler-api-architecture.md
```diff
@@ -0,0 +1,408 @@
...
+
+An effort is underway to bring compilation support into clang for HLSL based
+shaders.  A C style api will be built to enable applications to compile
+a shader from their own processes in addition to being able to launching the
+clang process.
+
+This document is to propose a more detailed architecture for the approved 
+proposal: [0005](0005-inproc-outofproc-compiler-api-support.md).
+
+The c api [0008](0008-c-interface-compiler-library.md) for compiling a shader
```
damyanp, line 23:
```suggestion
The C API [0008](0008-c-interface-compiler-library.md) for compiling a shader
```

I think we should be case sensitive.  `C` and `API` should always be uppercase when writing English text.

proposals/NNNN-outofproc-compiler-api-architecture.md
```diff
@@ -0,0 +1,408 @@
...
+will be implemented using this out of process architecture.
+
```
damyanp, line 25: The "Motivation" section is missing - and I was actually looking for it:

* why are we building an out of proc system?
* what problem does using process pools and things solve?

proposals/NNNN-outofproc-compiler-api-architecture.md
```diff
@@ -0,0 +1,408 @@
...
+## Proposed solution
+
+The architecture for an out of process design will take advantage of systems
+that have multiple processors, or multiple-core processors. A process pool will
+be created that creates a separate compiler process for each available
+processor. For example, if the system has four processors, then four compiler
+processes are created.
+
+The process pool is owned by a singleton object and shared between each api
+instance created in the calling process.
```
damyanp, line 35: "singleton object" is a bit of a implementation detail that I'm not sure belongs here.

```suggestion
Each process that uses the API has a single process pool that is shared between all instances of the API that process creates.,
```

damyanp: Possible alternative: there's a machine-global process pool that is shared between all processes that use the API.

(Might be worth including in the Alternatives Considered section)

proposals/NNNN-outofproc-compiler-api-architecture.md
```diff
@@ -0,0 +1,408 @@
...
+
+A set of low-level apis will be designed that will drive compilation work
+across multiple processes.
```
damyanp, line 38: > A set of low-level APIs will be designed that will drive compilation work across multiple processes.

IMO we should avoid saying things like "will be designed" and instead have the proposal be more definitive. However, I had trouble making a small change to make it that way. 

Since we've discussed this offline I know what this is intended to say, but I'm not convinced that is clear here.  For example, taking that sentence at face value leads to the question: "How does designing a set of low-level APIs drive compilation work across multiple processes?" And that's not what's intended here.

I think my overall suggestion here is that the low/high-level split of the libraries is a pretty fundamental part of your proposal and you should probably start the Proposed Solution section with an explanation of this. 

I don't think you're intending this proposal to be the design of the low-level API, so I think you should just have a hand-wavy description of it and a note that the design on this is out-of-scope of this proposal (in fact, the design should be guided by the requirements discovered while designing the high-level API).

After that you can go into more detail about the high-level API.

proposals/NNNN-outofproc-compiler-api-architecture.md
```diff
@@ -0,0 +1,408 @@
...
+
+Having a separate low-level api for this design enables other tooling clients
+to build their own high-level systems and implement their own policies.
+
+The high-level HLSL compiler c api implementation will be built on top of the
+low-level c apis and implement any required synchronous and asynchronous api
+policy behaviors. The high-level c api call will be blocked either waiting for
+a worker process to become available or an already dispatched work to be
+completed. This behavior depends on the type of call used
+(synchronous/asynchronous).
```
damyanp, line 50:
```suggestion
The high-level API supports both synchronous and asynchronous calling patterns.
```

?

proposals/NNNN-outofproc-compiler-api-architecture.md
```diff
@@ -0,0 +1,408 @@
...
+
+Communication with the process pool occurs in-process to the application
+and the pool's monitoring threads use a named pipe IPC mechanism with a
+protocol to communicate with each monitored clang process.
```
damyanp, line 54: > Communication with the process pool occurs in-process to the application

Maybe I'm not sure what you mean by "the process pool" here, but to me the "process pool" is outside of the current process.

proposals/NNNN-outofproc-compiler-api-architecture.md
```diff
@@ -0,0 +1,408 @@
...
+to build their own high-level systems and implement their own policies.
+
+The high-level HLSL compiler c api implementation will be built on top of the
+low-level c apis and implement any required synchronous and asynchronous api
+policy behaviors. The high-level c api call will be blocked either waiting for
+a worker process to become available or an already dispatched work to be
+completed. This behavior depends on the type of call used
+(synchronous/asynchronous).
+
+Communication with the process pool occurs in-process to the application
```
damyanp, line 52: > application

It might be helpful to define some nomenclature for this.  I think "calling process" has been used previously to mean the same thing as "application" does here?  Or maybe not and these are different things?

proposals/NNNN-outofproc-compiler-api-architecture.md
```diff
@@ -0,0 +1,408 @@
...
+
+## Detailed design
+
+> TBD: Design the low-level apis that support this architecture and include
+code snippets of how they will be called.
+
+### Out-of-process system initialization
+
+On first creation of a low-level compiler api instance, a singleton object is
+created in-process to the caller.
```
damyanp, line 64: > a singleton object

Could mean all sorts of things.  Does it have a type?  Is there a box on the diagram that is this "singleton object".

proposals/NNNN-outofproc-compiler-api-architecture.md
```diff
@@ -0,0 +1,408 @@
...
+
+The singleton object manages all state and request traffic from the calling
+process. The singleton contains an api dispatching system that interfaces with
+a process pool. Each worker process in the pool is monitored by a dedicated
```
damyanp, line 68: > The singleton contains an api dispatching system that interfaces with a process pool.

This is quite different from the design I imagined.  As soon as you have a process pool you've got a whole load of extra policy involved which makes the low-level API quite high level.  I was imagining more that the low-level API is maybe responsible for working with individual clang processes, and is where the RPC protocol is implemented.

Something like: (although trying to mix components and API layering doesn't really work here)

```mermaid
flowchart TD

    subgraph HLAPI["High-level API"]
        ApiBehaviorLogic("Api Behavior Logic<br/>(implements sync/async, packs/unpacks parameters)")
        ApiCallDispatcher{"Api Dispatcher<br/>(find available process and dispatch work)"}

        subgraph CompilerProcessPool["Compiler Process Pool"]
        subgraph MonitorThread1["Monitor Thread (1)"]
            subgraph JsonRPC ["Low-Level API"]
            end
        end
        subgraph MonitorThread2["Monitor Thread (2)"]
            subgraph JsonRPC2 ["Low-level API"]
            end
        end
        end

    end



    subgraph Process2[" "]
        ClangProcess2["clang.exe (2)"]
        db2[("compiled<br/>shader")]
    end

    subgraph Process1[" "]
        ClangProcess1["clang.exe (1)"]
        db[("compiled<br/>shader")]
    end

    

    style Process1 stroke-dasharray:10,10
    style Process2 stroke-dasharray:10,10


    CompilerApi[Compiler Api] -->|"Compile/CompileAsync( )"| ApiBehaviorLogic

    ApiBehaviorLogic <--> |"dispatch and monitor call"| ApiCallDispatcher

    ApiCallDispatcher -->JsonRPC<--> | json-rpc<br/>protocol | ClangProcess1 --> db

    ApiCallDispatcher --> JsonRPC2<-->| json-rpc<br/>protocol | ClangProcess2 --> db2
```

proposals/NNNN-outofproc-compiler-api-architecture.md
```diff
@@ -0,0 +1,408 @@
...
+thread. One thread to one process.
+
+Worker monitoring threads use named pipes as IPC to communicate with their
+workers and a different method (TBD) to communicate with the singleton's
+api dispatching system.
```
damyanp, line 73: I don't think we should have a design that requires a 1:1 mapping between threads in the calling process and the clang processes.

proposals/NNNN-outofproc-compiler-api-architecture.md
```diff
@@ -0,0 +1,408 @@
...
+    ApiCallDispatcher -->JsonRPC<--> | json-rpc<br/>protocol | ClangProcess1 --> db
+
+    ApiCallDispatcher --> JsonRPC2<-->| json-rpc<br/>protocol | ClangProcess2 --> db2
+```
+
+### Calling apis
+A high-level api call starts in the api entrypoint implementation which will
+call into low-level apis and implement sync/async behaviors. The low-level
+c-api is responsible for finding an available worker process (clang.exe) to
+perform the work.  If a worker is not available the high-level api behavior
```
damyanp, line 133: So I'm trying to understand the proposed design here since it is quite different to what I was expecting (see above). I can't quite understand what the difference is between the low- and high-level APIs in this design?

proposals/NNNN-outofproc-compiler-api-architecture.md
```diff
@@ -0,0 +1,408 @@
...
+for and Asynchronous call return immediately giving the caller enough context
+to monitor or cancel the compilation.
+
+The api implementations, dispatcher and process pool all live in-process to
+the calling application.
+
+The dispatching system uses IPC and a message protocol to communicate with
+each worker process.  Each worker process will know how to package up results
+and communicate them back over the IPC mechanism to their monitoring threads
+which will bubble up those results via the low-level apis.
```
damyanp, line 145: So all source files and generated binaries are transferred across this IPC channel?

proposals/NNNN-outofproc-compiler-api-architecture.md
```diff
@@ -0,0 +1,408 @@
...
+Additional params will be used to configure the process startup logic to setup
+the required IPC for communicating back to the thread that launched it.
+
+Using this same module keeps a single shipping binary for all compilation.
+
+## Alternatives considered
+An alternate design could be to reduce the total monitor threads down by
+using one thread to monitor all clang.exe processes. This would centralize all
+rpc traffic into a single place, but could bottleneck performance if 
+operations are waiting on workers to respond.
```
damyanp, line 365: Why was this alternative rejected?

> could bottleneck performance if operations are waiting on workers to respond

This would only happen if we had synchronous communication with the workers.

proposals/NNNN-outofproc-compiler-api-architecture.md
```diff
@@ -0,0 +1,408 @@
...
+This `Request`/`Response` pattern continues until the compilation is completed.
+Compilation is always concluded with a final `Response` closing the
+`Request`/`Response` pair.
+
+```
+REQUEST  - "compile" (include handler support is enabled)
+
+( compiler needs an include, calls IPC wrapped included handler callback )
+
+RESPONSE - "needinclude"
```
damyanp, line 310: This "needinclude" response feels more like clang wanting to call back into the calling process.  Any reason not to model this in that way, having a more consistent REQUEST/RESPONSE matching to a function and a return value?

eg:

```
---> REQUEST  id:1 compile...
<--- REQUEST  id:2 need include...
---> RESPONSE id:2 include result
<--- REQUEST  id:3 need include...
---> RESPONSE id:3 include result
<--- RESPONSE id:1 compilation result
````

Comment (coopp) [reply proposals/NNNN-outofproc-compiler-api-architecture.md]: I like this idea.  I misunderstood the json-rpc protocol on my first read through.  I thought the server was not allowed to send a request/response pair to the client, but I was wrong. That is totally supported.

File: proposals/NNNN-outofproc-compiler-api-architecture.md (hunk 13 above)
Comment (coopp) [reply proposals/NNNN-outofproc-compiler-api-architecture.md]: Yeah.. I changed my mind on this.  I wasn't really thinking through out the dispatcher could be built so a monitor thread for multiple workers is totally doable.

File: proposals/NNNN-outofproc-compiler-api-architecture.md (hunk 12 above)
Comment (coopp) [reply proposals/NNNN-outofproc-compiler-api-architecture.md]: I was thinking that the caller would specify output paths as part of their command line args packed to the compile request.  So clang would just do what it does today with generated binaries etc.

File: proposals/NNNN-outofproc-compiler-api-architecture.md (hunk 11 above)
Comment (coopp) [reply proposals/NNNN-outofproc-compiler-api-architecture.md]: Yes.  I changed this document.  It was going in multiple directions at the same time which was really confusing.  I removed lots of this type of content.

File: proposals/NNNN-outofproc-compiler-api-architecture.md (hunk 10 above)
Comment (coopp) [reply proposals/NNNN-outofproc-compiler-api-architecture.md]: +1  I agree.

File: proposals/NNNN-outofproc-compiler-api-architecture.md (hunk 5 above)
Comment (coopp) [reply proposals/NNNN-outofproc-compiler-api-architecture.md]: The high-level/low-level api thing was really confusing.  I removed this content.

File: proposals/NNNN-outofproc-compiler-api-architecture.md (hunk 4 above)
Comment (coopp) [reply proposals/NNNN-outofproc-compiler-api-architecture.md]: The high-level/low-level api thing was really confusing. I removed this content.

File: proposals/NNNN-outofproc-compiler-api-architecture.md (hunk 3 above)
Comment (coopp) [reply proposals/NNNN-outofproc-compiler-api-architecture.md]: I removed wording around the singleton implementation detail.  I will add your possible alternative suggestion as it is worth a conversation at least.

File: proposals/NNNN-outofproc-compiler-api-architecture.md (hunk 2 above)
Comment (coopp) [reply proposals/NNNN-outofproc-compiler-api-architecture.md:25]: +1  Added.

File: proposals/NNNN-outofproc-compiler-api-architecture.md (hunk 1 above)
Comment (coopp) [reply proposals/NNNN-outofproc-compiler-api-architecture.md]: I think I fixed these up, but I will do another pass.

File: proposals/NNNN-outofproc-compiler-api-architecture.md [hunk 15]
```diff
@@ -0,0 +1,271 @@
...
+
+### Include handler support
+Include handlers add some additional complexity but still keep to the
+`Request`/`Response` pair design.  If an include handler callback is provided
+to the c api, a special proxy implementation of that callback is created and 
+connected to the IPC messaging system and lives in the clang.exe process.
+This proxy callback will be used during compilation when an include is
+required. Calling the proxy callback results in an IPC communication with
+the client's dispatcher in the form of a `Request`/`Response` pair asking for a 
+buffer for the specified include.
```
tex3d, line 244: The current include handler architecture does a round-trip through the include handler for each potential include path constructed.

I think we will likely want to locate this include path search sequence on the client so it does not need to round trip through RPC for each search possibility.  That brings the round-trip frequency down from `includes * locations` to just `includes`.

That would mean capturing the path for the file containing the include in the "need include" request, then cycling through this and the locally tracked include paths in the client process and calling the include handler for each one until a match is found (or not), then issuing the response.

llvm-beanz: I think there are a couple different reasons why custom include handlers are useful:
1) They identify include dependencies.
2) They allow remapping directory structures.
3) They allow treating in-memory objects as files.

For (1), we should discourage using include handlers as the solution and instead have users adopt the compiler-generated dep files.

For (2), clang has a poorly-documented "header maps" feature that we could use, which would radically simplify our implementation needs.

For (3), having the server process doing file IO, defeats the point, so we'll likely need some model where we can communicate buffers of data from the host process to the compiler process. Streaming the data across a named pipe might be an acceptable initial implementation, but we likely need to do some shared memory solution to get decent performance and avoid redundant copies.

llvm-beanz (approved):
I think this is a great start and we should merge it. We'll need to do some more thinking and design work around the details, but this gives us a really solid starting point for further discussion.

coopp:
The reviews in this PR say I can merge these changes.  I have captured the comments in a separate noted document so the great feedack isn't lost and can be incorporated into the design at a later time.

PR #72, opened 2024-09-18
[0010] Dxil data scalarization proposal

farzonl:
This proposal covers:
- all the cases mentioned in: https://github.com/llvm/llvm-project/issues/107920
- The results of our design discussion
- the global multi-d to 1d array flattening cases mentioned in: https://github.com/llvm/llvm-project/issues/89646
- intro, background, and motivation for this work.

### DIFF
--- a/proposals/NNNN-DXIL-Data-Scalarization.md
+++ b/proposals/NNNN-DXIL-Data-Scalarization.md
@@ -53,9 +53,7 @@ and use the same layout transformations for  both `groupshared` and `static`
 vectors. Both layouts are legal so this shouldn't cause any problems. The plan
 will be to create two new pass. The first pass will convert vectors to arrays
 and potentially create multi-dimensional arrays. The second pass will flatten
-these arrays into one dimension. These two passes will run successively and
-potentially share data about which arrays need to be  flattened via pass
-preservation.
+these arrays into one dimension. These two passes will run successively.
 
 While these two pass could have an agnostic order; they will run after the scalarizer
 pass. The pass will also only operate on global data since the only cases not
--- a/proposals/NNNN-DXIL-Data-Scalarization.md
+++ b/proposals/NNNN-DXIL-Data-Scalarization.md
@@ -1,0 +1,56 @@
+# DXIL Data Scalarization
+
+* Proposal: [NNNN](NNNN-DXIL-Data-Scalarization.md)
+* Author(s): [Farzon Lotfi](https://github.com/farzonl)
+* Sponsor: [Farzon Lotfi](https://github.com/farzonl)
+* Status: **Under Consideration**
+
+## Introduction
+
+In Proposal [0009](0009-DXIL-Function-Scalarization.md) we covered scalarization
+of call instructions, and vector operations like math ops, logical ops, 
+bitcasts, loads, and stores. The goal of this proposal is to present a solution
+for scalarizing data structures via data layout transformations. As a note, 
+any case that results in an implicit cbuffer will not be covered by this
+proposal.
+
+## Motivation
+As mentioned in [DXIL.rst](https://github.com/microsoft/DirectXShaderCompiler/blob/main/docs/DXIL.rst#vectors) 
+"HLSL vectors are scalarized" and "Matrices are lowered to vectors". Therefore,
+ we need to be able to scalarize these types in clang DirectX backend. Without
+scalarizing the data structures and call instructions we can't generate legal
+DXIL.
+
+## Background
+In DXC we support five forms of data scalarization. data marked with
+`groupshared`, `static` function scope, `static` global scope, global data,
+and data in function scope. There are then subset behaviors depending on
+if we are dealing with arrays of vectors or vectors. The data 
+scalarization can then be broken down into three cases:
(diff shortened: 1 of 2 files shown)


proposals/NNNN-DXIL-Data-Scalarization.md
```diff
@@ -0,0 +1,62 @@
...
+In Proposal [0009](0009-DXIL-Function-Scalarization.md) we introduced the scalarizer
+pass this solves all the function scope data scalarization for us. What is left
+are the global scope cases. As mentioned in the intro the implicit cbuffer
+cases are not in scope for this work. What that leaves us with are the
+`groupshared` and `static` cases.
+
+To simplify things for upstream this proposal will make a deviation from `DXC`
+and use the same layout transformations for  both `groupshared` and `static`
+vectors. Both layouts are legal so this shouldn't cause any problems. The plan
+will be to create two new pass. The first pass will convert vectors to arrays
```
bogner, line 54:
```suggestion
will be to create two new passes. The first pass will convert vectors to arrays
```

proposals/NNNN-DXIL-Data-Scalarization.md
```diff
@@ -0,0 +1,62 @@
...
+In Proposal [0009](0009-DXIL-Function-Scalarization.md) we covered scalarization
+of call instructions, and vector operations like math ops, logical ops,
+bitcasts, loads, and stores. The goal of this proposal is to present a solution
+for scalarizing data structures via data layout transformations. As a note,
+any case that results in an implicit cbuffer will not be covered by this
+proposal.
+
+## Motivation
+
+As mentioned in [DXIL.rst](https://github.com/microsoft/DirectXShaderCompiler/blob/main/docs/DXIL.rst#vectors)
```
bogner, line 19: If you use an out-of-line link here the paragraph will flow in the text view a little better:
```
As mentioned in [DXIL.rst] etc etc ...

[DXIL.rst]: https://github.com/microsoft/DirectXShaderCompiler/blob/main/docs/DXIL.rst#vectors
```

proposals/NNNN-DXIL-Data-Scalarization.md
```diff
@@ -0,0 +1,62 @@
...
+     `N` by `M` and then a 1d.
+3. cbuffer usage for regular arrays.
+
+## Proposal
+
+In Proposal [0009](0009-DXIL-Function-Scalarization.md) we introduced the scalarizer
+pass this solves all the function scope data scalarization for us. What is left
+are the global scope cases. As mentioned in the intro the implicit cbuffer
+cases are not in scope for this work. What that leaves us with are the
+`groupshared` and `static` cases.
```
bogner, line 49: This whole paragraph is a bit redundant with the intro. It might be sufficient to just reduce this all the way to "We need to implement a solution to handle `groupshared` and `static` globals." or something like that.

bogner (approved):
This looks like it summarizes what we've been discussing fairly clearly, LGTM. I have some minor editorial comments, but you can take or leave them as you see fit.

proposals/NNNN-DXIL-Data-Scalarization.md
```diff
@@ -0,0 +1,62 @@
...
+
+In DXC we support five forms of data scalarization. data marked with
+`groupshared`, `static` function scope, `static` global scope, global data,
+and data in function scope. There are then subset behaviors depending on
+if we are dealing with arrays of vectors or vectors. The data
+scalarization can then be broken down into three cases:
+
+1. `static` scalar layouts
+   * vector of `M` size get scalarized into `M` elements.
+   * an array of `N` size with a vector of `M` will get scalarized into `M`
```
bob80905, line 35: Is this referring to an array of N vectors, each of size M?
If so, then why wouldn't it instead be scalarized into N arrays of M size? Or is that semantically equivalent?
I would think the dimension of the higher-dimensional structure would stay in its place as the outer dimension.

proposals/NNNN-DXIL-Data-Scalarization.md
```diff
@@ -0,0 +1,62 @@
...
+cases are not in scope for this work. What that leaves us with are the
+`groupshared` and `static` cases.
+
+To simplify things for upstream this proposal will make a deviation from `DXC`
+and use the same layout transformations for  both `groupshared` and `static`
+vectors. Both layouts are legal so this shouldn't cause any problems. The plan
+will be to create two new pass. The first pass will convert vectors to arrays
+and potentially create multi-dimensional arrays. The second pass will flatten
+these arrays into one dimension. These two passes will run successively and
+potentially share data about which arrays need to be  flattened via pass
```
bob80905: nit: extra space between be and flattened.
Also, I'm unsure why the two passes need to share data. Wouldn't the second pass simply attempt to flatten all multi-dimensional arrays into one dimension? Why does it need data from the first pass? Are there multidimensional arrays that should be left unflattened?

proposals/NNNN-DXIL-Data-Scalarization.md
farzonl: Thats just how DXC does it. We aren't going to carry forward that implementation detail.

PR #73, opened 2024-09-19
[hlslpm] Add command to update ProjectMilestone fields

damyanp:
The ProjectMilestone field allows us to generate [project boards](https://github.com/orgs/llvm/projects/4/views/20) that slice on project milestone.  

This works by ensuring that all issues tracked by a `category:Project Milestone` issue have the same value for the `ProjectMilestone` field.

While doing this is also detects and warns about issues that are tracked by multiple project milestone issues.

### DIFF
--- a/tools/hlslpm/github.py
+++ b/tools/hlslpm/github.py
@@ -1,20 +1,21 @@
 # Class for working with github's graphql API.
 
-from enum import Enum
-from dataclasses import dataclass, field
+import itertools
 import os
 import sys
-from typing import Any, Generator, List, Optional
+from typing import Any, Generator, List
 import requests
-from datetime import datetime, date
+from datetime import datetime
 import json
 
-from issue import Category, Issue
+from issue import Category, Issue, IssueState
 
 
 class GH:
     def __init__(self):
         self.accessToken = get_pat()
+        self.queries = {}
+        self.projectMilestoneFieldInfo = None
 
     def graphql(self, query: str, variables={}):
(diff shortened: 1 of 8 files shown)


tools/hlslpm/github.py
```diff
@@ -146,6 +236,13 @@ def read_file_relative_to_this_script(filename):
...
+        # Serialize 'issues' list to JSON
+        json_data = json.dumps([i.__dict__ for i in issues])
+        print(json_data)
+
+    if False:
+        r = gh.get_issues_tracked(["I_kwDOMbLzis6Rpmkm", "I_kwDOBITxeM5SPFcF"])
+        print(list(r))
+
+    r = gh.get_project_field_info("ProjectMilestone")
+    print(r)
```
bogner, line 248: This is an interesting comment/debugging style

bogner (approved):
My python's pretty rusty so you'll probably want a second look, but this lgtm.

tools/hlslpm/github.py
damyanp: Yeah, it's helpful to keep these tests around, but I don't want to accidentally run them against a live github.  But I also don't want to spend the time to build a mock github to test against.

damyanp:
@bogner - thanks for reviewing it.  I've pushed an improvement to the reporting of issues in multiple milestones that I didn't think would warrant a full new review.

joaosaffran (approved):

PR #75, opened 2024-10-07
HLSL Initializer List Support

llvm-beanz:
This PR adds a proposal for HLSL initializer list support to align Clang with DXC's behavior while providing a path forward to a better language implementation in the future.

### DIFF
--- a/proposals/xxxx-initializer-lists.md
+++ b/proposals/xxxx-initializer-lists.md
@@ -125,13 +125,13 @@ member accesses with type checking of each conversion as well. Take the
 following code and proposed AST:
 
 ```hlsl
-float4 F = 1.0.xxxx;
-struct A {
-  int A, B;
-  half C, D;
+float4 f = 1.0.xxxx;
+struct Blah {
+  int a, b;
+  half c , d;
 };
 
-A a = {F};
+Blah blah = {f};
 ```
 
 ```
--- a/proposals/xxxx-initializer-lists.md
+++ b/proposals/xxxx-initializer-lists.md
@@ -1,0 +1,169 @@
+<!-- {% raw %} -->
+
+# Initializer Lists
+
+* Proposal: [NNNN](NNNN-initializer-lists.md)
+* Author(s): [Chris Bieneman](https://github.com/llvm-beanz)
+* Status: **Design In Progress**
+
+
+## Introduction
+
+HLSL's initializer lists assume a member-by-member flattened initialization
+sequence. It is significantly different in behavior from C/C++ initializer
+lists, but in a way that may not be apparent to users that they are depending on
+the divergent behavior. This proposal suggests an approach to implementing
+HLSL's initializer lists with a forward-looking perspective.
+
+## Motivation
+
+As an example, the initializer lists in the code below are all valid in HLSL today:
+
+```hlsl
+struct A {
+  int a;
+  double b;
+};
+struct B {
+  A a[2];
+  int c;
+};
+B b = {{1, 1.2}, {2, 2.2}, 3};   // Array elements specified as members
+B b2 = {1, 2, 3, 4, 5};          // each field initialized separately
+B b3 = {{1, {2, 3}}, {4, 5}};    // Completely random grouping of arguments
+int4 i4 = {1,2,3,4};             // valid int4 in C-syntax
+B b4 = {i4, 5};                  // int4 implicitly expanded to 4 arguments
+```
+
+In HLSL, initializer arguments are scalarized, and a member-by-member
+initialization is performed in depth-first order from the initializer list to
+the destination type. All scalarized fields in a destination type must map to a
+scalarized field from an initializer argument, and the argument must be
+convertible to the field type. It is an error if the source or destination type
+have a different number of scalarized components when fully flattened.
+
+
+In C++, initializer lists must represent the structure of the object they are
+initializing, and each initializer in the list must be convertible to the
+corresponding argument, however not all fields must be initialized. If the
+initializer has less arguments than the target has fields, the remaining fields
+are zero initialized. It is an error if an initializer list has too many
+arguments, if an initializer has arguments that cannot be converted to the
+destination type, or if an initializer list has incorrect nested braces.
+
+
+The following examples which are valid in C++ are not valid in HLSL:
+
+```c++
+struct A {
+  int a;
+  double b;
+};
+struct B {
+  A a[2];
+  int c;
+};
+B b = {};
+A a = {0};
+B b2 = {{{0}}, 0};
+B b3 = {{0}, 0};
+```
+
+Given the differences in the language described above it is likely too
+significant of a change to make Clang follow C/C++ behavior without a transition
+period. Since Clang and DXC will have different underlying representations for
+resources and other built-in types, it is likely extremely difficult to make DXC
+and Clang match on a new behavior. It will also be difficult to provide
+migration tooling in DXC. For these reasons it is proposed to solve this problem
+in Clang.
+
+## Proposed solution
+
+This solution is based on an assumption that we will adopt
+[Proposal 0005 - Strict Initializer Lists](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0005-strict-initializer-lists.md)
+for HLSL 202y. The strict initializer lists proposal is motivated by a closer
+alignment with C++, better error handling, and resolving incompatibilities with
+variadic template initializer lists.
+
+To facilitate implementing C++ initializer list behavior while also providing
+compatibility with existing HLSL, this proposal suggests that Clang parse HLSL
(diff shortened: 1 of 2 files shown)


proposals/xxxx-initializer-lists.md
```diff
@@ -0,0 +1,169 @@
...
+```
+
+This AST structure allows the HLSL initializers to be implemented without any
+need to change code generation. In an alternate example the AST can contain
+member accesses with type checking of each conversion as well. Take the
+following code and proposed AST:
+
+```hlsl
+float4 F = 1.0.xxxx;
+struct A {
```
bharadwajy, line 129: Consider renaming `struct A` to avoid potential conflict with the earlier declaration viz.,
```
struct A {
  int a;
  double b;
};
```
specifically for the variable declaration `A a = {F.x, F.y, F.z, F.w};` later in the document.

proposals/xxxx-initializer-lists.md
```diff
@@ -0,0 +1,169 @@
...
+This full AST representation captures each element access and each type
+conversion, which enables complete and accurate diagnostics which DXC cannot
+provide today.
+
+Further, this AST representation provides a full syntax tree that can be printed
+out to a valid C++ initializer with the same behavior. The examples above would
+be rewritten as:
+
+```hlsl
+B b3 = {{{1, 2}, {3, 4}}, 5};
```
bharadwajy, line 162: Does this proposal intend to support zero-initialization of structure members that are omitted as noted in [Proposal 0005 - Strict Initializer Lists](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0005-strict-initializer-lists.md) reference in this one? In other words, is
```
B b3 = {{{1, 2}, {3, 4}}};
```
be considered valid with `b3.c` initialized to `0`?

llvm-beanz: No, this proposal does not change the language, so you cannot omit initializers and expect zero-initialization.

proposals/xxxx-initializer-lists.md
```diff
@@ -0,0 +1,169 @@
...
+};
+B b = {};
+A a = {0};
+B b2 = {{{0}}, 0};
+B b3 = {{0}, 0};
+```
+
+Given the differences in the language described above it is likely too
+significant of a change to make Clang follow C/C++ behavior without a transition
+period. Since Clang and DXC will have different underlying representations for
```
bharadwajy, line 74:
```suggestion
Given the differences in the languages as described above it is likely changes
needed to make Clang follow C/C++ behavior would be significant thereby requiring
a transition period. Since Clang and DXC will have different underlying representations for
```

proposals/xxxx-initializer-lists.md
```diff
@@ -0,0 +1,169 @@
...
+[Proposal 0005 - Strict Initializer Lists](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0005-strict-initializer-lists.md)
+for HLSL 202y. The strict initializer lists proposal is motivated by a closer
+alignment with C++, better error handling, and resolving incompatibilities with
+variadic template initializer lists.
+
+To facilitate implementing C++ initializer list behavior while also providing
+compatibility with existing HLSL, this proposal suggests that Clang parse HLSL
+initialization syntax into valid C++ initializer list ASTs.
+
+This would mean that given an example like:
```
bharadwajy, line 92:
```suggestion
This would mean that given an HLSL example code such as:
```

proposals/xxxx-initializer-lists.md
```diff
@@ -0,0 +1,169 @@
...
+  ImplicitCastExpr float->int
+    ExtVectorElementExpr .y
+      DeclRefExpr F
+  ImplicitCastExpr float->half
+    ExtVectorElementExpr .z
+      DeclRefExpr F
+  ImplicitCastExpr float->half
+    ExtVectorElementExpr .w
+      DeclRefExpr F
+```
```
bharadwajy, line 151: Rename `F` as `f` following changes in commit 690234038d470a147ec4205f027fc367cc042d25

proposals/xxxx-initializer-lists.md
```diff
@@ -0,0 +1,169 @@
...
+conversion, which enables complete and accurate diagnostics which DXC cannot
+provide today.
+
+Further, this AST representation provides a full syntax tree that can be printed
+out to a valid C++ initializer with the same behavior. The examples above would
+be rewritten as:
+
+```hlsl
+B b3 = {{{1, 2}, {3, 4}}, 5};
+A a = {F.x, F.y, F.z, F.w};
```
bharadwajy, line 163: Should `F` be renamed as `f` following changes in commit 690234038d470a147ec4205f027fc367cc042d25?

Given 
```
struct A {
  int a;
  double b;
};
```
is this initilization correct? Should it be `A a = {f.x, f.y}`?

proposals/xxxx-initializer-lists.md
```diff
@@ -0,0 +1,169 @@
...
+};
+B b3 = {{1, {2, 3}}, {4, 5}};    // Completely random grouping of arguments
+```
+
+The initializer list for `b3` would be parsed into an AST something like:
+
+```
+InitListExpr
+  InitListExpr
+    InitListExpr
```
tex3d, line 111: I wonder how source range would be set for these expressions.  Would we just point from the first element to the last element that the expression corresponds to, even though it will cross weird grouping boundaries?

proposals/xxxx-initializer-lists.md
```diff
@@ -0,0 +1,169 @@
...
+  half c , d;
+};
+
+Blah blah = {f};
+```
+
+```
+InitListExpr
+  ImplicitCastExpr float->int
+    ExtVectorElementExpr .x
```
tex3d, line 140: What would we set for source locations for expressions like this?  Just point to `f` for all of them?

tex3d (approved):
I think this is good.  Comments are for curiosity, no resolution expected for merging.

bogner (approved):
This looks good, though as Tex points out I think there are a lot of questions that this opens up about the debug info story. I suspect we'll have to just accept that source locations for these expressions will need to be very limited, since the AST doesn't actually match the source.

proposals/xxxx-initializer-lists.md
llvm-beanz: `InitListExpr` source ranges are basically the beginning of the first sub-expression to the end of the last. The locations could get a bit wonky, but for error reporting they should work well enough.

proposals/xxxx-initializer-lists.md
llvm-beanz: Yea, that's what I was thinking. Basically the source location for each of the implicit access expressions points to the argument they are derived from.

proposals/xxxx-initializer-lists.md
```diff
@@ -0,0 +1,169 @@
+<!-- {% raw %} -->
+
+# Initializer Lists
+
+* Proposal: [NNNN](NNNN-initializer-lists.md)
```
farzonl, line 5: small reminder to:
1. add proposal number, 
2. rename file
3. change status.

proposals/xxxx-initializer-lists.md
```diff
@@ -0,0 +1,169 @@
...
+* Author(s): [Chris Bieneman](https://github.com/llvm-beanz)
+* Status: **Design In Progress**
+
+
+## Introduction
+
+HLSL's initializer lists assume a member-by-member flattened initialization
+sequence. It is significantly different in behavior from C/C++ initializer
+lists, but in a way that may not be apparent to users that they are depending on
+the divergent behavior. This proposal suggests an approach to implementing
```
farzonl, line 15: This line reads weird start with the forward looking  modifier something like this:
```
This proposal offers a forward-looking approach to implementing HLSL's unique initializer behavior.
```

proposals/xxxx-initializer-lists.md
```diff
@@ -0,0 +1,169 @@
...
+period. Since Clang and DXC will have different underlying representations for
+resources and other built-in types, it is likely extremely difficult to make DXC
+and Clang match on a new behavior. It will also be difficult to provide
+migration tooling in DXC. For these reasons it is proposed to solve this problem
+in Clang.
+
+## Proposed solution
+
+This solution is based on an assumption that we will adopt
+[Proposal 0005 - Strict Initializer Lists](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0005-strict-initializer-lists.md)
```
farzonl, line 83: So the rich text looks fine but if you want to keep the column length  looking more consistent in the raw text I saw justin does the below example on his proposals.

```
[TEXT][REF_NAME] 

/// somewhere at the bottom of the file
[REF_NAME]:  URL
```

farzonl (approved):
LGTM

PR #90, opened 2024-10-24
[hlslpm] Add reports

damyanp:
This change adds a "report" command then generates some reports on milestones and workstreams along with some warnings about possibly misconfigured issues.

Also has a quality-of-life change so that issues are output as URLs rather than IDs, making it easier to navigate to them.


coopp (approved):
Looks good to me.  I've gone through this a few times and didn't see anything jumping out to be changed.

llvm-beanz (approved):

PR #94, opened 2024-11-05
Constant buffers design document

hekota:
Design document describing how HLSL constant buffers will be handled in Clang. Contains the design for parsing, codegen a lowering of `cbuffer` declarations, including the default constant buffer `$Globals`. 

Design for the `ConstantBuffer<T>` resource class is still work in progress.


proposals/NNNN-constant-buffers.md
```diff
@@ -0,0 +1,124 @@
...
+
+### Lowering ConstantBuffer to LLVM IR
+
+The result of codegen for `cbuffer` and `ConstantBuffer` should be identical, or at least very close.
+
+### Lowering `tbuffer` and `TextureBuffer` to LLVM IR
+
+These should be lowered to `target("dx.TypedBuffer", ..)`.Detailed design TBD.
+
+### Lowering constant buffer variable access
```
s-perron, line 90: I'm trying to understand how the layout rules for SPIR-V will work. The only real problem is cbuffers. There is a problem because cbuffers have a different layout than other buffer types, and I don't know how that will be represented in llvm-ir.

Here is the example I am thinking about: https://godbolt.org/z/f45r7vEoG.

In llvm-ir, will the load from the cbuffer be a single load of an object of type `S` with the cbuffer layout? Will the store to the structured buffer store an object of type `S` with the structured buffer layout? If the answer is yes to both, does there need to be a conversion from one to the other in between the load and store?

hekota: I have just updated the document. We would like to include the cbuffer layout on the LLVM target type so the backend does not need to know the specific cbuffer layout rules. Alternatively, it could be a type annotation, but we would like to investigate the including in on the target type first. It probably does not answer your question though.

bogner: I think that the conversion from one to the other will look like an address space cast, or something like that. We'll need to make sure we aren't mixing raw memory operations between cbuffer memory and other resource memories, but as long as we have each type of thing in its own address space the boundaries themselves should be clear at least.

proposals/NNNN-constant-buffers.md
```diff
@@ -0,0 +1,132 @@
+
+# Constant buffers
+
+* Proposal: [NNNN](NNNN-constant-buffers.md)
+* Author(s): [Helena Kotas](https://github.com/hekota)
+* Status: **Design In Progress**
+
+## Introduction
+
+Shader inputs usually include a number of constants which are stored in one or more buffer resources in memory with specific packing rules. These resources can be organized into two types of buffers: constant buffers and texture buffers. This document describes design decisions related to constant buffers.
```
damyanp, line 10: nit: we try and format the source md to 80 columns.

proposals/NNNN-constant-buffers.md
```diff
@@ -0,0 +1,132 @@
...
+  - There is a concern that losing the type information could lead to unnecessary copying of values.
+
+- Using type annotations to store cbuffer layout information. Subtypes would have its own layout annotations.
+  - This is something we can fall back to if encoding the cbuffer layout on LLVM target type turns out to be too unwieldy, especially when it comes to encoding the layout of an array of structures.
+
+## Open issues
+- How to encode the cbuffer layout into LLVM target type
+- How to implement `ConstantBuffer` member access
+- Handling of `$Globals` constant buffer
+- Nested `cbuffer` declarations
```
tex3d, line 195: I put together a bit of an example that illustrates how cbuffer works in various interesting cases here:
https://godbolt.org/z/E4cdx7xj8

Same shader with FXC:
https://shader-playground.timjones.io/39c0683735c0cd79b5970b5116765940

proposals/NNNN-constant-buffers.md
```diff
@@ -0,0 +1,132 @@
...
+
+*TBD*
+
+## Alternatives considered
+
+- Generate access to `cbuffer` varibles as memory accesses with the offset based on cbuffer layout and treat cbuffers as one big type-less memory blob in LLVM IR.
+  - There is a concern that losing the type information could lead to unnecessary copying of values.
+
+- Using type annotations to store cbuffer layout information. Subtypes would have its own layout annotations.
+  - This is something we can fall back to if encoding the cbuffer layout on LLVM target type turns out to be too unwieldy, especially when it comes to encoding the layout of an array of structures.
```
damyanp: I'd be interested to know why the proposed solution was chosen rather than the type annotation one.  What's the downside to using type annotations?

proposals/NNNN-constant-buffers.md
```diff
@@ -0,0 +1,132 @@
...
+
+One possibility is compressing the list of offsets into a smaller number of integers - taking advantage of the fact that outside of `packoffset` use the difference between two adjancent offsets is never more than 16. The compression could also include repetition construct that would help with encoding of array offset. But, as with any compressions, there's always the chance of degenerate cases that will end up with the compressed shape being the same size or larger than the original.
+
+> Note: The most tricky part of the layout encoding are probably arrays of structures because the structure-specific layout gets repeated many times and might not be easy to compress. One idea how to solve this could be translating structures embedded in `cbuffer` into separate target types with their own encoded layout and including them in the `cbuffer` target type. It is not clear though if this is possible (probably yes) or if it would actually make things easier or not.
```
tex3d: I don't think we should be encoding the offset of each element of arrays, nor should we be encoding the offsets of inner struct members for each instance of them in the outer struct.  I think we need to associate type information, containing member offsets, with struct types, like we use for type annotations in DXIL.  This is a lot more efficient, only encoding the offsets we need once.  Then we need some way to connect the target type with the annotated type, or the annotation metadata itself.  Perhaps we can have cbuffer type metadata which references the target type and the contained type, then lists member info, including these offsets.  Other member info will be useful/necessary, such as matrix identification and orientation information.

proposals/NNNN-constant-buffers.md
s-perron: @hekota Thanks. That answers part of my question and was in line with what I was thinking. The layout rules are a language feature. This is consistent with what I see in C/C++. The basic types are encoded in the [data layout](https://llvm.org/docs/LangRef.html#data-layout), and then the layout for the structs are set by being explicit with the padding in the struct.

@bogner I'll wait to see what you do. I don't see how the address space cast will work because that works on pointer, but we need to cast after the load. We also want to make sure that whatever we do, it can be easily optimized. In SPIRV there is an [OpCopyLogical](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#OpCopyLogical) instruction to do the cast, when it cannot be optimized away.

proposals/NNNN-constant-buffers.md
hekota: Thanks Tex! I've created an issue to track this: https://github.com/llvm/llvm-project/issues/118524

bogner (approved):

proposals/0015-constant-buffers.md
```diff
@@ -0,0 +1,380 @@
...
+@array = external addrspace(2) global [10 x i32]
+```
+
+Clang codegen will translate accesses to these global constants to `load`
+instruction from a pointer in `addrspace(2)`. These `load` instructions will be
+later replaced in an LLVM pass with constant buffer load intrinsic calls on a
+buffer resource handle. In order for the pass to generate the correct CBV loads
+it is going to need additinal information, such as which constants belong to
+which constant buffer, and layout of the buffer and any embedded structs.
+Codegen will generate this information as metadata.
```
s-perron, line 272: Is extra metadata requires for the offsets? Can't the layout struct be constructed so that the offsets can be obtained from it? We can add padding elements where needed.

In the SPIR-V backend, we will have to add offset decorations for each member in the constant buffer. I'd like to handle all resource types (ie StructuredBuffers) in the same way. Having a bespoke metadata to get the offsets will create special cases.

proposals/0015-constant-buffers.md
```diff
@@ -0,0 +1,380 @@
...
+
+A new pass `HLSLConstantAccess` will translate all `load` instructions in
+`hlsl_constant` address space to `llvm.{dx|spv}.resource.load.cbuffer`
+intrinsics. It will make use of the metadata generated by Clang codegen that
+maps the constants global variables to individual constant buffers and specifies
+the constant buffer layout. The pass will also transform related `getelementptr`
+instructions to use the constant buffer layout offsets.
+
+After the `HLSLConstantAccess` pass completes the constant globals and the
+constant buffer and layout metadata are no longer needed and should be removed.
```
s-perron, line 333: This is not true. The SPIR-V needs to know the offsets to generate valid code. I've mentioned this above.

proposals/0015-constant-buffers.md
```diff
@@ -0,0 +1,380 @@
...
+information that is not included here is the stride of the array elements. This
+can can be calculated as the size of the array element aligned to 16 (size of
+constant buffer row) because each array element always starts at the begining of
+a row.
+
+### Lowering to buffer load intrinsics
+
+A new pass `HLSLConstantAccess` will translate all `load` instructions in
+`hlsl_constant` address space to `llvm.{dx|spv}.resource.load.cbuffer`
+intrinsics. It will make use of the metadata generated by Clang codegen that
```
s-perron, line 327: What is the signature of this intrinsic? You can see how we generate code at https://godbolt.org/z/eq9E59WcM.

We need to the index of the member to generate the OpAccessChain. If the intrinsic contains the byte offset, then we will have to be able to translate the byte offset to member.

The ideal code pattern for SPIR-V is to access the CBuffers the same way we would access a StructuredBuffer. The only different would be the type for the handle. The index on the `resource_getpointer` would be the index of the member being accessed.

As I've mentioned in another comment, we would want the method of getting the offset for members of structuredBuffers and cbuffers to be the same.

proposals/0015-constant-buffers.md
bogner: I don't love the idea of the layout structure having explicit padding elements - messing with the number of elements is confusing at best. We also do need to have the offsets represented somehow, because if packoffset is used we can't rely on the standard algorithm for cbuffer.

So if we have something like:
```hlsl
cbuffer cb0 {
  int a;
  int b : packoffset(c2.y);
}
```

We want to encode that element 0 is at offset 0 and element 1 is at offset 36. As you mention elsewhere, in SPIR-V we'll need this later to decorate the members:
```
OpMemberDecorate %type_cb0 0 Offset 0
OpMemberDecorate %type_cb0 1 Offset 36
```

I also don't love that we're encoding this separately in metadata and would prefer to include this in the target extension type instead, but most of the obvious schemes to do that have downsides when it comes to clarity and verbosity.

s-perron: > I'm not very comfortable with the idea of the layout structure having explicit padding elements - messing with the number of elements is confusing at best.

I thought that was a standard llvm-ir solution to problem like that: https://godbolt.org/z/7q9avvdfr.

> We also do need to have the offsets represented somehow, because if packoffset is used we can't rely on the standard algorithm for cbuffer.

I don't see why the standard algorithm does not work with appropriate padding: `type { i32, [32 x i8], i32}`. If we don't want to add the padding, then I agree we need something else.

I'm open to other solutions. The important part to me is that the layout for structured buffers, templated load/store for byte address buffers, and cbuffer be represented the same way. For Vulkan, we have different layout options, and we need to be able to represent all of them.

bogner: I suppose we could assume that any array of `i8` is padding, but otherwise the problem with `type { i32, [32 x i8], i32 }` is that the indices get kind of funny. Here we need to know that the third element of the struct (ie, index 2) should be referred to by index 1, as if it's the second element.

s-perron: The main disadvantage of adding extra members for padding is that it become impossible to tell of the user added them or the compiler, and they could pollute the DXIL and SPIR-V. Reflection would want to be able to give the offset of each of the original members from the HLSL, but there is no way to filter out the padding from other unused members.

s-perron: Note that vulkan has the `vk::offset()` attribute, which can be used for resource types other than cbuffers. This is why we have issues for all resource types.

proposals/0015-constant-buffers.md
```diff
@@ -0,0 +1,380 @@
...
+### Layout Structure
+
+The `cbuffer` block can contain any declaration that can occur at a global
+scope, but not all of the declarations correspond to data in the CBV and
+contribute to the buffer layout. As part of the semantic analysis the
+declarations in `cbuffer` scope will be processed into a layout struct that will
+represent the actual content of the constant buffer.
+
+The layout struct will contain all declaration from the `cbuffer` block except:
+- static and groupshared variable declarations
```
bob80905, line 113: nit: I think this should be `or` right?

proposals/0015-constant-buffers.md
```diff
@@ -0,0 +1,380 @@
...
+- resource classes
+- empty structs
+- zero-sized arrays
+- any non-variable declarations (functions, classes, ...)
+
+If the constant buffer includes a struct variable, this struct it will also need
```
bob80905, line 119: nit: this struct / it

proposals/0015-constant-buffers.md
```diff
@@ -0,0 +1,380 @@
...
+  int array[10];
+};
+```
+
+Clang codegen will create these struct definition and global variabless:
+
+```
+$struct.S = type { float }
+%struct.__layout_Constants = type { i32, %struct.S }
+%struct.__layout_MyConstants = type { <4 x float>, [10 x i32] }
```
bob80905, line 255: This should be `OtherConstants`

bob80905 (approved):

PR #95, opened 2024-11-07
[0011] Remove struct test cases, clarify structs aren't allowed

bob80905:
The spec didn't properly reflect that typed buffers should never allow struct element types. This PR adds extra clarification and removes many of the struct test cases. Someone reading the spec should not be able to determine that structs are allowed as typed resource element types.

### DIFF
--- a/proposals/0011-resource-element-type-validation.md
+++ b/proposals/0011-resource-element-type-validation.md
@@ -28,7 +28,7 @@ and valid element types for `*StructuredBuffer` resources and `*ByteAddressBuffe
 
 Element types for typed buffer resources:
 * Are not intangible (e.g., isn't a resource type)
-* Must be vectors or scalars of arithmetic types, not bools nor enums nor arrays
+* Must be vectors or scalars of arithmetic types, not bools nor enums nor arrays nor structs
 * Should be a scalar or homogenous vector of a floating-point or integer type, with a maximum of 4 components after translating 64-bit components into pairs of uint32_t components
 Element types for raw buffer resources:
 * Are not intangible (e.g., isn't a resource type)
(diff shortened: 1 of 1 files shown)


damyanp (approved):

proposals/0011-resource-element-type-validation.md
```diff
@@ -28,7 +28,7 @@ and valid element types for `*StructuredBuffer` resources and `*ByteAddressBuffe
 
 Element types for typed buffer resources:
 * Are not intangible (e.g., isn't a resource type)
-* Must be vectors or scalars of arithmetic types, not bools nor enums nor arrays
+* Must be vectors or scalars of arithmetic types, not bools nor enums nor arrays nor structs
 * Should be a scalar or homogenous vector of a floating-point or integer type, with a maximum of 4 components after translating 64-bit components into pairs of uint32_t components
 Element types for raw buffer resources:
 * Are not intangible (e.g., isn't a resource type)
```
bogner, line 34: This "Are not intangible" bullet seems to be repeated twice. Mind cleaning that up while you're here?

bogner (approved):

proposals/0011-resource-element-type-validation.md
bob80905: It's intended to be there twice, I am trying to express that both typed buffer and raw buffer resource element types should not be intangible.

bogner: Ah. I think there's a formatting issue here:

image

You probably need a new line between the "typed buffer" bullet points and the "raw buffer" text.

bob80905: You're right, I'll edit that!

PR #97, opened 2024-11-08
[SPIR-V] Add proposal for I/O builtin in Clang

Keenuts:
Initial proposal to implement Input/Output built ins with both semantic & inline SPIR-V


s-perron (commented):
I believe that there should be a really high bar for changing the spec in a way that is not backwards compatible. For now, I don't think we should change the spec. However, I do not know the clang code very well. If others who know it better say that changing the spec is necessary, then I will go along with it.

proposals/NNNN-spirv-input-builtin.md
```diff
@@ -0,0 +1,159 @@
...
+ways:
+aggressive ways.
+ - A non-initialized `static const` is not allowed.
+ - Any dead-store to a `static int` can be removed.
+ - Meaning any pure computation parent to this store can be removed.
+
+Solving this can be done in 3 ways:
+ - Change the spec, allowing the syntax to be less "magical": use `extern`.
+ - Fixup the storage class of those variables in Clang.
+ - Emit an external constructor & destructor for those preventing optimization.
```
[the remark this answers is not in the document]
llvm-beanz: Using a "constructor" for custom builtins is pretty similar to the approach we're using for handling semantics. We haven't yet worked on support for writing back to semantics, but I don't think that will be a problem.

In Clang today we generate two versions of each entry function. One that takes parameters, and one that doesn't. The version that takes no parameters is the true entry, and it initializes local allocas to the system values (builtins in SPIR-V parlance), then passes those as parameters into the secondary function which is called with always_inline.

This could be extended to emit global builtin initialization into the entry function as well.

proposals/NNNN-spirv-input-builtin.md
```diff
@@ -0,0 +1,159 @@
...
+
+I'd be in favor of the first solution, which correctly expresses in HLSL what
+those built-in variables are, see this
+[spec issue](https://github.com/microsoft/hlsl-specs/issues/350).
+If we emit those as `extern [thread_local]`, everything works out of the box,
+and HLSL feels less magical.
```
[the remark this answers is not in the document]
Keenuts: That's where the `thread_local` comes in. Yes, a simple global `extern` is assumed to be shared across threads, but if we have an `extern thread_local`, then it would carry the semantic we want in HLSL, while being also find from LLVM's point of view.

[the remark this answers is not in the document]
s-perron: > We need to distinguish between how the variable works in llvm, and how it works in HLSL.

I'm thinking about this from the perspective of the person who is writing their code in HLSL. There is no `thread_local` in HLSL. The `static` keyword is used to indicate that a global variable is thread local. So I don't think the spec should be changed.

How it is represented in LLVM is an important issue, and I don't know what is best. I'll let @llvm-beanz  comment on what is more invasive in clang: static that is initialized or used in an intrinsic in the wrapper, or changing variable to extern thread local internally. This is in some ways orthogonal to changing the spec.

Keenuts:
Thanks, I have published another draft PR and updated this PR to avoid changing the spec.
This should be closer to what DXIL is doing, but will change a bit how the few semantics we already have are implemented (since we now have a generic intrinsic).

Keenuts:
@tex3d what were your concerns about this design?

proposals/NNNN-spirv-input-builtin.md
```diff
@@ -0,0 +1,157 @@
...
+```
+
+The emitted global variable for inline SPIR-V will be in the global scope,
+meaning it will require the `Private` storage class in SPIR-V.
+Because of that, globals linked to the added attributes will require a new
+address space `vulkan_private`.
+
+## Backend changes
+
+Backend will translate the new `vulkan_privatge` address space to
```
s-perron, line 135:
```suggestion
The backend will translate the new `vulkan_private` address space to
```

s-perron (commented):
This looks good to me. The only question is what we will do with the address space. I'll wait until that is finalized before I approve.

Keenuts:
> > @tex3d what were your concerns about this design?
> 
> Some of his question were about how well it could be optimized. The two main cases are:
> 
> 1. If the input variable is never used, will the call to the load intrinsic be removed?
> 2. If the output variable is never actually written it, could the store intrinsic be removed?

For 1 and 2, I'd assume if we optimize everything else, and just leave a load/store, driver should be able to optimize those away.

However, form the Vulkan spec, an `Output` BuiltIn starts with an undefined value.
So in this design, if no value is written by the user, a default value will be written to the output in the dtor.
I'd say that's OK, since we replace an undefined behavior with another behavior.

proposals/NNNN-spirv-input-builtin.md
```diff
@@ -0,0 +1,157 @@
...
+
+## Frontend changes
+
+Global variables marked with Inline SPIR-V will be marked using two new
+attributes:
+- `HLSLInputBuiltin`
+- `HLSLOutputBuiltin`
+
+In addition, a new address space will be added:
+- `vulkan_private`
```
llvm-beanz, line 52: I'd be okay calling this just `private`. Since DX is eventually going to adopt SPIR-V we might as well start getting the language ready for that and not force renaming things in the future.

llvm-beanz (approved):
This all seems reasonable to me and aligns with how we're handling inputs and outputs in DirectX today.

proposals/NNNN-spirv-input-builtin.md
```diff
@@ -0,0 +1,157 @@
...
+
+Because the 2 new intrinsics will be marked as `MemWrite` or `MemRead`, LLVM
+won't be able to optimize them away, and the initial/final load/store will
+be kept.
+
+Corresponding code could be:
+
+```
+[[vk::ext_builtin_input(/* GlobalInvocationId */ 28)]]
+static const uint3 myBuiltIn;
```
tex3d, line 104: How is the type checked for a given built-in?  Are the available built-ins predetermined?  Could they be symbolically represented instead of passing a number?  Or does the compiler just allow you to do anything you want here, and it's up to you to make it valid for the SPIR-V output?

proposals/NNNN-spirv-input-builtin.md
```diff
@@ -0,0 +1,157 @@
...
+void main(uint3 thread_id_from_param : SV_DispatchThreadID) {
+  [...]
+}
+
+constructor_myBuiltIn() {
+  myBuiltIn = spv_load_builtin(28);
+}
+
+destructor_myBuiltIn() {
+  spv_store_builtin(28, myBuiltIn);
```
tex3d, line 115: Why have destructor and store for an _input_ built-in?  Similarly, why have constructor and load for an _output_ built-in?

proposals/NNNN-spirv-input-builtin.md
```diff
@@ -0,0 +1,157 @@
...
+
+The destructor will call another intrinsic: `spv_store_builtin`. This builtin
+takes the SPIR-V built-in ID, as well as a value to store as parameter.
+
+And lastly, to support input parameters with semantics, the same intrinsic
+will be called in the generated entrypoint wrapper function.
+
+Because the 2 new intrinsics will be marked as `MemWrite` or `MemRead`, LLVM
+won't be able to optimize them away, and the initial/final load/store will
+be kept.
```
tex3d, line 98: When looking at what DXC does, I see it loads/stores only under control flow, when that's the way it's used in the entry.  I think this proposal is written to imply that it will not match that behavior.  I imagine that an unused output will have an unconditional write at the end of every entry that has such a global declaration in the file.

This proposal offers no mechanism for guaranteeing accesses under control flow remain under that control flow, and no additional accesses are added to the built-ins as a side effect of some implementation detail, for either load or store.

tex3d (commented):
I think input and output are different enough to warrant separate examples and descriptions of behavior.  I'm not sure why we would be storing to an input in a destructor or loading from an output in a constructor.

This seems mainly about defining a mechanism for declaring global static variables as built-in shader inputs/outputs, rather than the wrapped entry parameter mechanism. So as a note, the example would be different wrapping a main with an output parameter, and implications for IR (pointer arg instead of value arg) and optimization opportunities are different as well.  Perhaps this should only focus on the special global variable attribute, and not the standard HLSL entry parameter case.

I am a bit concerned how we specify that variables only read from or written to in control flow preserve these intended load/store locations.  Unless you are always expecting unconditional load of input before shader and unconditional store of output after shader, this scheme relies on an unspoken assumption that optimizations can translate these cases:

**From:** unconditional load of built-in before entry and store to static global var, then conditional load from static global under control flow
**To:** conditional load of built-in under control flow.

**From:** conditional store to static global var under control flow, then unconditionally load static global after entry and store to built-in.
**To:** conditional store to built-in under control flow.

Additionally, if a built-in is never read or written, the load or store should be eliminated.  Eliminating the store seems particularly problematic.  This is one reason I think inputs and outputs need to be explored separately.

This proposal doesn't seem to discuss what's required for this to work properly.

I don't know of a language construct that expresses these semantics clearly without special-handling for these globals. Without a description of any required special language and compiler implementation implications, it's hard to evaluate whether the proposal can solve these issues when using a constructor/destructor paradigm.

I'm not a fan of the language semantics being significantly changed for a static global variable decl by using an attribute, but I realize that's the existing construct implemented in DXC, so I guess we have to support it somehow.

proposals/NNNN-spirv-input-builtin.md
Keenuts: It is up to you to select the correct builtin type for the ID you pass:
 - BuiltIns can be added by extensions.
 - This means compiler doesn't know about this extension, hence would have no ability to validate it anyway.
 - spirv-val, if it knows the extension, can validate the type later on.
 
In DXC today, what happens is DXC always accept, then spirv-val complains if the builtin type is not the expected type.

What do you mean by symbolically represented?

proposals/NNNN-spirv-input-builtin.md
Keenuts: That is correct. Input could omit dtor, and output could omit ctor.

proposals/NNNN-spirv-input-builtin.md
Keenuts: Splitting 2 issues here:
 - unused input/output not being discarded/removed
 - control flow access
 
@s-perron to double check, but AFAIK in SPIR-V, The Built-in value is ONLY visible to the executing lane. Meaning having a load/store moved outside of a given control-flow wouldn't impact the behavior of the shader.
(Specifically talking about the control flow bit, non-discarded unused built ins is another issue)

Keenuts:
> I think input and output are different enough to warrant separate examples and descriptions of behavior. I'm not sure why we would be storing to an input in a destructor or loading from an output in a constructor.

Agree, I'll modify the proposal to remove dtor for input as those are not useful.
For ctor for output, see part below about storing undef.

> I am a bit concerned how we specify that variables only read from or written to in control flow preserve these intended load/store locations.

Ok I think I got it. The concern is: what if a lane conditionally stores to a built-in.
Since store to the builtin is done at the end, you worry the undefined value in the output will be changed in all paths, since the built-in store is now unconditional.

This specific issue can be solved by having a ctor on output builtins.
 - load undefined value in the global
 - shader modifies it or not in some control flow.
 - store back the value, which is either the original undefined, or a new value.

Keenuts:
@tex3d answered to your concerns, and changed the spec to share more details around those specifics handling of inputs/outputs.
Let me know if I missed something!

Keenuts:
@tex3d so Steven found a case which hinted as something incomplete: the PointSize BuiltIn.
Unlike this others, this one is assumed to have the value `1.0` if not stored to. But question was: would a load/store pair be OK.

The answer is actually not that clear.
In SPIR-V, those yield the same result:
 - Load, Store
 - Store 1.0
 - do nothing
All 3 options gives us a point size of 1.0.
And all are translated in a `reg = 1.f` in the final ISA I have access to (AMD RNDA).

But what's interesting is:
- `a = load(), store(a + 123)`
This specific case yields a `reg = NaN`. Which make me believe AMD has a special case for `store(load())`, but otherwise, a load yields an undefined value, hence why we have a `NaN`.

Additional weird detail:
```hlsl
printf("%f, %x", point_size, asuint(point_size)); // Prints "0, 0".
point_size = point_size + 10.f; // Yields point_size = NaN.
```

So this hints in the direction that the actual `store` could be assumed to have a hidden side-effect, and thus I shall not emit a `load`/`store` pair if the user hasn't written to the builtin.
So this means if a branch stores a value, and another doesn't, I cannot simple store in both.

Need to think how to solve this.

Keenuts:
Hello all!
So for the reasons outlined above (builtin store has an hidden side-effect), I changed the whole design to implement this as regular global variables, and remove the ctor/dtor/load/store builtins.

proposals/NNNN-spirv-input-builtin.md
```diff
@@ -0,0 +1,139 @@
...
+HLSL has semantic input/outputs parameters used to carry information
+to and from a shader, e.g: the system semantic `SV_GroupID` or `MY_SEMANTIC`,
+a user-defined semantic.
+
+In SPIR-V, those are translated to `Input` and `Output` variables,
+with either a `BuiltIn` or `Location` decoration.
+
+Input `BuiltIn` values are private to the executing lane. Reading their
+value has no side-effect. If not used, those built-in can safely be removed
+from the shader module (Unlike the inputs tagged with `Location`).
```
s-perron, line 18: In SPIR-V you can remove the user defined inputs (those with the Location decoration). Vulkan requires that the inputs are a subset of the output of the previous stage. They do not have to match exactly. You cannot remove outputs that are not written to. That will give a validation error. Here is an example where SPIR-V opt removes an unused input: https://godbolt.org/z/Mdo3WWssb.

proposals/NNNN-spirv-input-builtin.md
```diff
@@ -0,0 +1,139 @@
...
+}
+```
+
+This document explain how we plan on implementing those in Clang/LLVM.
+
+## Proposed Solution
+
+## Frontend changes
+
+Global variables marked with Inline SPIR-V will be marked using two new
```
s-perron, line 52: instead of saying "Inline SPIR-V" you should mention the specific attributes.

proposals/NNNN-spirv-input-builtin.md
```diff
@@ -0,0 +1,139 @@
...
+  let Spellings = [CXX11<"vk", "ext_builtin_output">];
+  let Args = [IntArgument<"BuiltIn">];
+  let Subjects = SubjectList<[HLSLOutputBuiltin], ErrorDiag>;
+  let LangOpts = [HLSL];
+}
+```
+
+When this attribute is encountered, several changes will occur:
+- AS will be set to `hlsl_input` for input built-ins.
+- AS will be set to `hlsl_output` for input built-ins.
```
s-perron, line 90: Spell out address space. It took me a few second to figure out what it meant.

s-perron, line 90:
```suggestion
- AS will be set to `hlsl_input` for input built-ins.
- AS will be set to `hlsl_output` for output built-ins.
```

proposals/NNNN-spirv-input-builtin.md
```diff
@@ -0,0 +1,139 @@
...
+- a `spirv.Decoration` metadata is added with the `BuiltIn <id>` decoration.
+
+The AS change will allow the back-end to correctly determine the variable
+storage class.
+The metadata will be converted `OpDecorate <reg> BuiltIn <id>`.
```
s-perron, line 95:
```suggestion
The metadata will be converted to `OpDecorate <reg> BuiltIn <id>`.
```

proposals/NNNN-spirv-input-builtin.md
```diff
@@ -0,0 +1,139 @@
...
+```
+
+In have the user's entrypoint returns a struct with semantic on fields,
+the entrypoint will have 1 store per semantic, and the module 1 global per
+semantic.
+
+## Backend changes
+
+Backend will translate the new `vulkan_input` address space to
+`StorageClass::Input`, and `vulkan_output` to `StorageClass::Output`.
```
s-perron, line 125: Do you mean `hlsl_input` instead of `vulkan_input`?

s-perron: You should also mention which backend.

```suggestion
The SPIR-V backend will translate the new `hlsl_input` address space to
`StorageClass::Input`, and `hlsl_output` to `StorageClass::Output`.
```

proposals/NNNN-spirv-input-builtin.md
```diff
@@ -0,0 +1,139 @@
...
+}
+
+void generated_entrypoint() {
+  output_semantic = main(dispatch_thread_id);
+}
+```
+
+In have the user's entrypoint returns a struct with semantic on fields,
+the entrypoint will have 1 store per semantic, and the module 1 global per
+semantic.
```
s-perron, line 120:
```suggestion
If the entrypoint returns a struct with semantic on fields,
the entrypoint will have 1 store per semantic, and the module 1 global per
semantic.
```

proposals/NNNN-spirv-input-builtin.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Backend changes
+
+Backend will translate the new `vulkan_input` address space to
+`StorageClass::Input`, and `vulkan_output` to `StorageClass::Output`.
+
+Decoration lowering is already available.
```
s-perron, line 127:
```suggestion
The SPIR-V backend already accepts the `spirv.Decoration` metadata.
```

s-perron (approved):
From a SPIR-V perspective this looks good to me.

proposals/NNNN-spirv-input-builtin.md
Keenuts: Right, only outputs are an issue. Removed this line, and also rephrased the one above to just say "we don't talk about Location"

Keenuts:
Thanks for the review, feedback applied!

Keenuts:
@llvm-beanz is there something blocking a reviews from the MS side?

llvm-beanz (approved):

PR #99, opened 2024-11-11
Add a proposal for consistently naming `llvm.dx` intrinsics.

bogner:


tex3d (approved):
I think it's a good start (if I do say so myself ;).  Should we merge this then file issues for open questions, and update it as we resolve these?

proposals/NNNN-consistent-naming-for-dx-intrinsics.md
```diff
@@ -0,0 +1,228 @@
...
+#### GS Stream operations
+```llvm
+; GS stream operations (or just ".stream."?)
+llvm.dx.gsstream.emit ; (EmitStream)
+llvm.dx.gsstream.cut ; (CutStream)
+llvm.dx.gsstream.emitthencut ; (EmitThenCutStream)
+```
+
+Open questions:
+- `gstream`, or `stream`?
```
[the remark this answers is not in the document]
bogner: Makes sense to me. Updated the heading to spell out geometry shader and removed the open question.

proposals/NNNN-consistent-naming-for-dx-intrinsics.md
```diff
@@ -0,0 +1,224 @@
...
+llvm.dx.resource.samplecmpbias ; (SampleCmpBias)
+llvm.dx.resource.samplecmpgrad ; (SampleCmpGrad)
+llvm.dx.resource.samplecmplevel ; (SampleCmpLevel)
+llvm.dx.resource.samplecmplevelzero ; (SampleCmpLevelZero)
+llvm.dx.resource.texturesamplepos ; (Texture2DMSGetSamplePosition)
+llvm.dx.resource.gather ; (TextureGather)
+llvm.dx.resource.gathercmp ; (TextureGatherCmp)
+llvm.dx.resource.gatherraw ; (TextureGatherRaw)
+llvm.dx.resource.atomicbinop ; (AtomicBinOp)
+llvm.dx.resource.atomiccmpxchg ; (AtomicCompareExchange)
```
inbelic, line 96: This and `pos` for position above seem to be the only places we use abbreviations of the dxil op name, maybe better to write it all out everywhere?

bogner: I think these two abbreviations are both defensible:
- In the "pos" case we're abbreviating from "Texture2DMSGetSamplePosition" to "texturesamplepos". I'm fine with calling this "texturesampleposition" rather than "pos", but it still wouldn't match and I think saying "texture" rather than "texture2dms" is better here.
- "cmpxchg" is an extremely common abbreviation for "compare and exchange" and easy to parse at a glance, whereas "compareexchange" is arguably harder to read.

I don't feel super strongly here, but I have a weak preference for keeping these as is.

inbelic: Any reason to keep it as-is is good enough for me

llvm-beanz:
cc @s-perron & @Keenuts, since we do rely on some of the DX and SPV intrinsics sharing name suffixes.

Keenuts (approved):
Fine with this!

llvm-beanz (approved):

PR #104, opened 2024-11-27
List of things to improve in the HLSL language

hekota:
This PR starts documenting a list of features that we might consider removing in a future version of the HLSL language.  It includes things that:
- have unusual syntax or odd behavior
- are remnants of outdated features that are no longer used or relevant
- are construct that are generally considered "not very nice" from a language point of view

### DIFF
--- a/proposals/0000-things-to-improve.md
+++ b/proposals/0000-things-to-improve.md
@@ -1,0 +1,69 @@
+<!-- {% raw %} -->
+
+# List of Things to Improve in the HLSL Language
+
+* Proposal: [NNNN](NNNN-things-to-improve.md)
+* Author(s): [Helena Kotas](https://github.com/hekota)
+* Status: **Design In Progress**
+
+## Introduction
+
+This document includes a list of features that we might consider removing in a future
+version of the HLSL language. It includes things that:
+- have unusual syntax or odd behavior
+- are remnants of outdated features that are no longer used or relevant
+- are construct that are generally considered "not very nice" from a language point of view
+
+## Motivation
+
+We strive to evolve HLSL into a modern shader language. In order to so we need to recognize
+odd, outdated or obscure features that are no longer used or that we do not wish to support
+going forward.
+
+## List of Things to Improve
+
+### 1. Global variables outside of `cbuffer` context
+
+All global variables declared outside of `cbuffer` context go implicitly into `$Globals` constant buffer. If the variable has an initializer it gets ignored.
+
+We should either require all global variables to be declared withing `cbuffer` context and stop supporting `$Globals`, or we should require any top-level non-static global variable to be declared with `const` and report error if it is initialized.
+
+### 2. Nested `cbuffer` declarations
+
+Nesting of `cbuffer` declarations and allowing functions inside `cbuffer` declaration context.
+
+```c++
+cbuffer CB {
+  int a;
+
+  cbuffer CB {  // duplicate name ok
+    int b;
+
+    int foo(int x) {
+        return x * 2*;
+    };
+  }
+};
+```
+
+This is equivalent to:
+
+```c++
+cbuffer CB {
+  int a;
+};
+
+cbuffer CB {  // duplicate name ok
+  int b;
+};
+
+int foo(int x) {
+    return x * 2*;
+};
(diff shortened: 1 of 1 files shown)


proposals/0000-things-to-improve.md
```diff
@@ -0,0 +1,69 @@
...
+
+We strive to evolve HLSL into a modern shader language. In order to so we need to recognize
+odd, outdated or obscure features that are no longer used or that we do not wish to support
+going forward.
+
+## List of Things to Improve
+
+### 1. Global variables outside of `cbuffer` context
+
+All global variables declared outside of `cbuffer` context go implicitly into `$Globals` constant buffer. If the variable has an initializer it gets ignored.
```
damyanp, line 27: line wrapping

damyanp, line 27: > If the variable has an initializer it gets ignored.

A little ambiguous whether "it" refers to the variable or the initializer.  Maybe "Any initializers on global variables are ignored".

proposals/0000-things-to-improve.md
```diff
@@ -0,0 +1,69 @@
+<!-- {% raw %} -->
+
+# List of Things to Improve in the HLSL Language
```
damyanp, line 3: I'm wondering if a proposal document is the right way to capture this?  Any reason not to use issues for each thing?  Then a proposal document can be written with a concrete proposal on addressing the thing-that-needs-to-be-improved.

proposals/0000-things-to-improve.md
```diff
@@ -0,0 +1,69 @@
...
+odd, outdated or obscure features that are no longer used or that we do not wish to support
+going forward.
+
+## List of Things to Improve
+
+### 1. Global variables outside of `cbuffer` context
+
+All global variables declared outside of `cbuffer` context go implicitly into `$Globals` constant buffer. If the variable has an initializer it gets ignored.
+
+We should either require all global variables to be declared withing `cbuffer` context and stop supporting `$Globals`, or we should require any top-level non-static global variable to be declared with `const` and report error if it is initialized.
```
damyanp, line 29: Why are static global variables different here?

I'm wondering if we should avoid suggesting solutions in this document since debate over solutions might detract from the main purpose of it.

damyanp, line 29: This suggested solution implies that the only issue with global variables outside of cbuffer is that initializers are ignored.  Is that the primary thing we're trying to solve here, or are there more reasons that $Globals is undesireable?

proposals/0000-things-to-improve.md
```diff
@@ -0,0 +1,69 @@
...
+  int b;
+};
+
+int foo(int x) {
+    return x * 2*;
+};
+
+```
+https://godbolt.org/z/PeW4E3aj3
+
```
damyanp, line 66: My 2c: we should get rid of cbuffer entirely and just use `ConstantBuffer` instead so we have more regular syntax for other types of buffers.

proposals/0000-things-to-improve.md
hekota: I've created this list based on a @llvm-beanz's suggestion that it would be helpful to have one place that lists all of the not-so-nice features and obscure constructs in the language. We can then refer to it when designing a new language version. Issues would probably too as long as they are labeled such that they're easy to find.

proposals/0000-things-to-improve.md
hekota: @tex3d - would you have more to add here why we should get rid of `$Globals` constant buffer?

hekota: I think it would be useful to list potential solutions without going to too much detail.

hekota: I've removed the `non-static`. I've update it to list possible solution without going into too much of a detail.

llvm-beanz:
@hekota, thank you for starting to put this together! 

Because this is language-related we should probably instead have this over in https://github.com/microsoft/hlsl-specs/.

I do wonder if this is something we should track as a proposal, or just as more general documentation though. Maybe we should start writing some docs around longer term language direction that we can collect in hlsl-specs?

PR #105, opened 2024-11-28
[NNNN] Target Extension Types for Inline SPIR-V and Decorated Types

cassiebeckley:
This adds target extension types to represent `SpirvType` from [HLSL 0011, Inline SPIR-V](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0011-inline-spirv.md).

### DIFF
--- a/proposals/nnnn-inline-spirv-and-decorated-types.md
+++ b/proposals/nnnn-inline-spirv-and-decorated-types.md
@@ -5,6 +5,7 @@
 * Proposal: [NNNN](NNNN-filename.md)
 * Author(s): [Cassandra Beckley](https://github.com/cassiebeckley)
 * Status: **Design In Progress**
+* PRs: [#125316]([https://github.com/llvm/llvm-project/pull/NNNN](https://github.com/llvm/llvm-project/pull/125316))
 
 ## Introduction
 
--- a/proposals/nnnn-inline-spirv-and-decorated-types.md
+++ b/proposals/nnnn-inline-spirv-and-decorated-types.md
@@ -32,10 +32,10 @@ To represent `vk::SpirvType` and `vk::SpirvOpaqueType` in LLVM IR, we will add
 three new target extension types:
 
 | Type                     | HasZeroInit | CanBeGlobal | CanBeLocal |
-|--------------------------|-------------|-------------|------------|
-| `spirv.Type`             | - [ ]       | - [x]       | - [x]      |
-| `spirv.IntegralConstant` | - [ ]       | - [ ]       | - [ ]      |
-| `spirv.Literal`          | - [ ]       | - [ ]       | - [ ]      |
+| ------------------------ | ----------- | ----------- | ---------- |
+| `spirv.Type`             | false       | true        | true       |
+| `spirv.IntegralConstant` | false       | false       | false      |
+| `spirv.Literal`          | false       | false       | false      |
 
 `IntegralConstant` and `Literal` are used to encode arguments to `Type`, and
 may not be used outside that context. They are necessary because target
--- a/proposals/nnnn-inline-spirv-and-decorated-types.md
+++ b/proposals/nnnn-inline-spirv-and-decorated-types.md
@@ -1,0 +1,130 @@
+<!-- {% raw %} -->
+
+# Target Extension Types for Inline SPIR-V and Decorated Types
+
+* Proposal: [NNNN](NNNN-filename.md)
+* Author(s): [Cassandra Beckley](https://github.com/cassiebeckley)
+* Status: **Design In Progress**
+
+## Introduction
+
+In this proposal, we define the `spirv.inline.Type` and `spirv.DecoratedType`
+target extension types for the SPIR-V backend.
+
+## Motivation
+
+We would like to implement `SpirvType` and `SpirvOpaqueType` from [Inline
+SPIR-V (HLSL proposal 0011)](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0011-inline-spirv.md#types)
+in the SPIR-V backend. These allow users to create arbitrary `OpType`
+instructions, and use them like any other HLSL type. Using these types, a
+vendor creating a new extension can expose it to users by creating a header
+file without needing to modify the compiler.
+
+Additionally, we need a way to represent types with SPIR-V decorations in LLVM
+IR.
+
+## Proposed solution
+
+### SpirvType
+
+To represent `vk::SpirvType` and `vk::SpirvOpaqueType` in LLVM IR, we will add
+three new target extension types:
+
+* `spirv.inline.Type`
+* `spirv.inline.IntegralConstant`
+* `spirv.inline.Literal`
+
+`IntegralConstant` and `Literal` are used to encode arguments to `Type`, and
+may not be used outside that context. They are necessary because target
+extension types must have all type arguments precede all integer arguments,
+whereas SPIR-V type instructions may have an arbitrary number of type,
+immediate literal, and constant id operands in any order.
+
+#### `spirv.inline.Type`
+
+```
+target("spirv.inline.Type", operands..., opcode)
+```
+
+`opcode` is an integer literal representing the opcode of the `OpType`
+instruction to be generated. `operands` represents a list of type arguments
+encoding the operands of the `OpType` instruction. Each operand must be one of:
+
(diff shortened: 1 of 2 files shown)


proposals/nnnn-inline-spirv-and-decorated-types.md
```diff
@@ -0,0 +1,130 @@
...
+IR.
+
+## Proposed solution
+
+### SpirvType
+
+To represent `vk::SpirvType` and `vk::SpirvOpaqueType` in LLVM IR, we will add
+three new target extension types:
+
+* `spirv.inline.Type`
```
s-perron, line 33: Have you looked into how the size and alignment information can be encoded? How do you get `TargetExtType::getLayoutType()` to return the right value?

https://llvm.org/doxygen/classllvm_1_1TargetExtType.html#ae473a097292848f10f82f0961245e012

Are we okay with it returning `void`? It seems like the information comes from `getTargetTypeInfo` in `Type.cpp`. Will you be able to set this properly? Do we need optional integer arguments that have the size and alignment?

If we do have the size and alignment, what is the layout type? 

Also, which properties should be set?

proposals/nnnn-inline-spirv-and-decorated-types.md
```diff
@@ -0,0 +1,130 @@
...
+Here's an example of using these types to represent an array of images:
+
+```
+%type_2d_image = type target("spirv.Image", float, 1, 2, 0, 0, 1, 0)
+%integral_constant_28 = type target("spirv.inline.IntegralConstant", i32, 28)
+%integral_constant_4 = type target("spirv.inline.IntegralConstant", i32, 4)
+%ArrayTex2D = type target("spirv.inline.Type", %type_2d_image, %integral_constant_4, 28)
+```
+
+### Type decorations
```
s-perron, line 87: You should mention that this is related to the decoration attribute in inline spir-v, which can only be applied to types and member variables. Talk about how that attribute will be translated into something of this type. I believe this has alternative solutions that we discussed. The might be worth mentioning in the alternative solutions seciton.

Lastly, you need to consider all versions of the decoration, decorateId and decorateString? Do we want multiple versions? Use the IngralConstant, Literal, and some other String class?

Sorry, I should have mentioned these variants when I asked you to include the decorations in this design.

proposals/nnnn-inline-spirv-and-decorated-types.md
```diff
@@ -0,0 +1,130 @@
...
+`spirv.DecoratedType` target extension type.
+
+```
+target("spirv.DecoratedType", type, decoration)
+```
+
+`type` is any valid LLVM IR type, and `decoration` is the integer value of the
+decoration, as defined in Section [3.20. Decoration](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#_decoration)
+of the SPIR-V Specification. This will create a unique SPIR-V type with the
+specified `OpDecoration`.
```
s-perron, line 99: I would call out more explicitly this does not add a decoration to the id for `type`. In must create a new type with a new id.

s-perron (commented):
Thanks. This looks good. We will start updating other proposals that will depend on this proposal.

proposals/nnnn-inline-spirv-and-decorated-types.md
s-perron: Also I had not noticed `int_spv_assign_decoration` in https://github.com/llvm/llvm-project/blob/main/llvm/docs/SPIRVUsage.rst#target-intrinsics. We should explain why we are not using it, if we decide not to use it. If it does work for us, then that could be a better solution.

proposals/nnnn-inline-spirv-and-decorated-types.md
```diff
@@ -0,0 +1,130 @@
...
+## Proposed solution
+
+### SpirvType
+
+To represent `vk::SpirvType` and `vk::SpirvOpaqueType` in LLVM IR, we will add
+three new target extension types:
+
+* `spirv.inline.Type`
+* `spirv.inline.IntegralConstant`
+* `spirv.inline.Literal`
```
s-perron, line 35: Is there a reason we need `inline` in the name? I think it could be removed in all of the types.

proposals/nnnn-inline-spirv-and-decorated-types.md
```diff
@@ -0,0 +1,130 @@
...
+%ArrayTex2D = type target("spirv.inline.Type", %type_2d_image, %integral_constant_4, 28)
+```
+
+### Type decorations
+
+In order to represent types with SPIR-V decorations added, we define a new
+`spirv.DecoratedType` target extension type.
+
+```
+target("spirv.DecoratedType", type, decoration)
```
s-perron, line 93: Some decorations have operands. This might need to be more general.

proposals/nnnn-inline-spirv-and-decorated-types.md
s-perron: https://github.com/llvm/llvm-project/blob/3173a4fc3aca3b8b21e9064cc38383be3539b3a4/llvm/lib/IR/Type.cpp#L965

cassiebeckley: As discussed, the spec has been updated to add `size` and `alignment` parameters to `spirv.Type`. The implementation will need to update `getTargetTypeInfo` to use these values. I've also specified the property values for the types.

proposals/nnnn-inline-spirv-and-decorated-types.md
cassiebeckley: No, it's just to distinguish them from the concrete SPIR-V types. I'll remove it from the spec.

s-perron (commented):
This look reasonable to me. When we have a draft PR, we can run it past other, and get it discussed at the SPIR-V backend meeting and/or the HLSL implementation meeting.

damyanp:
@llvm-beanz - gentle ping

s-perron (approved):

proposals/nnnn-inline-spirv-and-decorated-types.md
```diff
@@ -0,0 +1,132 @@
...
+#### `spirv.Type`
+
+```
+target("spirv.Type", operands..., opcode, size, alignment)
+```
+
+`opcode` is an integer literal representing the opcode of the `OpType`
+instruction to be generated. `size` and `alignment` are integer literals
+representing the number of bytes a single value of the type occupies and the
+power of two that the value will be aligned to in memory. An opaque type can be
```
Keenuts, line 55: nit: I don't think the spec requires the alignment to be a power of 2

proposals/nnnn-inline-spirv-and-decorated-types.md
```diff
@@ -0,0 +1,132 @@
...
+
+#### `spirv.IntegralConstant`
+
+```
+target("spirv.IntegralConstant", integral_type, value...)
+```
+
+`integral_type` is the type argument for the `OpConstant` instruction to be
+generated, and `value` is its literal integer value. To represent a value larger
+than a single 32-bit word, multiple 32-bit words may be passed as arguments.
```
Keenuts, line 73: nit: shall we mention the order of the passed chunks? LSB then MSB or MSB then LSB?

Keenuts: (same below)

Keenuts (approved):
LGTM, just 2 nits

proposals/nnnn-inline-spirv-and-decorated-types.md
cassiebeckley: Hm, this value is specified as being a power of 2 in the [HLSL `SpirvType` proposal](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0011-inline-spirv.md). It makes implementation on the backend simpler as well, as we can just use an appropriately-sized int as the layout type. If we have specific requests from users maybe we can relax this requirement, but otherwise I think it would be fine to keep it.

proposals/nnnn-inline-spirv-and-decorated-types.md
cassiebeckley: Sure, as stated in the SPIR-V specification, these should be little-endian. I've updated this proposal with this information.

s-perron:
Update the proposal number, and I can merge.

cassiebeckley:
@s-perron done!

PR #106, opened 2024-12-05
Changing Codegen section name in Root Signature proposal

joaosaffran:


proposals/0002-root-signature-in-clang.md
```diff
@@ -400,7 +399,7 @@ signatures with functions.
 Example for same root signature as above:
 
 ```llvm
-!dx.rootsignatures = !{!2} ; list of function/root signature pairs
+!hlsl.rootsignatures = !{!2} ; list of function/root signature pairs
```
damyanp, line 402: Why this change to the design?

proposals/0002-root-signature-in-clang.md
```diff
@@ -385,13 +384,13 @@ parsedRootSignature = RootSignature{
...
-During frontend code generation an IR-based representation of the root signature
-is generated from the in-memory data structures stored in the AST. This is
-stored as metadata nodes, identified by named metadata. The metadata format
-itself is a straightforward transcription of the in-memory data structure - so
-it is a list of root elements.
+LLVM IR code generation, during entry point emition,
+an IR-based representation of the root signature is generated from the in-memory 
+data structures stored in the AST. This is stored as metadata nodes, identified 
+by named metadata. The metadata format itself is a straightforward transcription 
+of the in-memory data structure - so it is a list of root elements.
```
damyanp, line 393: I don't understand why this has been changed.  Before reviewing it too closely (I'm having trouble understanding the new version) I think it'd help to understand what different information is intended to be communicated.

proposals/NNNN-root-signature-dx-containe-dx-container-codegen.md
```diff
@@ -0,0 +1,130 @@
...
+*During the review process, add the following fields as needed:*
+
+* PRs: [#NNNN](https://github.com/microsoft/DirectXShaderCompiler/pull/NNNN)
+* Issues:
+  [#NNNN](https://github.com/microsoft/DirectXShaderCompiler/issues/NNNN)
+  -->
+
+## Introduction
+
+[Purposal 0002](https://github.com/joaosaffran/wg-hlsl/blob/purposal/root-signatures/proposals/0002-root-signature-in-clang.md?plain=1)
```
damyanp, line 20:
```suggestion
[Proposal 0002](0002-root-signature-in-clang.md)
```

"Proposal" is spelled incorrectly, but more importantly this links to a version in your fork which isn't the right thing to do.

proposals/NNNN-root-signature-dx-containe-dx-container-codegen.md
```diff
@@ -0,0 +1,130 @@
...
+in a binary serialized format need to be added into clang.
+
+This change proposes adding:
+
+* Conversion of the metadata representation to the binary serialized format.
+* Testing solutions to validate root signature blob generation within DX Container.
+
+## Motivation
+
+The motivation is the same as the one in [Purposal 0002](https://github.com/joaosaffran/wg-hlsl/blob/purposal/root-signatures/proposals/0002-root-signature-in-clang.md#motivation).
```
damyanp, line 31:
```suggestion
The motivation is the same as the one in [Proposal 0002](0002-root-signature-in-clang.md#motivation).
```

proposals/NNNN-root-signature-dx-containe-dx-container-codegen.md
```diff
@@ -0,0 +1,130 @@
...
+
+## Proposed solution
+### Generating of Root Signatures BLOB
```
damyanp, line 34:
```suggestion
### Generating of Root Signature blob
```

damyanp (commented):
Some notes added.

I'm not sure we should approach this as "splitting" the proposal.

0002 can be the overview. I think we may need to add a section to the overview about how the data itself will end up in the DXIL container.

We could add an additional proposal for the testing strategy if we think we need to build some new tooling to help with that and you'd like to have a discussion about the design of that tool.

proposals/0002-root-signature-in-clang.md
joaosaffran: We have 2 layers of metadata being generated, when the metadata comes from hlsl and goes into llvm ir, we called it an hlsl metadata, so that is why I changed to hlsl. I am following the pattern done for MS Style If attributes, that also generate metadata

damyanp: This is different - this is DX-specific and is similar to things like `!dx.valver` and `!dx.entryPoints`. I don't think we should use a different naming convention for this.

damyanp (approved):

inbelic (approved):

PR #173, opened 2025-02-11
[0002] RootSignatures - Add default values to proposal

### LINKED ISSUE #172: [0002] RootSignatures - Add default values to proposal
At there moment there is no specification of what optional parameters will have their default value set to when using the front-end parser.

This issue is resolved when there is clear documentation of what the default value for each optional parameter is.

inbelic:
- adds default value information to the Root Signature specification

Resolves #172

### DIFF
--- a/proposals/0002-root-signature-in-clang.md
+++ b/proposals/0002-root-signature-in-clang.md
@@ -397,7 +397,7 @@ RootSignature[
 ]
 ```
 
-When parsed will produce a the equivalent of:
+When parsed will produce the equivalent of:
 
 ```c++
 parsedRootSignature = RootSignature{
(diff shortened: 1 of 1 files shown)


proposals/0002-root-signature-in-clang.md
```diff
@@ -414,6 +414,50 @@ parsedRootSignature = RootSignature{
...
+Many of the parameters of each root element are optional. If they are not
+specified they will take the following default values. These comply with
+previous documentation
+[here](https://learn.microsoft.com/en-us/windows/win32/direct3d12/specifying-root-signatures-in-hlsl).
+
+General Parameters:
+
+- `visibility = SHADER_VISIBLITY_ALL`
+
+Descriptor Specific:
```
damyanp, line 428: Looks like this is a mixture of root descriptor and descriptor range specific ones.  Would be good to get those split out and explicit.

damyanp (commented):
A good start - I think we should avoid combining root descriptor and descriptor range settings.

proposals/0002-root-signature-in-clang.md
```diff
@@ -414,6 +414,50 @@ parsedRootSignature = RootSignature{
 };
 ```
 
+### Default Values of Optional Parameters
+
+Many of the parameters of each root element are optional. If they are not
+specified they will take the following default values. These comply with
+previous documentation
+[here](https://learn.microsoft.com/en-us/windows/win32/direct3d12/specifying-root-signatures-in-hlsl).
```
joaosaffran, line 422: There are 3 places already linking to this URL within this doc. It might be interesting adding this to a references sections, so it is easier to reference. Link it is done [here](https://github.com/llvm/wg-hlsl/blob/c18e560142fa58d2b2636bff67c2208a240277c5/proposals/0002-root-signature-in-clang.md?plain=1#L736C1-L737C1.).

```suggestion
[here][add-to-hlsl-docs].
[add-to-hlsl-docs]: https://learn.microsoft.com/en-us/windows/win32/direct3d12/specifying-root-signatures-in-hlsl.
```

damyanp (approved):

joaosaffran (approved):

PR #187, opened 2025-02-26
Proposal for handling member function address spaces.

s-perron:
As we expose new address spaces in HLSL, we will have a problem with the
address space for the `this` pointer in member functions.

### DIFF
--- a/proposals/NNNN-this-address-space.md
+++ b/proposals/NNNN-this-address-space.md
@@ -1,0 +1,158 @@
+<!-- {% raw %} -->
+
+# Allowing multiple address spaces for the `this` pointer
+
+*   Proposal: [NNNN](http://NNNN-this-address-space.md)
+*   Author(s): [Nathan Gauer](https://github.com/keenuts),
+    [Steven Perron](https://github.com/s-perron)
+*   Status: **Design In Progress**
+
+*During the review process, add the following fields as needed:*
+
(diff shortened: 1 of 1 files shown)


proposals/NNNN-this-address-space.md
```diff
@@ -0,0 +1,158 @@
...
+the member function to be called.
+
+These address space casts cause issues for SPIR-V because SPIR-V’s Generic
+storage class is not allowed by the Vulkan environment for SPIR-V, and a pointer
+cannot be cast from one storage class to another. The address space casts will
+be removed through optimizations and a fix-up pass.
+
+All calls to member functions will be inlined, even if the function is marked as
+noinline. Once inlined, a pass that will propagate the address space before the
+cast to all uses of the address space cast can be run.
```
Keenuts, line 86: maybe we should mention linking here:
 - if we do real libraries, then we'd need to keep the original function if exported, and apply the same transformation post-linking

Keenuts (approved):
LGTM, just one bit

proposals/NNNN-this-address-space.md
s-perron: Since we don't have a good idea of how linking will work, I don't want to say what we will do in that case. I'm not sure how member functions will work. Are we allowed to mark member functions as export (https://godbolt.org/z/8WeEfrGGc)?

llvm-beanz (approved):

Issue #190, opened 2025-03-10
[0002] RootSignatures - Missing `offset` field in metadata specification of Descriptor Ranges


inbelic:
The metadata description of descriptor ranges [here](https://github.com/llvm/wg-hlsl/blob/main/proposals/0002-root-signature-in-clang.md#descriptor-ranges) is missing a field to denote the `offset` of the descriptors.

The `offset` field should not be discarded to allow for specifying ranges that overlap each other and allow for register aliasing.

This issue is resolved when the field is added to the metadata specification.

PR #191, opened 2025-03-10
[0002] RootSignatures - Add missing Descriptor Ranges `offset`

inbelic:
- it appears the `offset` was mistakenly dropped as discarding this information no longer allows the user to provide a manual offset from the start of the descriptors heap and allow for register aliasing

- additionally, we do a small fix-up to update the flag values to be their correct default as we adjust the parameter values

Resolves https://github.com/llvm/wg-hlsl/issues/190

### DIFF
--- a/proposals/0002-root-signature-in-clang.md
+++ b/proposals/0002-root-signature-in-clang.md
@@ -392,7 +392,7 @@ RootSignature[
  "StaticSampler(s1),"
  "DescriptorTable("
  "  SRV(t0, numDescriptors=unbounded),"
- "  UAV(u5, space=1, numDescriptors=10))"
+ "  UAV(u5, space=1, numDescriptors=10, offset=5))"
 ]
 ```
 
@@ -482,8 +482,8 @@ Example for same root signature as above:
 !5 = !{ !"RootCBV", i32 0, i32 1, i32 0, i32 0 } ; register 0, space 1, 0 = visiblity, 0 = flags
 !6 = !{ !"StaticSampler", i32 1, i32 0, ... } ; register 1, space 0, (additional params omitted)
 !7 = !{ !"DescriptorTable", i32 0, !8, !9 } ;  0 = visibility, range list !8, !9
-!8 = !{ !"SRV", i32 0, i32 0, i32 -1, i32 0 } ; register 0, space 0, unbounded, flags 0
-!9 = !{ !"UAV", i32 5, i32 1, i32 10, i32 0 } ; register 5, space 1, 10 descriptors, flags 0
+!8 = !{ !"SRV", i32 0, i32 0, i32 -1, i32 -1, i32 4 } ; register 0, space 0, unbounded descriptors, offset append, flags 4
+!9 = !{ !"UAV", i32 5, i32 1, i32 10, i32 5, i32 2 } ; register 5, space 1, 10 descriptors, offset 5, flags 2
 ```
 
 See [Metadata Schema](#metadata-schema) for details.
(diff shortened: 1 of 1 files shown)


inbelic:
This was noted when working on implementation here: https://github.com/llvm/llvm-project/pull/125131/files#r1985869019.

It may be the case that it was intentional for `offset` to be omitted. In either case, I have created the issue/pr to have some documented record of why this was the case.

V-FEXrt (approved):

damyanp:
Nice catch!

D3D12_DESCRIPTOR_RANGE_OFFSET_APPEND is a magic value as well that should be mentioned in the proposal (see https://learn.microsoft.com/en-us/windows/win32/api/d3d12/ns-d3d12-d3d12_descriptor_range)

proposals/0002-root-signature-in-clang.md
```diff
@@ -413,6 +413,15 @@ parsedRootSignature = RootSignature{
 };
 ```
 
+### Special Parameter Values
```
inbelic, line 416: The motivation to create a seperate section is that there are some other special values that should also be documented, eg `numDescriptors = unbounded`.

inbelic:
Whoops, accidently pushed my other branch. No changes were made.

proposals/0002-root-signature-in-clang.md
```diff
@@ -413,6 +413,15 @@ parsedRootSignature = RootSignature{
...
+
+There are a couple of parameters that have enum specific values with interpreted
+meanings:
+
+- `offset`: Can be assigned the value `DESCRIPTOR_RANGE_OFFSET_APPEND`, which
+denotes that the described descriptor range immediately follows the preceding
+range. This will be resolved to a fixed numerical offset during serialization.
```
damyanp, line 423: Do we need to pass 'append' all the way through?  If we fix this up in the frontend and don't have a way to represent 'append' in the metadata, then this makes the metadata requirements simpler.

inbelic: We are able to compute this when we generate the metadata (computed as part of register range overlap anyways). So we would then be able to omit this from the metadata spec. However, a user would still be able to specify `-1` in the metadata and then the backend will interpret this as an append not as the literal offset value.

I am personally happy with that

damyanp: I guess what I'm suggesting here is that we don't allow -1 in the metadata.  So it's up to the frontend to specify the right offsets in the metadata.

inbelic: Ah I see. I have created an extra testcase here: https://github.com/llvm-beanz/offload-test-suite/pull/48. It demonstrates that we do not need to compute the discrete offset before passing it to the D3D CreateRootSignature api.

From offline discussion, this implies that we should not compute this in the frontend of clang so that we will generate the same dxil container. And so it follows that the metadata representation should allow `-1` to mean append.

Updated the spec to properly reflect this behaviour.

proposals/0002-root-signature-in-clang.md
```diff
@@ -712,6 +723,7 @@ Operands:
 * i32: number of descriptors in the range
 * i32: base shader register
 * i32: register space
+* i32: offset
```
joaosaffran, line 726: I would add the information about `DESCRIPTOR_RANGE_OFFSET_APPEND`, in this section. Or maybe just link to DX12 docs as we have done with other fields, https://learn.microsoft.com/en-us/windows/win32/api/d3d12/ns-d3d12-d3d12_descriptor_range

joaosaffran (approved):
LGTM

PR #193, opened 2025-03-17
Implicit Resource Binding - document current DXC behavior

hekota:
Initial commit of the implicit resource binding proposal.
Includes brief overview of resource bindings and documents current DXC behavior.

### DIFF
--- a/proposals/0000-implicit-resource-binding.md
+++ b/proposals/0000-implicit-resource-binding.md
@@ -1,0 +1,189 @@
+* Proposal: [NNNN](NNNN-implicit-resource-binding.md)
+* Author(s): [Helena Kotas](https://github.com/hekota)
+* Status: **Design In Progress**
+
+## Introduction
+
+HLSL Resources are runtime-bound data that are provided as input, output, or
+both to shader programs. Resources need to be bound to the graphics or compute
+pipeline for shaders to access them. In DirectX this is done via descriptors,
+which are sometimes referred to as views or handles.
+
+Binding of resources to descriptors in HLSL is done via `register` annotation on
+the resource variable declaration, such as:
+
+```c++
+RWBuffer<float4> Data : register(u2); 
+```
+
+DirectX has four types of descriptors/handles:
+- `UAV` - unordered access view (read-write) - using `u` registers
+- `SRV` - shader resource view (read-only) - using `t` registers
+- `CBV` - constant buffer view (read-only) - using `b` registers
+- `Sampler` - sampler (read-only) - using `s` registers
+
+The `register` annotation specifies the first descriptor the resource is bound
+ to. For simple resources that is the only descriptor the resource needs. In
+ case of arrays the `register` specifies the start of the description range
+ needed for the resource array. The dimensions of the array determine the size
+ of the description range.
+
+The annotation can also include virtual register space:
+
+```c++
+Texture2D<float4> Tx : register(t3, space1); 
+```
+
+_Note: See the HLSL Language specification section Resource Binding for more
+details._
+
+Specifying resource binding on a resource is not mandatory. If a binding is not
+provided by the user, it is up to the compiler to assign a descriptor to such
+resource. This is called _implicit resource binding_ and it is the focus of this
+document.
+
+## Current behavior in DXC
+
+This section documents the implicit resource binding assignment in DXC. For
+simplicty the examples below use the just the `u` registers and
+`RWBuffer<float>` resource, but same binding assignment rules apply to other
+register and resource types.
+
+### Simple Resource Declarations
+Unbound resources are processed in the order they appear in the code. They are
+assigned the first available register slot (or range) in `space0` where the
+resource fits. Unused resources are optimized away and do not participate in the
+implicit binding assignment.
+
+#### Example 1
+```c++
+RWBuffer<float> A : register(u0);
+RWBuffer<float> B;                 // gets u1
+RWBuffer<float> C: register(u2);   // unused
+RWBuffer<float> D;                 // gets u2
+
+[numthreads(4,1,1)]
+void main() {
+  A[0] = B[0] + D[0]; // all resources except C are used
+}
+```
+
+If `C` is used then `D` gets assigned register `u3`. 
+
+### Resource Arrays
+
+Constant-sized resource arrays are assigned a range of register slots regardless
+of whether the individual resources in the array are used or not. If none of the
+resources are used though the whole resource declaration is optimized away and its
+register range is available for use by other resources.
+
+#### Example 2
+```c++
+
+RWBuffer<float> A : register(u2);
+RWBuffer<float> B[4];                 // gets u3 because it does not fit before A (range 4)
+RWBuffer<float> C[2];                 // gets u0 because it fits before A (range 2)
+RWBuffer<float> D[50] : register(u6); // unused - no register assignment
+RWBuffer<float> E[2];                 // gets u7 which is right after B (range 4)
+
+[numthreads(4,1,1)]
+void main() {
+  A[0] = B[2][0] + C[1][0] + E[1][0]; // all resources except D are used
+}
+```
+
+### Unboud Resource Arrays
+
+Unbound resources are placed in `space0` after the highest register slot
+assigned and they take up the rest of the register slots.
+
+If there are resources declared after the unbound array that do not fit into the
+remaining space available in `space0` DXC reports an error.
+
+#### Example 3
+
+```c++
+// assume all resources are used
+RWBuffer<float> A : register(u2);
+RWBuffer<float> B[];               // gets u6 (unbounded range)
+RWBuffer<float> C;                 // gets u0 because it fits before A
+RWBuffer<float> D : register(u5);
+```
+
+#### Example 4
+DXC reports `error: resource E could not be allocated`:
+```c++
+// assume all resources are used
+RWBuffer<float> A : register(u2);
+RWBuffer<float> B[];               // gets u6 (unbounded range)
+RWBuffer<float> C;                 // gets u0 because it fits before A
+RWBuffer<float> D : register(u5);
+RWBuffer<float> E[10];             // error - does not fit before A
+```
+
+It looks like DXC never attempts to assing registers slots from virtual register
+space other than `space0`.
+
+#### Example 5
+Moving declaration of `E` before `B` compiles successfully and `B` gets assigned
+register slot `u13`:
+```c++
+// assume all resources are used
+RWBuffer<float> A : register(u2);
+RWBuffer<float> D[10];             // gets u3 (range 10)
+RWBuffer<float> B[];               // gets u13 (unbounded range)
+
+[numthreads(4,1,1)]
+void main() {
+  A[0] = B[2][0] + D[1][0];
+}
+```
+
+#### Example 6
+Since DXC seems to only assigns register slots from `space0` it reports error
+when there are:
+- two or more unbound resource arrays without explicit binding
+- a resource array with explicit binding in `space0` and one unbound resource
+  array
+
+```c++
+// assume all resources are used
+RWBuffer<float> C[];  // gets u0
+RWBuffer<float> D[];  // error
+```
+
+### Register spaces other than `space0`
+
+Use of other virtual register spaces does not seem to affect the implicit
+assignments in `space0`:
+
+```c++
+// assume all resources are used
+RWBuffer<float> A : register(u0, space1); 
+RWBuffer<float> B[10];   // gets u0 in space0 (range 10)
+RWBuffer<float> C[5][4] : register(u10, space5);
+RWBuffer<float> D;       // gets u10 in space0
+```
+
+### Miscelanious Notes
+
+The maximum register slot or virtual space number is `UINT_MAX` (max. value of
+32-unsigned integer).
+
+## Motivation
+
+We need to support implicit resource bindings in Clang as it is a fundamental
+part of the HLSL language. 
+
+## Proposed solution
+
+TBD
+
+## Detailed design
+
+## Alternatives considered (Optional)
+
+## Acknowledgments (Optional)
+
+
+<!-- {% endraw %} -->


proposals/0000-implicit-resource-binding.md
```diff
@@ -0,0 +1,348 @@
...
+```
+https://godbolt.org/z/Mcabxq4jG
+
+Based on resources identified as unused in the example above it is very likely
+that DXC is assigning implicit bindings later in the compiler after codegen and
+code optimizations.  
+
+## Motivation
+
+We need to support implicit resource bindings in Clang as it is a fundamental
```
bfavela, line 334: I will probably lose this argument, but I would challenge this statement. Implicit bindings are not allowed under certain scenarios, and make various tasks like GBV (gpu based validation) more difficult to debug as the implementation must be kept in the loop. Rather than being able to validate the root signature offline.

Yes, we can have a container that doesn't have a root signature, but, in the absence of one, the app needs to create one. If the app creates one, and they mismatch with one that's in a container, the PSO creation fails.

In my opinion, this is a shortcoming of DXC, not behavior we want to bring forward.

llvm-beanz: I'm not sure I follow the logic here. We can have implicit bindings and a root signature. We can also have no implicit bindings and have a runtime-provided root signature. For example this is completely valid:

```hlsl
cbuffer RootConstants {
  float4 C;
};

struct Input {
  float4 A;
  float4 B;
};

struct Output {
  float4 A;
};

StructuredBuffer<Input> In;
RWStructuredBuffer<Output> Out;

[RootSignature("CBV(b0), SRV(t0), UAV(u0)")]
[numthreads(1,1,1)]
void main(uint GI : SV_GroupIndex) {
  Out[GI].A = In[GI].A * In[GI].B * C;
}
```

Are you arguing that this should not be valid?

damyanp: Before I started working on the compiler I'd never considered embedding a root signature in a shader and always constructed them in my C++ code.  As far as I was concerned, my C++ code was specifying the interface - the root signature - that the shaders I used had to comply with.  Although there are reasons that this workflow isn't the most efficient, I don't think we should prohibit it.

llvm-beanz: I think we're also talking about different things.

Whether or not a root signature is provided in the shader, and whether or not resources are assigned register values implicitly are completely different.

At the DXIL level you can't even tell if a shader's bindings are implicit or explicit. Consider for example this shader:
```
#if IMPLICIT
#define BINDING(val)
#else 
#define BINDING(val) : register(val)
#endif

RWBuffer<int4> Buf BINDING(u0);

[numthreads(1,1,1)]
void main() {
    Buf[0] += Buf[0];
}
```
[Compiler Explorer](https://godbolt.org/z/WKe5bes4o)

Whether you compile this with `-DIMPLICIT` or not, the output is identical.

bfavela: I think I have to do some more reading as I'm confusing myself. Sorry, please disregard.

proposals/0000-implicit-resource-binding.md
```diff
@@ -0,0 +1,368 @@
...
+```
+
+DirectX has four types of descriptors/handles:
+- `UAV` - unordered access view (read-write) - using `u` registers
+- `SRV` - shader resource view (read-only) - using `t` registers
+- `CBV` - constant buffer view (read-only) - using `b` registers
+- `Sampler` - sampler (read-only) - using `s` registers
+
+The `register` annotation specifies the first descriptor the resource is bound
+ to. For simple resources that is the only descriptor the resource needs. In
```
inbelic, line 26:
```suggestion
to. For simple resources that is the only descriptor the resource needs. In
```

proposals/0000-implicit-resource-binding.md
```diff
@@ -0,0 +1,368 @@
...
+
+## Current behavior in DXC
+
+This section documents the implicit resource binding assignment in DXC. For
+simplicty the examples below use the just the `u` registers and
+`RWBuffer<float>` resource, but same binding assignment rules apply to other
+register and resource types as well.
+
+### Simple Resource Declarations
+ Resource without binding assignment that are declared directly at the global
```
inbelic, line 53:
```suggestion
### Simple Resource Declarations

Resource without binding assignment that are declared directly at the global
```

proposals/0000-implicit-resource-binding.md
```diff
@@ -0,0 +1,368 @@
...
+} s;
+
+[numthreads(4,1,1)]
+void main() {
+  A[0] = s.D[0] + s.B[0];
+}
+```
+https://godbolt.org/z/6a1b7Kvz3
+
+The maximum value of the register slot and virtual space number is `UINT_MAX`
```
inbelic, line 95: Just wanted to link here: https://github.com/llvm/wg-hlsl/blob/main/proposals/0002-root-signature-in-clang.md#all-the-values-should-be-legal. The register spaces do have some reserved values in the root signature. Maybe that is worth noting here?

proposals/0000-implicit-resource-binding.md
```diff
@@ -0,0 +1,368 @@
...
+
+### Constant-size Resource Arrays
+
+Constant-size resource arrays declared directly at the global scope are also
+processed in the order they appear in the source code. They are assigned the
+first range of available decriptors in `space0` that fits the array size
+regardless of whether the individual resources in the array are used or not, as
+long as at least one of the them is accessed.
+
+If none of the resources in the are used, the whole resource declaration is
```
inbelic, line 106:
```suggestion
If a resource isn't used, the whole resource declaration is...
```

proposals/0000-implicit-resource-binding.md
```diff
@@ -0,0 +1,368 @@
...
+assigned descriptor, or after the highest implictly assigned descriptor so far,
+whichever is greater. They take up the rest of the descriptor slots in `space0`.
+
+#### Example 3.1
+
+```c++
+RWBuffer<float> A : register(u1);
+RWBuffer<float> B[];     // gets u6 (unbounded range)
+RWBuffer<float> C : register(u5);
+RWBuffer<float> D[3];    // gets u2 because it first between A and C but not before A
```
inbelic, line 187:
```suggestion
RWBuffer<float> D[3];    // gets u2 because it fits between A and C but not before A
```

proposals/0000-implicit-resource-binding.md
```diff
@@ -0,0 +1,368 @@
...
+RWBuffer<float> D[4];    // error - does not fit in the remaining descriptor ranges in space0 
+
+[numthreads(4,1,1)]
+void main() {
+  A[0] = D[2][0] + C[0] + B[10][0];
+}
+```
+https://godbolt.org/z/6bYoG599P
+
+It looks like DXC never attempts to assing descriptors from virtual register
```
inbelic, line 215:
```suggestion
It looks like DXC never attempts to assigning descriptors from virtual register
```

proposals/0000-implicit-resource-binding.md
```diff
@@ -0,0 +1,368 @@
...
+RWBuffer<float> C : register(u5);
+
+[numthreads(4,1,1)]
+void main() {
+  A[0] = D[2][0] + C[0] + B[10][0];
+}
+```
+https://godbolt.org/z/cjP35n63d
+
+Since DXC seems to only assigns descriptors from `space0`, it reports error
```
inbelic, line 236:
```suggestion
Since DXC seems to only assign descriptors from `space0`, it reports an error
```

proposals/0000-implicit-resource-binding.md
```diff
@@ -0,0 +1,368 @@
...
+whenever there are:
+- two or more unbound resource arrays without explicit binding
+- one unboud resource array with explicit binding in `space0` and one or more
```
inbelic, line 239:
```suggestion
- one unbound resource array with explicit binding in `space0` and one or more
```

inbelic (commented):
Very useful to read through, thanks!

proposals/0000-implicit-resource-binding.md
hekota: This is probably validated just on the root-signature side. The compiler allows it: https://godbolt.org/z/4bfWz3E6r

proposals/0000-implicit-resource-binding.md
```diff
@@ -0,0 +1,369 @@
...
+[numthreads(4,1,1)]
+void main() {
+  A[0] = E[2][0] + C[1][0] + B[1][0];
+}
+```
+https://godbolt.org/z/qha84sjvT
+
+However, if the resource array is defined in a `struct`, the binding seems to be
+assigned to the individual array elements in the order they are used in the
+code. In other words, the array elements are treated as unrelated individual
```
bob80905, line 127: If the substitution for the word `code` that I made in my comment above is true, you might want to replace `code` too.
And if you can determine whether this behavior is solid or not, you may want to remove `seems to be`.

proposals/0000-implicit-resource-binding.md
```diff
@@ -0,0 +1,369 @@
...
+```
+https://godbolt.org/z/45j8aqTaf
+
+This seems wrong. Resource arrays should be bound to a continuous description
+range and the range should be reflected in the `createHandleFromBinding`
+arguments. It also makes the binding susceptible to change whenever the code
+changes and is not something that can be relied upon. 
+
+Array resources inside structs are also not allowed to use dynamic indexing. In
+the following example DXC reports error when the resource array `B` is accessed
```
bob80905, line 155: nit: Did you mean `DXC reports errors when the resource array B is accessed while the dynamic indexing of A is fine` ?

proposals/0000-implicit-resource-binding.md
```diff
@@ -0,0 +1,369 @@
...
+[numthreads(4,1,1)]
+void main() {
+  A[0] = s.E[2][0] + s.C[1][0] + s.B[3][0] + s.E[1][0];
+}
+```
+https://godbolt.org/z/45j8aqTaf
+
+This seems wrong. Resource arrays should be bound to a continuous description
+range and the range should be reflected in the `createHandleFromBinding`
+arguments. It also makes the binding susceptible to change whenever the code
```
bob80905, line 151: If code changes, use-order may change, which may cause changes in register assignment. I don't think this is a good secondary argument for changing the register assignment behavior, because either way register assignments change when the code changes, right? Maybe I'm not understanding.

proposals/0000-implicit-resource-binding.md
```diff
@@ -0,0 +1,369 @@
...
+void main() {
+  A[0] = D[2][0] + C[0] + B[10][0];
+}
+```
+https://godbolt.org/z/cjP35n63d
+
+Since DXC seems to only assign descriptors from `space0`, it reports an error
+whenever there are:
+- two or more unbound resource arrays without explicit binding
+- one unbound resource array with explicit binding in `space0` and one or more
```
bob80905, line 240: Note: it also emits an error with 2 explicitly bound unbounded resource arrays.
So perhaps this could be simplified to:
`two or more unbounded resource arrays in the same register space.`

proposals/0000-implicit-resource-binding.md
```diff
@@ -0,0 +1,369 @@
...
+[numthreads(4,1,1)]
+void main() {
+  A[0] = D[0] + C[3][3][0] + B[5][0];
+}
+```
+https://godbolt.org/z/EeW17MncE
+
+### Space-only register annotations
+
+DXC support binding annotations that only specify the register space from which the descriptor binding should be assigned.
```
bob80905, line 281: nit: supports

V-FEXrt (approved):

proposals/0000-implicit-resource-binding.md
```diff
@@ -0,0 +1,369 @@
...
+  for (int i = 0; i < 5; i++) {
+    if (i > 6)
+      s.D[0] = 1.0; // unreachable
+    if (bar(i))
+      foo();        // unreachable
+    if (bar(gi.x))
+      A[0] = E[2][i];
+  }
+}
+```
```
bob80905, line 346: Could you add a newline after? Strange how the text is blue here.

bob80905 (commented):
mostly nits, with a couple questions

proposals/0000-implicit-resource-binding.md
```diff
@@ -0,0 +1,369 @@
...
+void main() {
+  A[0] = D[0] + B[0];
+}
+```
+https://godbolt.org/z/sMs3c8j7T
+
+If `C` is used then `D` gets assigned descriptor `u3`. 
+
+However, if the resources are declared inside a `struct`, their binding seems be
+assigned in the order they are referenced in the code, and not the order in
```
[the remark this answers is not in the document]
bob80905: Maybe it's obvious that resources are never assigned to registers in a library shader and so we don't need to even mention that.

proposals/0000-implicit-resource-binding.md
hekota: I am not sure what you mean. The blue link to compiler explorer?

proposals/0000-implicit-resource-binding.md
hekota: There are two issues here:
1. the register assignment changes depending on use-order
2. the array is not bound to a continuous range of registers
Both are wrong, and we should fix that. Tex agreed that this is a bug in DXC.

proposals/0000-implicit-resource-binding.md
bob80905: All the text under the tick marks is blue, including the link, and `Based on resources ident...`

proposals/0024-implicit-resource-binding.md
```diff
@@ -1,4 +1,4 @@
-* Proposal: [NNNN](NNNN-implicit-resource-binding.md)
+* Proposal: [0024](NNNN-implicit-resource-binding.md)
```
bob80905, line 1: Did you also want to update the file name too?

hekota: Yes! Thank you! :)

proposals/0000-implicit-resource-binding.md
```diff
@@ -0,0 +1,369 @@
...
+RWBuffer<float> C: register(u2);   // unused
+
+struct S {
+  RWBuffer<float> B;                 // gets u2
+  RWBuffer<float> D;                 // gets u1
+} s;
+
+[numthreads(4,1,1)]
+void main() {
+  A[0] = s.D[0] + s.B[0];
```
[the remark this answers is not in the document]
hekota: The resource accesses here are not function parameters, so that rule does not apply. I would guess that the evaluation order of binary expressions is guaranteed, but even if not, this way of assigning bindings is certainly not something that can be relied upon.

proposals/0000-implicit-resource-binding.md
```diff
@@ -0,0 +1,369 @@
...
+S s;
+
+[numthreads(4,1,1)]
+void main() {
+  A[0] = s.E[2][0] + s.C[1][0] + s.B[3][0] + s.E[1][0];
+}
+```
+https://godbolt.org/z/45j8aqTaf
+
+This seems wrong. Resource arrays should be bound to a continuous description
```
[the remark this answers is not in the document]
hekota: I believe `continuous` is a better fit here as in "uninterrupted in sequence" :)

bob80905 (approved):

proposals/0000-implicit-resource-binding.md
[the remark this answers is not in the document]
bob80905: Hm, would've thought addition was a function technically, but regardless your point that it's use dependent still stands.

Issue #192, opened 2025-03-13
[0002] RootSignatures - Fixes to Root Signature Specifiction


inbelic:
This issue tracks a collection of small issues to the [root signature specification](https://github.com/llvm/wg-hlsl/blob/main/proposals/0002-root-signature-in-clang.md) that were uncovered as part of [testing work](https://github.com/llvm/llvm-project/issues/130826).

This will be marked as designing until the below list of notes below is comprehensive.

AC:

- [ ] Denote that keywords are case-insensitive, and registers are not (`B3` is not valid)
- [ ] `numDescriptors` also allows the value of `unbounded` inclusive of an unsigned number
- [ ] Manual `offset` must be able to be user-specified and has `DESCRIPTOR_RANGE_OFFSET_APPEND` enum value
- [ ] Revisit the use of `NUMBER` in the grammar. Many values only accept an unsigned integer and DXC will error if given otherwise
- [ ] Document that all parameters (including mandatory) can be specified in any order and can only be specified exactly once
- [ ] Double-check the `AllowLowTierReservedHwCbLimit` root flag, it is not recognized in `dxc`
- [ ] Add validation of the maximum values of `offset`, `register`, `space`, `numDescriptors`, `num32BitConstants`

PR #195, opened 2025-03-20
[0002] RootSignatures - Misc updates to spec from testing

inbelic:
This issue addresses the discrepancies found from testing [here](https://github.com/llvm/llvm-project/issues/130826).

The pr has been broken up to commits that address the bullet points in the original issue. Briefly:

- First 2 commits work to have a concrete EBNF description of the grammar that denotes that all parameter args can be specified in any order
- It is also addressed that `NUMBER` can be more restrictive with respect to what the dxc/clang compiler will accept
- Missing documentation for `unbounded` is added
- Add additional validation of some parameters that surface as errors in dxc
- Remove the specification of 'AllowLowTierReservedHwCbLimit'
- A general reformat to be compliant with [EBNF notation](https://en.wikipedia.org/wiki/Extended_Backus%E2%80%93Naur_form)

Resolves https://github.com/llvm/wg-hlsl/issues/192


proposals/0002-root-signature-in-clang.md
```diff
@@ -154,161 +154,174 @@ to ensure that our solution doesn't unnecessarily tie the non-HLSL parts to it.
...
-               'DENY_DOMAIN_SHADER_ROOT_ACCESS' |
-               'DENY_GEOMETRY_SHADER_ROOT_ACCESS' |
-               'DENY_PIXEL_SHADER_ROOT_ACCESS' |
-               'DENY_AMPLIFICATION_SHADER_ROOT_ACCESS' |
-               'DENY_MESH_SHADER_ROOT_ACCESS' |
-               'ALLOW_STREAM_OUTPUT' |
-               'LOCAL_ROOT_SIGNATURE' |
-               'CBV_SRV_UAV_HEAP_DIRECTLY_INDEXED' |
-               'SAMPLER_HEAP_DIRECTLY_INDEXED' |
-               'AllowLowTierReservedHwCbLimit'
```
inbelic, line 178: We remove `AllowLowTierReservedHwCbLimit` from the specification in this pr. This was done is you can't specify the flag in dxc. I am not 100% sure this is correct and would like reviewer input.

damyanp: I think it's ok to leave this out.

proposals/0002-root-signature-in-clang.md
```diff
@@ -603,17 +603,31 @@ The additional semantic rules not already covered by the grammar are listed here
...
 
   `CBV(b0, space=4294967295)` is invalid due to the use of reserved space 0xFFFFFFFF.
 
-
 - Resource ranges must not overlap.
 
   `CBV(b2), DescriptorTable(CBV(b0, numDescriptors=5))` will result in an error
   due to overlapping at b2.
 
+  Note that a valid value for `numDescriptors` is `unbounded` add requires
```
damyanp, line 627: Typo - I think "add" should be "and"?

proposals/0002-root-signature-in-clang.md
```diff
@@ -603,17 +603,31 @@ The additional semantic rules not already covered by the grammar are listed here
   - MaxAnisotropy cannot exceed 16.
   - MipLODBias must be within range of [-16, 15.99].
 
+- RootConstants: `num32BitConstants`
+  A value of `num32BitConstants` that will cause the root signature size to
+  exceed 64 DWORDs of size can be validated. We can ensure that
+  `num32BitConstants <= 64 - (# of descriptors)`.
```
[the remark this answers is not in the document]
inbelic: Updated to be more descriptive and provide the source. [For reference](https://learn.microsoft.com/en-us/windows/win32/direct3d12/root-signature-limits#memory-limits-and-costs).

inbelic:
Okay, it should be properly formatted now, for review. Created a very strange git diff initially. I will open the formatting pr once this one is resolved.

proposals/0002-root-signature-in-clang.md
```diff
@@ -581,17 +601,32 @@ The additional semantic rules not already covered by the grammar are listed here
...
   - MipLODBias must be within range of [-16, 15.99].
 
+- RootConstants: `num32BitConstants`
+  A value of `num32BitConstants` that will cause the root signature size to
+  exceed 64 DWORDs of size can be validated. We can ensure that
+  `num32BitConstants <= 64 - (# of descriptor tables) - 2 * (# of root descriptors)`.
+  As described by the memory limits [here](https://learn.microsoft.com/en-us/windows/win32/direct3d12/root-signature-limits#memory-limits-and-costs).
+
+  `RootConstants(b0, num32BitConstants = 65)` will cause an error as it would
+  generate a root signature that is larger than 64 DWORDs.
```
joaosaffran, line 611: This seems to be a validation that encompass more than Root Constants, maybe we could create a new validation after this instead: https://github.com/llvm/wg-hlsl/blob/644eb3c643cd12987de0ff5ece823f84c32d0576/proposals/0002-root-signature-in-clang.md#texturestypedbuffers-cannot-be-bound-to-root-descriptors?

proposals/0002-root-signature-in-clang.md
```diff
@@ -174,35 +179,38 @@ to ensure that our solution doesn't unnecessarily tie the non-HLSL parts to it.
                'ALLOW_STREAM_OUTPUT' |
                'LOCAL_ROOT_SIGNATURE' |
                'CBV_SRV_UAV_HEAP_DIRECTLY_INDEXED' |
-               'SAMPLER_HEAP_DIRECTLY_INDEXED' |
-               'AllowLowTierReservedHwCbLimit'
+               'SAMPLER_HEAP_DIRECTLY_INDEXED'
+
+    RootConstants = 'RootConstants' '('
+      ( 'num32BitConstants' '=' POS_INT ) : bReg
```
joaosaffran, line 185: nit: Are we sure that those can happen in any order? Checking the DX12 docs, makes it seems that they order here is important: https://learn.microsoft.com/en-us/windows/win32/direct3d12/specifying-root-signatures-in-hlsl#root-constants

joaosaffran (commented):
This PR has the same changes as https://github.com/llvm/wg-hlsl/pull/191. Should it be targeting that branch instead?

proposals/0002-root-signature-in-clang.md
inbelic: Yep, see here: https://github.com/llvm-beanz/offload-test-suite/pull/45/files#diff-a947af2708f83e5c8a83e667a90f7db8abadca8a5197eeea169300e520c8b4a0.

I was also confused and thought the order mattered. The current parser pr is actually implemented such that the order matters when it shouldn't.

proposals/0002-root-signature-in-clang.md
inbelic: I investigated this a bit offline. The error that I encountered when writing the test is generated by `D3D12Core!CCreateRootSignatureValidator::Validate`. Notably, the allowed size is not actually always 64 DWORDs and can depend on the target platform. This validate function allows for providing much better context in the error message. So we will not do this validation here.

I will update accordingly.

inbelic:
Good catch, yes it had the commits for offset as well. Have rebased to remove those commits

proposals/0002-root-signature-in-clang.md
```diff
@@ -154,13 +154,18 @@ to ensure that our solution doesn't unnecessarily tie the non-HLSL parts to it.
 
 ### Root Signature Grammar
 
+The root signature DSL is defined using a slightly modified version of Extended
+Backus-Naur form. We define the additional symbol `:` to denote a
+comma-seperated list of components in any order:  `A : B = (A ',' B | B ',' A)`.
+Additionally, all keywords and enums are case-insensitive.
```
bogner, line 160: Can't we define constructions like the following to do this kind of thing without extending EBNF?
```
value = 'A' | 'B' | 'C';
list = value, {',', value};
```
I suspect that will make this a little bit easier to read.

However, it probably makes sense to do that in a follow up as reading the diff of the grammar will be easier if it's NFC

bogner (approved):

proposals/0002-root-signature-in-clang.md
inbelic: Not quite, as the new notation tries to capture that they can be specified in any order and only once. Which is maybe also okay. Either way, we can move that discussion to the follow-up NFC.

joaosaffran (approved):
LGTM

PR #196, opened 2025-03-28
Implicit resource bindings design

### LINKED ISSUE #177: Implicit resource bindings design
Once #176 is fleshed out, we need a design for implementing implicit resource bindings in Clang. We need to decide whether we are going to try to match DXC behavior or not, what information will be needed at various points in the compiler, and where will the implicit bindings and diagnostic take place.

There are two approaches to this:
1. Mark the bindings as "implicit" and calculate their actual locations later in the compiler
2. Fill out the bindings as we go.

Approach (2) would be a departure from DXC. There is some discussion of this starting in https://github.com/llvm/llvm-project/issues/58051#issuecomment-2239952237 but we need to discuss this in further detail if we want to pursue it. We would probably need to do some reachability analysis if we wanted to do this.

Some parts of approach (1) are probably necessary regardless of whether we adopt approach (2), because explicitly unbound resources exist in library shaders.

hekota:
Proposed design for implicit resource bindings in Clang.

Closes #177


bogner (approved):

proposals/0000-implicit-resource-binding.md
```diff
@@ -356,10 +356,129 @@ part of the HLSL language.
...
+
+The pass will scan the module to gather all instances of
+`@llvm.dx.resource.handlefromimplicitbinding` calls and sort them by
+`%order_id`. Then it will process the list sequentially based on the
+`%order_id`, and for each group of calls operating on the same unique resource
+it will find an available resource binding based on the required resource class,
+range and space. It will then replace all of the
+`@llvm.dx.resource.handlefromimplicitbinding` calls and all of its uses with new
+`@llvm.dx.resource.handlefrombinding` calls and add the new binding to the
+`DXILResourceBindingAnalysis` map.
```
V-FEXrt, line 480: How will this effect Increment/Decrement counter? Specifically, the need to uniquely identify a resource from the handle. (https://github.com/llvm/llvm-project/issues/125126)

Right now we inspect the `handlefrombinding`  calls which will happen during `DXILResourceBindingAnalysis` (after the changes from https://github.com/llvm/wg-hlsl/blob/main/proposals/0022-resource-instance-analysis.md are implemented) which means it will see `handlefromimplicitbinding` instead.

Doesn't `handlefromimplicitbinding` have any meaningful impact on uniqueness?

V-FEXrt (approved):

proposals/0000-implicit-resource-binding.md
hekota: The `DXILResourceImplicitBindings` pass needs to run before any other passes that work with resources. There are two options how to handle the interaction with `DXILResourceBindingAnalysis`:
1. The pass will use the `DXILResourceBindingAnalysis` and update it with the newly assigned bindings
2. The pass will do its own collecting of existing `@llvm.dx.resource.handlefrombinding` calls just to create a "map" of what registers are occupied. The `DXILResourceBindingAnalysis` can run as requested by the follow-up passes after `DXILResourceImplicitBindings` pass is completed.

I think the correct way to do this is #2 because the `DXILResourceBindingAnalysis` is no longer going to be just about bindings but it will analyze the counter usage and other things. I will update the doc.

V-FEXrt: sounds good thanks!

PR #197, opened 2025-03-28
Resource constructors proposal

### LINKED ISSUE #209: [HLSL] Design of constructors for resource handle initialization
> This needs design / planning work before we can estimate it.

Resource handle initialization should be done via resource class constructors. If a resource is declared in a global variable, the compiler should make sure to call a constructor that takes binding information, which will later be translated to `dx.handle.fromBinding`.

Resources declared as local variables would use a default constructor that will initialize the handle to 'empty' or 'null', or simply leave it undefined. Another resource can be assigned to these local variables and a copy constructor and/or assignment operator will take care of copying the resource handle.

This should greatly simplify handling of resources, including when they are dynamically bound, used as function in/out parameters, declared in unbounded arrays or used defined types.

Resource classes should look like this (although at this point we have to implement it in HLSLExternalSemaSource since it uses features HLSL doesn't have yet):

```
struct Binding;

template <typename T> struct RWBuffer {
private:
  using handle_t = __hlsl_resource_t
      [[hlsl::contained_type(T)]] [[hlsl::resource_class(UAV)]];
  handle_t h;

public:
  // For uninitialized handles
  RWBuffer() { h = __builtin_hlsl_create_poison_handle(); }
  
  // Resources are copyable.
  RWBuffer(RWBuffer &LHS) = default;
  
  // For handles constructed as globals with binding information, or from the
  // descriptor heap.
  RWBuffer(Binding &B) { h = __builtin_create_handle_from_binding(B); }

  ~RWBuffer() = default;

  RWBuffer &operator=(RWBuffer &LHS) {
    h = LHS.h;
    return *this
  }
...  
}
```

Follow-up from llvm/llvm-project#105076.

Open Questions:

- [ ] What does __builtin_hlsl_create_poison_handle() actually do?
- [ ] How does `struct Binding` work? Maybe constructor should just take register info.
- [ ] How will this work for SPIR-V?
- [ ] How does this interact with llvm/wg-hlsl#127 and llvm/wg-hlsl#175

hekota:
Proposal for resource classes constructors.

Fixes #209


hekota:
> So I've tried to understand this, and maybe I've been reading it too quickly, but I don't understand how any of these non-default constructors would get called?

Right, I did not mention that. Only the default constructor call gets generated automatically by Clang. The others will be set up by Sema after processing the resource binding attribute, or a lack of it. Basically it will be changing:
```RWBuffer<float> A : register(u3)```
into
```RWBuffer<float> A(Binding(0,3,1,0));```

**Comment by damyanp** [comment]
Ah, thanks, that helps!

> The others will be set up by Sema after processing the resource binding attribute, or a lack of it.

What will Sema use to identify things that need to be fixed up in this way?

**Comment by hekota** [comment]
> What will Sema use to identify things that need to be fixed up in this way?

Sema detects whether a global variable is resource class by checking if the first field of the declaration is a resource handle.

File: proposals/NNNN-resource-constructors.md [hunk 1]
```diff
@@ -0,0 +1,228 @@
...
+
+`RWBuffer<float> A : register(space13);`
+
+with
+
+`RWBuffer<float> A(13,1,0);`.
+
+### Copy constructor and assignment operator
+
+The copy constructor and/or the assignemt operator will take care of copying the
```
bob80905, line 161: nit: assignment

proposals/NNNN-resource-constructors.md
```diff
@@ -0,0 +1,228 @@
...
+      [[hlsl::contained_type(T)]] [[hlsl::resource_class(UAV)]];
+  handle_t __handle;
+
+  // For resources with explicit binding
+  RWBuffer(unsigned register, unsigned space, int rage, unsigned index) {
+    __handle = __builtin_hlsl_resource_createhandlefrombinding(__handle, register, space, range, index);
+  }
+
+  // For resources with implicit binding
+  RWBuffer(unsigned space, int rage, unsigned index) {
```
bob80905, line 200: should the space param be `unsigned space = 0` to express that the default space arg is 0 if not specified in the declaration?

proposals/NNNN-resource-constructors.md
```diff
@@ -0,0 +1,228 @@
...
+  RWBuffer() {
+    __handle = __builtin_hlsl_resource_createpoisonhandle(__handle);
+  }
+
+  // Resources are copyable.
+  RWBuffer(RWBuffer &LHS) = default;
+
+  // Resources are assignable.
+  RWBuffer &operator=(RWBuffer &LHS) {
+    __handle = LHS.__handle;
```
bob80905, line 215: I think __handle is the only field in RWBuffer, so I don't understand why this function can't also be `= default;`'d?

proposals/NNNN-resource-constructors.md
hekota: This is private constructor that will never be used directly from the shader, so having a default value for an argument does not really help anything. Sema will create call to this constructor when it needs to initialize a resource without binding and it will always provide the correct default space.

bob80905 (approved):
Looks good to my untrained eye

proposals/NNNN-resource-constructors.md
```diff
@@ -0,0 +1,222 @@
...
+
+  // Resources are assignable.
+  RWBuffer &operator=(RWBuffer &LHS) = default;
+  ...
+};
+```
+
+### Dynamically-bound resources
+
+TBD
```
llvm-beanz, line 179: I think for this case we probably can have something where `ResourceDescriptorHeap` is an array of some magic "untyped handle", and we have a constructor that constructs from that untyped handle to a resource with a different builtin to initialize the handle.

That doesn't need to be designed for this document, but I think it follows well with the patterns here.

llvm-beanz (approved):
I like this.

I was a little skeptical about having the constructor calls in the AST, but honestly, this is great.

PR #214, opened 2025-04-11
Resource binding: add another example and notes about resources in structs

hekota:
Add one more example showing binding of resource arrays in user-defined structs.
Add a note to re-evaluate whether to bind resource arrays in structs individually or as a continuous descriptor range later on when we start working on the resources in structs design (llvm/wg-hlsl##212).

### DIFF
--- a/proposals/0024-implicit-resource-binding.md
+++ b/proposals/0024-implicit-resource-binding.md
@@ -176,9 +176,9 @@ void main() {
 ```
 https://godbolt.org/z/5rf47heTe
 
-This example shows that when if we reserve a continuous descriptor range for
-resource array in structs we might break existing code that uses explicit
-binding and relies on the current behavior.
+This example shows that if we reserve a continuous descriptor range for resource
+array in structs we might break existing code that uses explicit binding and
+relies on the current behavior.
 
 Array resources inside structs are also not allowed to use dynamic indexing. In
 the following example DXC reports error when the resource array `B` is accessed
@@ -420,7 +420,7 @@ range of descriptiors should be reserved, the same was it is reserved for
 resource arrays declared at global scope. DXC currently treats these as
 individual resources (see [Example 2.2](#example-22)).
 
-There is a chance that this break existing user code that relies on the the
+There is a chance that this breaks existing user code that relies on the
 current DXC behavior, even if it uses only explicit binding (see [Example
 2.3](#example-23)). We should re-evaluate the impact of this decision when we
 start working on the support of resources in user-defined structs.
--- a/proposals/0024-implicit-resource-binding.md
+++ b/proposals/0024-implicit-resource-binding.md
@@ -137,7 +137,6 @@ struct S {
     RWBuffer<float> E[3]; // s.E.2 gets u0
 };                        // s.E.1 gets u4
 
-
 RWBuffer<float> A : register(u2);
 S s;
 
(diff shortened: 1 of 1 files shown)


proposals/0024-implicit-resource-binding.md
```diff
@@ -148,16 +147,44 @@ void main() {
...
+This seems wrong. Resource arrays should be bound to a continuous descriptor
 range and the range should be reflected in the `createHandleFromBinding`
 arguments. It also makes the binding susceptible to change whenever the code
 changes and is not something that can be relied upon. 
 
+Another interesting case is when a resource array inside a struct has explicit
+binding but not all of the array elements are used. DXC seems to pre-assign the
+register slots to the array elements, but does not actually reserve them, which
+seems like a bug. So when a particular array element is not used, its register
+slot can be (implicitly or explicitly) assigned to another resource.
```
bob80905, line 159: Though this is only speculation, I suspect DXC does the preassignment so that:
s1.0.D1 -> u1
s1.0.D2 -> u2 (unused)
s1.0.D3 -> u3 (unused)

Then s2 comes along so that:
s2.0.D1 -> u4
s2.0.D2 -> u5 (unused)
s2.0.D3 -> u6 (unused)


Then B comes along and takes u2 which is unused.
But DXC counts it as a shift and so the compiler "shifts up"  the unused slots in a later pass so that
s1.0.D2 -> u3 (unused)
s1.0.D3 -> u4 (unused)
s2.0.D1 -> u5
s2.0.D2 -> u6 (unused)
s2.0.D3 -> u7 (unused)

And though these aren't actually reserved, it occupies available slots on where to assign new resources.
So I would guess declaring an s3 would start at u8. Up to you if you want to include this speculation or not.

proposals/0024-implicit-resource-binding.md
```diff
@@ -148,16 +147,44 @@ void main() {
...
+RWBuffer<float> B; // gets u2
+
+[numthreads(1,1,1)]
+void main() {
+  A[0] = s1[0].D[1][0] + s2[1].D[1][0] + B[0];
+}
+```
+https://godbolt.org/z/5rf47heTe
+
+This example shows that when if we reserve a continuous descriptor range for
```
bob80905, line 179: nit: `that if we reserve a continuous descriptor range for resource arrays...`

proposals/0024-implicit-resource-binding.md
```diff
@@ -393,6 +420,11 @@ range of descriptiors should be reserved, the same was it is reserved for
 resource arrays declared at global scope. DXC currently treats these as
 individual resources (see [Example 2.2](#example-22)).
 
+There is a chance that this break existing user code that relies on the the
```
bob80905, line 423: nit: `that this breaks existing user code that relies on the current`

proposals/0024-implicit-resource-binding.md
hekota: I don't think the compiler shifts up anything. It honors the explicit assignment and reserves the slots for s1 and s2 like this:
```
s1.0.D0 -> u0 (unused by s1, used by A)
s1.0.D1 -> u1
s1.0.D2 -> u2 (unused)
s1.1.D0 -> u3 (unused)
s1.1.D1 -> u4 (unused)
s1.2.D2 -> u5 (unused)

s2.0.D0 -> u1 (unused)
s2.0.D1 -> u2 (unused)
s2.0.D2 -> u3 (unused)
s2.1.D0 -> u4 (unused)
s2.1.D1 -> u5 
s2.1.D2 -> u6 (unused)
```
And then `B` gets the first unused slot, which is`u2`.
BTW In you comment you are indexing the `D` array from 1 instead of 0 and that might have mixed things up.

bob80905: This makes me wonder why s2.0.D0 in your example starts at u1. You might think "it starts at u1 because u0 is used".
But then, why does s1.0.D0 start at u0? u0 is used by A. That's determined way before everything else. It's not consistent. With that logic, s1.0.D0 should start at u1 instead.
So, I do suspect s1.0.D0 would start with u1. Which is what I meant to write in my comment using my 1-indexing.
Then the way I could see it working out is that all of s1.1 is left unassigned since none of it is used, and s2.0 starts where s1.1 would have started (since none of any of the elements in s1.1 are used), shifted up by one because B occupies an unused slot.
Though I do understand that the point is to describe DXC's behavior, and the final binding output is what should matter here. But establishing an understanding might be helpful so we know what should be done in more crazy cases.
In your example, it's also strange that s2.0.D0 -> u1, since in fact, it is used previously by s1.0.D1.

hekota: `s1.0.D0` starts at `u0` and `s2.0.D0` starts at `u1` because they have explicit binding specifying that:
```
S s1[2] : register(u0);
S s2[2] : register(u1);
```

bob80905: Ah, that was helpful. Got confused with binding rules given conflicting explicit bindings. Well, consider this resolved.

bob80905 (approved):

bogner (approved):

PR #229, opened 2025-04-23
Update issue_tracking.md for latest thinking

damyanp:
Describe epics, scenarios, deliverables and tasks, and the custom fields used in the HLSL Support project.


docs/issue_tracking.md
```diff
@@ -2,62 +2,137 @@
...
+Issues can be blocked for some different reasons:
+
+* PR - waiting for a PR to complete
+* Refinement - waiting for agreement that it is ready to go from Planning to
+  Ready
+* Blocked - waiting for some other dependency before work on this one can
+  continus.
+
+When marking an issue as blocked we try and add a comment on the issue
+explaining why.
```
bogner, line 75: The "PR" and "Refinement" blocked states explain themselves, so I assume that this is saying we do this when we use the generic "Blocked", but that isn't entirely clear to me from the text.

bogner (approved):

docs/issue_tracking.md
damyanp: Updated to:

> When marking an issue as "Blocked" we try and add a comment on the issue
explaining why.

inbelic (approved):

PR #287, opened 2025-06-06
[NNNN] Implementation for `vk::constant_id`

s-perron:
Design for the Clang implmentation of`vk::constant_id`.

### DIFF
--- a/proposals/NNNN-vk-const-id.md
+++ b/proposals/NNNN-vk-const-id.md
@@ -1,0 +1,96 @@
+<!-- {% raw %} -->
+
+# SPIR-V Specialization Constants
+
+*   Proposal: [NNNN](NNNN-vk-const-id.md)
+*   Author(s): [Steven Perron](https://github.com/s-perron)
+*   Status: **Design In Progress**
+
+*During the review process, add the following fields as needed:*
+
+*   PRs:
+
+    [#143179](https://github.com/llvm/llvm-project/pull/143179)
+    [#143180](https://github.com/llvm/llvm-project/pull/143180)
+
+## Introduction
+
+SPIR-V has a feature called
+[Specialization Constants](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#_specialization_2)
+that allows shader code to have constants that are set by the driver. This
+feature is exposed in HLSL via the `vk::constant_id` attribute.
+
+## Motivationus
+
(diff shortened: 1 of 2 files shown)


proposals/NNNN-vk-const-id.md
```diff
@@ -0,0 +1,109 @@
...
+    [#143180](https://github.com/llvm/llvm-project/pull/143180)
+
+## Introduction
+
+SPIR-V has a feature called
+[Specialization Constants](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#_specialization_2)
+that allows shader code to have constants that are set by the driver. This
+feature is exposed in HLSL via the `vk::constant_id` attribute.
+
+## Motivationus
```
Keenuts, line 23:
```suggestion
## Motivations
```

proposals/NNNN-vk-const-id.md
```diff
@@ -0,0 +1,109 @@
...
+During code generation, this built-in will be replaced with a call to the
+existing SPIR-V builtin, `__spirv_SpecConstant`, which has the same semantics:
+
+```
+llvm @_ZL11my_constant = internal addrspace(10) global float 0.000000e+00, align 4
+...
+define internal spir_func void @__cxx_global_var_init() #4 {
+entry:
+  %0 = call token @llvm.experimental.convergence.entry()
+  %1 = call reassoc nnan ninf nsz arcp afn float @_Z20__spirv_SpecConstantif(i32 0, float 1.000000e+00)
```
Keenuts, line 68: I don't think the `nnan`, `ninf`, `nsz` flags should be there.
As for `arcp` and `afn` not sure if those should be there without requiring fastmath no?

s-perron: I just copied what was being created by my implementation. I agree those should not be there, but I wonder how they got there. I'll try to figure that out, and update accordingly.

s-perron: This seems to be coming from a general option in the compiler. All function in the llvm-ir have the attributes `"no-infs-fp-math"="true" "no-nans-fp-math"="true" "no-signed-zeros-fp-math"="true"`. I'll remove them from the description here, so that it is not assumed that the must be there.

I won't change the compiler implementation. This is an orthongal issue.

Keenuts (approved):

PR #293, opened 2025-06-06
Resource arrays in DXC 

hekota:
Initial part of the proposal for resource arrays that summarizes what constructs are supported by DXC.

### DIFF
--- a/proposals/NNNN-resource-arrays.md
+++ b/proposals/NNNN-resource-arrays.md
@@ -1,0 +1,258 @@
+<!-- {% raw %} -->
+
+# Resource arrays
+
+* Proposal: [NNNN](NNNN-resource-arrays.md)
+* Author(s): [Helena Kotas](https://github.com/hekota)
+* Status: **Design In Progress**
+
+## Introduction
+
+HLSL Resources are runtime-bound data that are provided as input, output, or
+both to shader programs. Shader authors can declare resources as individual
+variable declarations or as arrays of resources at local or global scope and
+inside user-defined structures.
+
+This document focuses on resource arrays declared at global scope of the shader
+or as local variables and function arguments. It includes an analysis what is
+currently supported in DXC and a proposal on how to implement that in Clang, and
+any potential changes between Clang and DXC. Resources declared inside
+user-defined structs will be covered in a separate proposal
+([#175](https://github.com/llvm/wg-hlsl/issues/175)).
+
+## Motivation
+
+Resource arrays are fundamental part of the HLSL language. We need to support
+them in Clang.
+
+## Current behavior in DXC
+
+For simplicity the examples below use just the `RWBuffer<float>` resource type,
+but same rules apply to other resource types as well.
+
+### Resource Arrays at Global Scope
+
+#### Fixed-size One-dimensional Array
+
+A group of resources can be declared as one-dimensional array of fixed size:
+
+```
+RWBuffer<float> A[4] : register(u10);
+```
+
+Each element of the array corresponds to a single resource that is initialized
+by an individual `@dx.op.createHandleFromBinding` DXIL intrinsic call. The fact
+that this resource is part of an array is encoded in the `dx.types.ResBind`
+argument of the call. The `dx.types.ResBind` structure contains binding
+information for the whole array - the lower bound and upper bound values for the
+register slots and a virtual register space. Third argument of
+`@dx.op.createHandleFromBinding` is an index of the individual resource in the
+descriptor range from the range `<lower bound, upper bound>` (inclusive).
+
+For example for the resource `A` declared above that is bound to register `u10`,
+the DXIL call to initialize resource handle for `A[2]` will have lower bound
+value `10`, upper bound `13`, space `0` and array index of `2`:
+```
+call %dx.types.Handle @dx.op.createHandleFromBinding(i32 217, %dx.types.ResBind { i32 10, i32 13, i32 0, i8 1 }, i32 2, i1 false),
+```
+
+Initialization of other resources in the same array would differ only by the
+index. The values for upper and lower bound and register space will stay the
+same.
+
+https://godbolt.org/z/ff3bnKxov
+
+#### Fixed-size Multi-dimensional Array
+
+Groups of resources can also be declared as multi-dimensional arrays of fixed
+size
+
+```
+RWBuffer<float> B[4][4] : register(u2);
+RWBuffer<float> C[2][2][5] : register(u10, space1);
+```
+
+When initializing individual resources of multi-dimensional array the arguments
+of `@dx.op.createHandleFromBinding` look like as if the array was flattened to a
+single dimension.
+
+That means the array `B[4][4]` is treated as if it was declared as `B[16]`, and
+`C[2][2][5]` as if it was `C[20]`.
+
+For example a call to initialize `C[1][0][3][0]` from the array declared above
+would have lower bound value `10`, upper bound `29`, space `1`, and array index
+`23` = `10 + 1*2*5 + 0*5 + 3`:
+```
+%dx.types.Handle @dx.op.createHandleFromBinding(i32 217, %dx.types.ResBind { i32 10, i32 29, i32 1, i8 1 }, i32 23, i1 false), 
+```
+
+https://godbolt.org/z/8MMWheeGc
+
+#### Arrays of Unbounded Size
+
+Groups of resource arrays can be declared with undefined size. These are called
+_unbounded arrays_:
+```
+RWBuffer<float> D[] : register(u5);
+```
+For unbounded arrays the `@dx.op.createHandleFromBinding` calls to initialize
+individual resource handles will have upper bound set to `-1`. For example an
+initialization of handle `D[100]` from the `D` array declared above looks like
+this:
+```
+%dx.types.Handle @dx.op.createHandleFromBinding(i32 217, %dx.types.ResBind { i32 5, i32 -1, i32 0, i8 1 }, i32 105, i1 false)
+```
+https://godbolt.org/z/hx99s5dzP
+
+#### Unbounded-size Arrays and Multiple Dimensions
+
+For multi-dimensional arrays of resources only the first dimension can be
+unbounded:
+```
+RWBuffer<float> E[][4] : register(u5);
+```
+When initializing individual resources of an unbounded multi-dimensional array,
+the arguments of `@dx.op.createHandleFromBinding` again look like as if the
+array was flattened to a single dimension. For example the call to create
+resource handle `E[100][4]` from array `E` declared above will have index `409`
+= `5 + 4 x 100 + 4`:
+```
+call %dx.types.Handle @dx.op.createHandleFromBinding(i32 217, %dx.types.ResBind { i32 5, i32 -1, i32 0, i8 1 }, i32 409, i1 false),
+```
+https://godbolt.org/z/YrEnYhrEj
+
+Arrays with unbounded sizes in the other dimensions are not allowed, though the
+error message DXC reports does not clearly indicate this:
+
+```
+RWBuffer<float> F[4][];
+
+error: array has incomplete element type 'RWBuffer<float> []'
+```
+https://godbolt.org/z/osq7dzvdh
+
+### Resource Arrays at Local Scope
+
+#### Local Resource Arrays with Fixed Size
+
+Resources array with fixed size can be declared as local variables in a function.
+Resource handles for the individual resources in the array are uninitialized
+until they are assigned a value from an initialized resource. DXC does not
+report error when an uninitialized resource is accessed and instead creates a
+call to `dx.op.annotateHandle` with zero-initialized handle value.
+
+```
+RWBuffer<float> Buf;
+RWBuffer<float> Out;
+
+float foo() {
+  RWBuffer<float> LocalArray[3];
+  LocalArray[2] = Buf;
+  // LocalArray[0] is not initialized
+  return LocalArray[2][0] + LocalArray[0][0];
+}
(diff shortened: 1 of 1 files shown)


proposals/NNNN-resource-arrays.md
```diff
@@ -0,0 +1,258 @@
...
+}
+
+[numthreads(4,1,1)]
+void main() {
+  Out[0] = foo();
+}
+```
+https://godbolt.org/z/43rGnY4Ee
+
+#### Resource Arrays as Function Arguments
```
llvm-beanz, line 190: A few notes on parameter passing as we implemented it in Clang:

`IncompleteArrayType` arguments (aka arrays of unknown size) are passed as pointers via array->pointer decay like C/C++. This is easy for resources because that's just the behavior we want anyways.

`ConstantArrayType` arguments go through the "HLSL array r-value" cast which effectively copy-initializes a temporary array creating an `ArrayParameterType`.

IMO, this is the "correct" behavior, because it allows passing an array of resources into a function and that array behaves just like any other array argument to a function in HLSL where it is copied in and should not mutate the original array. This is also an area where DXC has a lot of bugs...

For example consider this shader:

```hlsl
RWBuffer<int> X : register(u0, space0);
RWBuffer<int> Y : register(u1, space0);

void SomeFn(RWBuffer<int> Arr[2], uint Idx, int Val0) {
  Arr[0] = Y;
  Arr[0][Idx] = Val0;
}

[numthreads(1,1,1)]
void main(uint GI : SV_GroupIndex) {
  RWBuffer<int> Arr[2] = {X, Y};
  SomeFn(Arr, GI, 1);
  Arr[0][GI] = 2;
}
```
[Compiler Explorer](https://godbolt.org/z/PM9Mv4KP3)

I would expect this shader to write 1 to `Y` and 2 to `X`, but instead both writes go to `Y`, so the end result is that `X` is unused.

I suspect we're going to find a number of behaviors around resource arrays where what DXC does today doesn't align with any consistent language behavior, so we may need to expect behavior changes between DXC and Clang. This might inform some design decisions to bias us toward source breaking behaviors rather than behavior changing.

(tagging @tex3d in case he has additional thoughts)

proposals/NNNN-resource-arrays.md
```diff
@@ -0,0 +1,258 @@
...
+this:
+```
+%dx.types.Handle @dx.op.createHandleFromBinding(i32 217, %dx.types.ResBind { i32 5, i32 -1, i32 0, i8 1 }, i32 105, i1 false)
+```
+https://godbolt.org/z/hx99s5dzP
+
+#### Unbounded-size Arrays and Multiple Dimensions
+
+For multi-dimensional arrays of resources only the first dimension can be
+unbounded:
```
llvm-beanz, line 110: This is an oddity of the C array grammar. The dimension of a base array can be unbounded, but you can't have an array where the element size is unknown, so the element must be a complete type (hence the weird diagnostic). For your example below you're declaring an unbounded array of `RWBuffer<float>[4]` elements. It makes sense when you think about how multi-dimensional array indexing works, but isn't directly intuitive (at least it isn't to me).

proposals/NNNN-resource-arrays.md
```diff
@@ -0,0 +1,258 @@
...
+error message DXC reports does not clearly indicate this:
+
+```
+RWBuffer<float> F[4][];
+
+error: array has incomplete element type 'RWBuffer<float> []'
+```
+https://godbolt.org/z/osq7dzvdh
+
+### Resource Arrays at Local Scope
```
llvm-beanz, line 134: I don't know how exhaustive you're trying to be here, but you can also put a resource array (or multi-dimensional resource array) inside a struct.

proposals/NNNN-resource-arrays.md
```diff
@@ -0,0 +1,258 @@
...
+For multi-dimensional arrays of resources only the first dimension can be
+unbounded:
+```
+RWBuffer<float> E[][4] : register(u5);
+```
+When initializing individual resources of an unbounded multi-dimensional array,
+the arguments of `@dx.op.createHandleFromBinding` again look like as if the
+array was flattened to a single dimension. For example the call to create
+resource handle `E[100][4]` from array `E` declared above will have index `409`
+= `5 + 4 x 100 + 4`:
```
tex3d, line 118: We probably shouldn't use an out-of-bounds example.

proposals/NNNN-resource-arrays.md
tex3d: Yeah, array of resources in struct and array of struct containing resource(s) should probably be called out, unless we want a separate document for that?

proposals/NNNN-resource-arrays.md
```diff
@@ -0,0 +1,258 @@
...
+  Out[0] = foo();
+}
+```
+https://godbolt.org/z/Gdc3q674W
+
+#### Unbounded Local Resource Arrays
+
+DXC even allows unbounded resource arrays as local variables in a function. This
+seems wrong from a language point of view. How is the compiler supposed to
+allocate such variable?
```
tex3d, line 166: I strongly believe this should be considered a bug that DXC allows this.

proposals/NNNN-resource-arrays.md
tex3d: Yup, we don't copy-in the array that's modified in SomeFn like we should.

DXC normally does this in the `hl-legalize-parameter` pass (I know, the original Clang CodeGen doesn't reflect the copy-in for arguments like this), but skips this for arrays of objects/resources.  It appears this was done on purpose:
https://github.com/microsoft/DirectXShaderCompiler/blob/2084643e7ae44f23c80eb74d0cd549a8179c7443/lib/HLSL/HLLegalizeParameter.cpp#L261-L264

Perhaps this was done because otherwise it broke some real-world shaders that dynamically indexed arrays passed through layers of function calls through arguments, ultimately originating from some global resource array.  With all the copies, we might not be able to resolve the dynamic indexing back to the original resource range.  In any case, it's not correct, especially since it does pointer analysis to check whether you write to the array before inserting the copy.

proposals/NNNN-resource-arrays.md
```diff
@@ -0,0 +1,258 @@
...
+  Out[0] = foo();
+}
+```
+https://godbolt.org/z/43rGnY4Ee
+
+#### Resource Arrays as Function Arguments
+
+Both fixed-size and unbounded resource arrays can be used as function arguments.
+The dimensions of array passed into a function must match exactly the dimension
+of the declared argument
```
tex3d, line 194: I don't think we should continue to allow unbounded resource arrays as function arguments, unless there's a compelling use case.  If we did, I think they might be special, like only accepting another unbounded array, and requiring const inout (or const reference syntax when we have that).  This should restrict them to only ever coming from an unbounded global resource array, and we can then map them directly, without worrying about copy-in/copy-out.

proposals/NNNN-resource-arrays.md
hekota: Oops :)

proposals/NNNN-resource-arrays.md
hekota: I agree.

proposals/NNNN-resource-arrays.md
hekota: I've added this example to the document with a note that local arrays need to always be a copy.

tex3d (approved):

proposals/NNNN-resource-arrays.md
```diff
@@ -0,0 +1,356 @@
...
+When initializing individual resources of multi-dimensional array the arguments
+of `@dx.op.createHandleFromBinding` look like as if the array was flattened to a
+single dimension.
+
+That means the array `B[4][4]` is treated as if it was declared as `B[16]`, and
+`C[2][2][5]` as if it was `C[20]`.
+
+For example, a call to initialize `C[1][0][3][0]` from the array declared above
+would have lower bound value `10`, upper bound `29`, space `1`, and array index
+`23` = `10 + 1*2*5 + 0*5 + 3`:
```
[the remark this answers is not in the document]
spall: Should this example be 'C[1][0][3]', because that would be the resource handle that needs to be initialized right? The additionally zero index would access the first float in that resource?

proposals/NNNN-resource-arrays.md
```diff
@@ -0,0 +1,356 @@
...
+  Out[0] = foo();
+}
+```
+https://godbolt.org/z/Gdc3q674W
+
+#### Unbounded Local Resource Arrays
+
+DXC even allows unbounded resource arrays as local variables in a function. From
+a language point of view, this seems very wrong. How is the compiler supposed to
+allocate such variable? The fact that DXC allows this should be considered a bug.
```
spall, line 169: such variables

proposals/NNNN-resource-arrays.md
```diff
@@ -0,0 +1,356 @@
...
+}
+```
+https://godbolt.org/z/6Eh3fnME9
+
+Note that array `K` of size `3` is passed into function `foo` via an unbounded
+array argument and it is indexed beyond its size.
+
+#### Subsets of Multi-Dimensional Arrays
+ 
+Multi-dimensional array can be indexed to refer to a lower-dimensional subset of
```
spall, line 224: A Multi-dimensional array ...

proposals/NNNN-resource-arrays.md
```diff
@@ -0,0 +1,356 @@
...
+will have lower bound `0`, upper bound `49`, and index `38 = 0 + 7*5 + 3`:
+
+```
+call %dx.types.Handle @dx.op.createHandleFromBinding(i32 217, %dx.types.ResBind { i32 0, i32 49, i32 0, i8 1 }, i32 38, i1 false)
+```
+https://godbolt.org/z/YejsdsTKc
+
+At the same time, `P` is a local variable and as such it is editable. It should
+be initialized by copy-in array semantics to refer to resource handles from the
+original larger array `N`, but its individual array elements can be overridden,
```
spall, line 259: individual array elements refers to the resource handles? Or the floats in the resource the resource handle points to?

proposals/NNNN-resource-arrays.md
```diff
@@ -0,0 +1,356 @@
...
+
+This process quickly becomes complex because local arrays are moved using
+`memcpy`, and since they are editable, correctly resolving all resource array
+accesses is not always straightforward. Given this, the incorrect behavior in
+the above example is not surprising.
+
+#### Resource Arrays in User-defined Structs
+
+Resources and resource arrays can also be included in user-defined structs which
+can be declared at local or global scope. While this feature will covered by a
```
spall, line 302: will be covered

proposals/NNNN-resource-arrays.md
```diff
@@ -0,0 +1,356 @@
...
+
+[numthreads(4,1,1)]
+void main() {
+  Out[0] = s.P[2][0] + s_array[1].P[4][0];
+}
+```
+https://godbolt.org/z/Y8h3cWMov
+
+In DXC, when a user-defined struct contains an array of resources, each element
+of the array is treated as individual standalone resource instance. For `s.P[2]`
```
spall, line 329: as an individual

proposals/NNNN-resource-arrays.md
```diff
@@ -0,0 +1,356 @@
...
+in the example above the compiler creates a resource named `s.P.2` mapped to
+register `u12` with range `1`. Its handle initialization does not reflect that
```
spall, line 331: Does this affect the copy semantics if this resource array was passed into a function?

proposals/NNNN-resource-arrays.md
```diff
@@ -0,0 +1,356 @@
...
+A brief inspection of the DXC code generation and optimization pipeline shows
+that DXC allows Clang to handle resource arrays as if they were regular arrays
+of objects until after the LLVM optimization phase. Once all functions have been
+inlined in the generated code, DXC locates all references to resource arrays and
+replaces them with calls to initialize the resource handle.
+
+This process quickly becomes complex because local arrays are moved using
+`memcpy`, and since they are editable, correctly resolving all resource array
+accesses is not always straightforward. Given this, the incorrect behavior in
+the above example is not surprising.
```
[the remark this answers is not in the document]
hekota: Ok, I'll adjust the wording here.

proposals/NNNN-resource-arrays.md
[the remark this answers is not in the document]
hekota: The 23 is an index into the register/descriptor space. For resource from an array, it can have values between lower bound and upper bound inclusively. If this was a single resource bound to `u10`, the lower bound, upper bound and index would all have the same value of `10`.

I have updated the description to use "index within the range" instead of "array index" and removed the last `[0]` subscript that accesses the resource data.

proposals/NNNN-resource-arrays.md
hekota: As far as I know DXC is currently not doing any local array copies for function arguments (see Tex's comment https://github.com/llvm/wg-hlsl/pull/293#discussion_r2162743166).

proposals/NNNN-resource-arrays.md
```diff
@@ -0,0 +1,356 @@
...
+https://godbolt.org/z/8MMWheeGc
+
+#### Arrays of Unbounded Size
+
+Groups of resource arrays can be declared with undefined size. These are called
+_unbounded arrays_:
+```
+RWBuffer<float> D[] : register(u5);
+```
+For unbounded arrays the `@dx.op.createHandleFromBinding` calls to initialize
```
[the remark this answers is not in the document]
hekota: We don't have documentation for individual DXC errors.

PR #296, opened 2025-06-18
[0031] Proposal for shader semantics

Keenuts:
This is another proposal on how to implement semantic input in Clang, given DXIL & SPIR-V have drasticly different handlings, but some parts could be shared.

The POC for this implementation is in https://github.com/llvm/llvm-project/pull/149363

Another proposal exists: #112 which also suggest a sema change.

### DIFF
--- a/proposals/0031-semantics.md
+++ b/proposals/0031-semantics.md
@@ -1,6 +1,6 @@
 # HLSL shader semantics
 
-* Proposal: [0031](http://0031-semantics.md)
+* Proposal: [0031](0031-semantics.md)
 * Author(s): [Nathan Gauër](https://github.com/Keenuts)
 * Status: **Design In Progress**
 


proposals/NNNN-input-semantics.md
```diff
@@ -0,0 +1,245 @@
...
+sample, the parameter `e` uses `E`, which explicitly declares two fields with
+the same semantic. This is illegal. \
+`b` has the semantic `SB` applied on the whole struct. Meaning all its fields
+share the same semantic `SB`. This is legal because the duplication comes
+from inheritance.
+Lastly, `f` explicitly duplicates the semantic `FC`. But because those are
+shadowed by the semantic `SF`, this is valid HLSL.
+
+**Note**: Implicit semantic duplication is allowed for user semantics, but
+always forbidden for system semantics.
```
Keenuts, line 91: @tex3d @s-perron  this a rule I gathered from DXC's behavior. But I'm not sure that should be the case.
SPIR-V has no issue with this, but I feel it should be allowed on inputs, and disallowed on outputs:
 - the compiler should make sure the semantic appears once in the signature, and then duplicate the registers in the entrypoint wrapper.
 - for outputs, I feel this should be disallowed as the order of assignment is not defined.

proposals/NNNN-input-semantics.md
```diff
@@ -0,0 +1,245 @@
...
+
+The proposal is to provide a sorted list in `CGHLSLRuntime`:
+
+```cpp
+struct SemanticIO {
+  // The active semantic for this scalar/field.
+  HLSLAnnotationAttr *Semantic;
+
+  // Info about this field/scalar.
+  DeclaratorDecl *Decl;
```
Keenuts, line 223: This would be useful to handle the `[[vk::location(X)]]` attribute. Would be handled later, on top of this.

proposals/NNNN-input-semantics.md
```diff
@@ -0,0 +1,245 @@
...
+* Proposal: [NNNN](http://NNNN-input-semantics.md)
+* Author(s): [Nathan Gauër](https://github.com/Keenuts)
+* Status: **Design In Progress**
+
+## Introduction
+
+HLSL shaders can read form the pipeline state using [semantics](https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-semantics).
+This proposal looks into how to implement input semantics in Clang.
+Output semantics are out of scope of this proposal, but some parts will be
+similar.
```
tex3d, line 12: I don't think input semantics should be separated from output semantics here.

A better distinction would be between all the processing and diagnostics that we want to perform before clang code gen (in Sema), and how to implement shader inputs and outputs in the IR during or after clang codegen.

proposals/NNNN-input-semantics.md
tex3d: Incidentally, this example appears to fail with `-spirv` due to what looks like a lack of array-of-struct support.

> error: Multiple stage variables have a duplicated pair of location and index at 0 / 0

tex3d: Another interesting difference between DXIL and SPIR-V target is when you use multiple semantics - which is legal in HLSL!

https://godbolt.org/z/nsoeMoGG9
```
struct A {
  float x : X : Foo;
};

[shader("pixel")]
void main(A a, out float o : SV_Target) {
    o = a.x;
}
```

FXC and DXC with DXIL target uses the last one: `Foo`, while DXC with `-spirv` uses the first one: `X`.

Keenuts: > First, the full semantic used for the pipeline is always a combination of a case-insensitive semantic name and a semantic index. Semantic `seM` could be canonicalized to `SEM0` (these are exactly the same semantic functionally, even though the original identifier supplied is not the same, and no index was explicitly provided).

Ah interesting. So that's something the SPIR-V backend fails to do: `A` and `A0` are different semantics:
https://godbolt.org/z/KjsnTscaW


> If you write `SEM3`, the trailing numbers will be captured as the semantic index starting value. If a decl has multiple rows (array/matrix), or sub-fields (structs), semantic index is incremented after each row or field.

Also something we ignore in the SPIR-V backend (see above example, Location for A2 is 2 but A1 is 3.

> Additionally, if in one place you use `SEM` and another uses `SEM1`, these are not duplicate semantics (unless the decl with `SEM` expands to multiple fields or rows, thus assigning `SEM1`).
> 
> Taking into account this index, there should be no duplicate semantics on input or output (where the semantic is both the case-insensitive name prefix and the index). The same semantic used for an input and for an output is not a duplicate, of course, these are limited to within the set of inputs or within the set of outputs.
> 

From what I observe there is a difference between the way DXIL is handling system and user semantics on that point:
https://godbolt.org/z/rjY3eEjxT
In this case there is no index assignment, just implicit duplication of the semantic.


> On whether duplicates should be supported [..] and I could probably go on.

Right, that's right. So that's something we should refuse on the SPIR-V side too then.



>   * `inout` produces both an input and an output signature element for the parameter or each field (for struct params).

Oh, `inout` is allowed on semantic parameters? Interesting.

> Here's a slightly more complex example to illustrate: https://godbolt.org/z/759xEnqdq

Oh, another case the SPIR-V backend fails to handle even in DXC.

proposals/NNNN-input-semantics.md
```diff
@@ -0,0 +1,245 @@
...
+In this example:
+- `a` is linked to the semantic `SA`.
+- `b.b1` and `b.b2` are linked to the semantic `SB` because `SB` shadows the
+  semantics attached to each field.
+- `c.c1` has the semantic `SC1`, and `c.c2` the semantic `SC2`.
+- `d.d1`, hence `d`, is illegal: no semantic is attached to `d.d1`.
+- `e.e1` and `e.e2` are invalid: `EC` usage is duplicated without being inherited.
+- `f.f1` and `f.f2` semantic is `SF`, shadowing the duplicated `FC` semantic.
+
+**Note**: HLSL forbids explicit **non-shadowed** semantic duplication. In this
```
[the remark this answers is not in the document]
Keenuts: Yes, I'll rephrase this to say something like "active semantic" or "effective".

proposals/NNNN-input-semantics.md
tex3d: > From what I observe there is a difference between the way DXIL is handling system and user semantics on that point:
https://godbolt.org/z/rjY3eEjxT
In this case there is no index assignment, just implicit duplication of the semantic.

I'm not sure what you mean - it assigns index 0 and index 1. The first error even indicates the exact issue:
```
error: SV_Position semantic index exceeds maximum (0).
```

See what's being validated with `-Vd`
```
; Name                 Index   Mask Register SysValue  Format   Used
; -------------------- ----- ------ -------- -------- ------- ------
; SV_Position              0   xyzw        0      POS   float   xyzw
; SV_Position              1   xyzw        1      POS   float   xyzw
```

That "Index" column is the semantic index, which was incremented by field.

Keenuts: Sorry the link was building for DXIL, but the visible behavior is on the SPIR-V side:
- we create a single FragCoord builtin, hence having a duplicated semantic and no index change (unlike in DXIL)

tex3d (commented):
Partial review of comments, thanks for all this work!

Keenuts:
Thanks, rephrased mostly as suggested, reorganized a bit and added more intermediate examples, should be clearer

llvm-beanz (approved):
A couple comments, but I think this looks like a good direction to me.

proposals/NNNN-semantics.md
```diff
@@ -0,0 +1,516 @@
...
+// c    : A2 -> Location 0
+// d    : A0 -> Location 1
+// Semantic index does not contribute, only the parameter ordering does.
+// Input and Outputs are sorted independently.
+```
+
+It is also possible to explicitly set the index, using the
+`[[vk::location(/* Index */)]]` attribute. \
+Mixing implicit and explicit location assignment is **not legal**.
+Hence interaction between both mechanism is out of scope.
```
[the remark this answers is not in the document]
Keenuts: I don't think we need to implement this for now, but we might want to look into reflection helpers like `HLSLSemantic` or `HLSLCounter` decorations in the future.

proposals/NNNN-semantics.md
```diff
@@ -0,0 +1,518 @@
...
+System semantics are either translated to `Location` or `BuiltIn` decorated
+variables depending on the stage and semantic.
+
+In the example above, there are no system semantics, meaning every
+parameter would get a `Location` decorated variable associated.
+Each scalar field/parameter is associated with a unique index starting at 0,
+from the first parameter's field to the last parameter's field.
+Each scalar takes one index, and arrays of size N takes N indices.
+The semantic index does not impact the Location assignment.
+Indices are independent between `Input` and `Output` semantics.
```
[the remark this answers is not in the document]
Keenuts: From what I see, DXC assigned no Location at all when targeting library, and this causes validation issues.
See https://godbolt.org/z/jbf9zz5sE
I also think this should be thought more once we have a plan for libraries (same with decorations, capabilities, etc, multiple shaders in 1 spir-v module is a bit special)

Keenuts:
Thanks all for the review!

PR #298, opened 2025-06-18
Resource arrays design

### LINKED ISSUE #206: Resource arrays design
Create a proposal how to implement resource arrays in Clang.

hekota:
Initial proposal of the resource arrays design in Clang. 

Closes #206


llvm-beanz (approved):
This all seems correct and reasonable to me. @bogner, any concerns?

Keenuts:
Just to check: if I understand the proposal, the goal is to directly lower in codegen `buffer[index]` to a call to the `handlefrombinding` intrinsic. So seems OK.
But the part about local copies:

If we have:
```
void foo(RWStructuredBuffer<int> buffers[]) {
  buffers[0][0] = buffers[3][1];
}

void main() {
  foo(buffer);
}
```

Does this means we'll lower a `foo` function which takes a zero-sized array of target-specific types as parameter?
If we do, I think this should be OK as long as the inlining & optimizations remove all mention of `[0 x target("spirv.VulkanBuffer", ...)]` before we reach some parts of the backend

Keenuts (approved):
> The plan is to disallow incomplete arrays everywhere except in global resource declarations ([microsoft/hlsl-specs#141](https://github.com/microsoft/hlsl-specs/issues/141)). 

Ah cool, perfect, thanks! LGTM on our end

inbelic (approved):
LGTM, from our pair review.

Just some typos I found when reading through.

s-perron (approved):
This looks good to me. It should work well for SPIR-V. The only issue is with the text from previous commits. You seem inconsistent on whether the index include the binding or not. For example, at line 56 the index is 2. It does not include the base binding.  However, on line 86 the index is 23, and not 13. It include the base binding. This should be updated to not include the starting value for the binding. That is what I believe we agreed on in another proposal.

hekota:
> This looks good to me. It should work well for SPIR-V. The only issue is with the text from previous commits. You seem inconsistent on whether the index include the binding or not. For example, at line 56 the index is 2. It does not include the base binding. However, on line 86 the index is 23, and not 13. It include the base binding. This should be updated to not include the starting value for the binding. That is what I believe we agreed on in another proposal.

Good catch! For the DXIL op `dx.op.createHandleFromBinding` the index always includes the binding while for the `llvm.{dx|spv|.resource.handlefrombinding` intrinsics have zero-based index (see test [here](https://github.com/llvm/llvm-project/blob/c188e1db6a4c469f5ad92c4ce088fd113335410f/llvm/test/CodeGen/DirectX/CreateHandleFromBinding.ll#L42)). Since this document is referencing only the DXIL ops, I'll update the index on line 56 to be 12.

PR #311, opened 2025-07-07
Explicit Padding in CBuffers Proposal

bogner:


s-perron (commented):
There is one issue that I don't think has been addressed. That is type conversions. Consider this example: https://godbolt.org/z/xY4MshTr1.

```
struct S {
    float f[4];
};

RWStructuredBuffer<S> sb;

S s;

[numthreads(1,1,1)]
void main() {
  sb[0] = s;
}
```

From an HLSL perspective, the type stored in the structured buffer and the type stored in the cbuffer are the same type `S`, so you can assign one to the other. With this proposal, the type in the cbuffer will now be different than the type in the structured buffer in the AST. We will have to have some type of conversion from one to the other.

proposals/NNNN-explicit-padding.md
```diff
@@ -0,0 +1,191 @@
...
+- A well-known named type `%pad8`, defined as a named struct containing a
+  single `i8`. This is the simplest option but requires backends that are
+  interested in this type to participate in a secret handshake.
+- Target types such as `target("dx.pad8")` and `target("spirv.pad8")`. This is
+  somewhat awkward because the type isn't really tied to a target, but target
+  types need to be. Targets that don't need to differentiate between padding
+  and actual members could simply use `i8`.
+
+> TODO: Choose one of these three options and move the others to the
+> "alternatives" section.
```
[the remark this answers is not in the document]
s-perron: 1. If you use the target type, you can do something like we do for the `vk::SpirvType` where the size is a parameter. So the padding is always just one instance of the type, not an array. However, not a significant advantage.
2. The "well known name" solution has the problem that theoretically, the optimizer does not know the name is special, and might change it in some way. Let me know if there is something in the llvm-ir spec that would guarantee that it will not be changed.
3. The `pad8` is a great idea if it is accepted by the llvm community. But it could take a while.

My thoughts are to do the target type for now. See if the LLVM community is interested in `pad8`. If so, we can switch to it when it is added.

I do not have strong opinions on this, and I will not hold up this proposal if you do something different.

proposals/NNNN-explicit-padding.md
```diff
@@ -0,0 +1,175 @@
+<!-- {% raw %} -->
+
+# Explicit Padding in Structs and CBuffer Arrays
+
+* Proposal: [NNNN](NNNN-explicit-padding.md)
+* Author(s): [Justin Bogner](https://github.com/bogner)
+* Status: **Design In Progress**
```
damyanp, line 7:
```suggestion
---
title: "[NNNN] - Explicit Padding in Structs and CBuffer Arrays"
params:
    status: Design In Progress
    authors:
        - bogner: Justin Bogner
---

```

proposals/NNNN-explicit-padding.md
```diff
@@ -0,0 +1,175 @@
...
+- Target types such as `target("dx.pad8")` and `target("spirv.pad8")`. This is
+  somewhat awkward because the type isn't really tied to a target, but target
+  types need to be. Targets that don't need to differentiate between padding
+  and actual members could simply use `i8`.
+  
+We settled on a target type with a size parameter for now.
+
+[llvm-project/wg-hlsl#171]: https://github.com/llvm/wg-hlsl/pull/171
+
+<!-- {% endraw %} -->
```
damyanp, line 175:
```suggestion
```

proposals/NNNN-explicit-padding.md
```diff
@@ -0,0 +1,175 @@
...
+    authors:
+        - bogner: Justin Bogner
+---
+
+
+## Introduction
+
+We introduce an explicit padding type for HLSL, and construct cbuffer arrays
+and structs that are annotated with `packoffset` or `vk::offset` using this
+type to unambiguously lay out these objects.
```
hekota, line 14:
```suggestion
We introduce an explicit padding type for HLSL, and construct cbuffer arrays,
structs, and cbuffer elements that are annotated with `packoffset` or `vk::offset` using this
type to unambiguously lay out these objects.
```

hekota: Structs in cbuffer must also start on a 16-byte boundary and we need to insert padding before if needed: https://godbolt.org/z/YEjc7TEY9

proposals/NNNN-explicit-padding.md
```diff
@@ -0,0 +1,175 @@
...
+Introduce an explicit padding type and use this type for the padding in the
+various constructs that need it.
+
+The padding type will be a target type with a parameter for the size in bytes.
+For DirectX, this would look like `target("dx.Layout", 4)` for 4 bytes of
+padding. For SPIR-V, `target("spirv.Padding", 8)` for 8 bytes of padding. We
+may attempt to come up with a first-class type in LLVM for these purposes in
+the future.
+
+### Structs with annotations
```
hekota, line 49: I would probably move this section after "### Arrays in a cbuffer" because this is way less common than ararys in cbuffer.

proposals/NNNN-explicit-padding.md
```diff
@@ -0,0 +1,175 @@
...
+CBuffers in HLSL have very specific layout rules. Each "object" starts at a
+16-byte boundary, which is mostly explainable as a 16-byte alignment
+requirement, but applies to array elements rather than the array itself in a
+way that doesn't match the general language. We can emulate this with an array
+of objects that consist of a struct containing the element type and padding to
+16 bytes, followed by a single instance of the element type itself. See
+[CBuffer Padded arrays at the HLSL-level] for details.
+
+[CBuffer Padded arrays at the HLSL-level]: #cbuffer-padded-arrays-at-the-hlsl-level
+
```
hekota, line 84: Add section "### Struct in a cbuffer" (here or before the array section) describing that structs in cbuffer must also start on a 16-byte boundary and we need to insert padding before if as needed.

proposals/NNNN-explicit-padding.md
```diff
@@ -0,0 +1,175 @@
...
+  target("dx.Padding", 12"),
+  <{ [1 x <{ <3 x double>, target("dx.Padding", 8) }>], <3 x double> }>,
+  target("dx.Padding", 8),
+  [ 2 x <4 x i32> ],
+  <{ [3 x <{ half, target("dx.Padding", 14) }>], half }>
+}>
+```
+
+[__cblayout]: https://github.com/llvm/wg-hlsl/blob/4570a9cfc5c4b1e5bc0b773a6fb7b22014ac6d3b/proposals/0016-constant-buffers.md#lowering-constant-buffer-resources-to-llvm-ir "Lowering Constant Buffer Resources to LLVM IR"
+
```
hekota, line 136: Add example of cbuffer with struct that has padding before.

proposals/NNNN-explicit-padding.md
bogner: I found a separate section for this more confusing so I instead reworked this a bit to talk about scalars, vectors, structs, and arrays explicitly.

proposals/NNNN-explicit-padding.md
```diff
@@ -108,12 +117,32 @@ cbuffer cb0 : register(b0) {
...
+cbuffer cb1 : register(b0) {
+  int i; // offset   0,  size 4  (+12)
+  S s;   // offset  16,  size 4
+  int j; // offset  20,  size 4
+}
+
+```llvm
+%__cblayout_cb1 = type <{
+  i32, target("dx.Padding", 12),
+  i32,
```
hekota, line 135: Are we flattening the struct `S` here? I though we preserve it.

bogner: The struct is preserved. I've fixed the example.

hekota (approved):
LGTM!

PR #320, opened 2025-07-16
[0029] Proposal: Add Root Signature related driver options

inbelic:
DXC provided a number of command line options for working with root signatures. Namely: `Frs`, `Qstrip_rootsignature`, `setrootsignature`, `extractrootsignature`, `verifyrootsignature`, `force-rootsig-ver` and `rootsig-define`.

This pr proposes how the functionality of these options will be implemented into clang and documents the expected behaviour that they should provide.

### DIFF
--- a/proposals/0029-root-signature-driver-options.md
+++ b/proposals/0029-root-signature-driver-options.md
@@ -2,7 +2,7 @@
 
 # Root Signature Driver Options
 
-* Proposal: [NNNN](NNNN-root-signture-driver-options.md)
+* Proposal: [0029](0029-root-signture-driver-options.md)
 * Author(s): [Finn Plummer](https//github.com/inbelic)
 * Status: **Accepted**
 * Impacted Project(s): Clang


proposals/0002-root-signature-in-clang.md
```diff
@@ -936,6 +936,153 @@ Operands:
...
+as the only part
+
+ - Note: It is not possible to use `-rootsig-define` to overwrite which root
+signature will be used as the "entry" root signature.
+
+Design Notes: Commonly used. Planned in Clang.
+
+#### Option `-Qstrip-rootsignature`
+
+Strips the root signature attribute from the entry function.
```
bogner, line 1020: I find this wording a little confusing. I think something like "Omits the root signature part from the compiled shader" would be clearer.

proposals/0002-root-signature-in-clang.md
```diff
@@ -936,6 +936,153 @@ Operands:
...
+ - Parse and perform syntactic validations of "entry" root signature
+ - Perform the 'basic' sub-set of validations that are available without access
+to resource bindings
+ - Produces the DXIL container with the root signature part (RTS0) omitted
+
+Design Notes: Usable within .hlsl source files. Planned in Clang unless there is
+a signal of it being unused.
+
+Design Notes: Plan to match DXC behaviour directly such that we still do
+the parsing and basic validations of the root signataure.
```
bogner, line 1039: Should this be mentioned in the description of what the attribute does? Whether or not the root signature is validated and stripped late or ignored completely by the frontend is a fairly important detail here.

proposals/0002-root-signature-in-clang.md
```diff
@@ -936,6 +936,153 @@ Operands:
...
+
+```
+  -setrootsignature <DXIL Container>
+```
+
+Behaviour:
+
+ - Insert/overwrite the RTS0 part of the compiled DXIL Container with the RST0
+contents of the input container
+ - Does not perform any validation with the input root signature
```
bogner, line 1056: Does this still validate root signatures that were specified by the attribute (or some other way) and then drop them and overwrite them with the given signature, or does it ignore them? Why doesn't this validate the input root signature?

proposals/0002-root-signature-in-clang.md
```diff
@@ -936,6 +936,153 @@ Operands:
...
+ - Requires input to be a valid DXIL container with or without an RST0 part
+
+Design Notes: Planned in Clang. If compiling a stand-alone DXIL Container with
+just the RTS0 is useful then this is presumably also required in that workflow.
```
bogner, line 1060: I think we need to know more about when/why this is used to decide whether to plan this. I'm uncomfortable with the fact that we don't validate this at all.

proposals/0002-root-signature-in-clang.md
```diff
@@ -936,6 +936,153 @@ Operands:
...
+```
+
+Behaviour:
+
+ - Behaviour is the same as the
+[root signature target option](#target-root-signature-version), with the
+exception that it will run all validations with the resource bindings in the
+shader code
+
+Design Notes: Not planned in Clang, as it seems shadowed.
```
bogner, line 1080: What does "it seems shadowed" mean in this context?

bogner (commented):
Might be better to move this proposal into its own numbered file. Proposal 0002 is already very large, and this is largely orthogonal. A few questions and comments inline.

proposals/0002-root-signature-in-clang.md
inbelic: Does attribute refer to the root signature function attribute here?

In DXC, the root signature is parsed and does the basic validations for the root signature and is stripped late. The question is if we want to match that behaviour in clang.

proposals/0002-root-signature-in-clang.md
```diff
@@ -936,6 +936,159 @@ Operands:
...
+Behaviour:
+
+ - Parse and perform syntactic validations of "entry" root signature
+ - Perform the 'basic' sub-set of validations that are available without access
+to resource bindings
+ - Produces a DXIL container with only the serialized "entry" root signature
+as the only part
+
+ - Note: It is not possible to use `-rootsig-define` to overwrite which root
+signature will be used as the "entry" root signature.
```
tex3d, line 1017: Perhaps it should be noted why (uses `-E` instead)

```suggestion
signature will be used as the "entry" root signature, because this is specified
using the `-E` option when compiling directly to a root signature target.
```

proposals/0002-root-signature-in-clang.md
```diff
@@ -936,6 +936,159 @@ Operands:
...
+Design Notes: Commonly used. Planned in Clang.
+
+#### Target Root Signature Version
+
+Compiles the "entry" root signature, specified by the given macro expansion, to
+a DXIL Container with just the (version specific) RTS0 part.
+
+Usage:
+
+Given a defined macro in the source file (`#define RS "CBV(b0)"`)
```
tex3d, line 1000: Perhaps we should note that it's not necessary to specify the root signature macro in the source file, since it may be specified with a `-D` option on the command line instead.  DXC supports this, but doesn't accept an empty source file. The source file will not even be used in this case, so it could in theory be omitted, but DXC requires it, and it seems to require that it has at least 4 bytes of content (that's likely due to the BOM detection).

proposals/0002-root-signature-in-clang.md
```diff
@@ -936,6 +936,159 @@ Operands:
...
+as the only part
+
+ - Note: It is not possible to use `-rootsig-define` to overwrite which root
+signature will be used as the "entry" root signature.
+
+Design Notes: Commonly used. Planned in Clang.
+
+#### Option `-Qstrip-rootsignature`
+
+Omits the root signature part (RTS0) from the produced DXIL Container.
```
tex3d, line 1023: It also can be used to remove the root signature from an input DXIL container file, when that container file is specified along with the `-dumpbin` option.

proposals/0002-root-signature-in-clang.md
```diff
@@ -936,6 +936,199 @@ Operands:
...
+Design Notes: Usable within .hlsl source files. Planned in Clang unless there is
+a signal of it being unused.
+
+Design Notes: Plan to match DXC behaviour directly such that we still do
+the parsing and basic validations of the root signataure.
+
+#### Option `-setrootsignature`
+
+Overwrite the root signature part (RTS0) with the RTS0 of provided DXIL
+Container.
```
tex3d, line 1047: This can be used to override the root signature part used either in the resulting DXIL container when compiling a shader, or by adding or replacing the root signature part of an already compiled shader in a DXIL Container.

```suggestion
Override the root signature part (RTS0) in the DXIL Container produced either
when compiling a shader, or in a container loaded using the `-dumpbin` option.
```

proposals/0002-root-signature-in-clang.md
```diff
@@ -936,6 +936,199 @@ Operands:
...
+
+```
+  -setrootsignature <DXIL Container>
+```
+
+Behaviour:
+
+ - Insert/overwrite the RTS0 part of the compiled DXIL Container with the RST0
+contents of the input container
+ - Does not perform any validation with RTS0 of the provided DXIL Container
```
tex3d, line 1059: This does perform validation of the RTS0 with the shader you're attaching it to (making sure it supplies the required bindings).  Why is this statement here?

proposals/0002-root-signature-in-clang.md
```diff
@@ -936,6 +936,199 @@ Operands:
...
+ - Requires input to be a valid DXIL container with or without an RST0 part
```
tex3d, line 1060: Which input?  The input container for the `-setrootsignature` option is required to be a valid DXIL container *with* an `RTS0` part.

The shader input (whether compiling from HLSL or loading a DXIL Container) need not have a root signature attached.

If one is present in the shader input, that root signature must still be well-formed, but it will not be checked against the shader for binding compatibility, and it will not be used for the final DXIL Container output.

proposals/0002-root-signature-in-clang.md
```diff
@@ -936,6 +936,199 @@ Operands:
...
+ - Note: Even when provided, this will parse/validate a root signature
+specified as an attribute (or another above) before dropping it.
```
tex3d, line 1062: It will parse and check basic validity, but not check against the shader bindings, right?  Perhaps we should clarify that.

Plus, we could change this design to completely ignore the attribute when overriding the root signature instead (as a divergence from what DXC does).  In any case, we should be consistent between the `-rootsig-define` behavior and the `setrootsignature` behavior here.

proposals/0002-root-signature-in-clang.md
```diff
@@ -936,6 +936,199 @@ Operands:
...
+
+#### Option `-extractrootsignature`
+
+Specifies to compile the shader as normal but will only output the RTS0 part
+into the produced DXIL Container.
+
+Usage:
+
+```
+  -extractrootsignature /Fo <DXIL Container>
```
tex3d, line 1075: I wish there was a better way to structure things to indicate these cases that work both when compiling a shader and when loading a compiled shader in a DXIL Container using `-dumpbin`.

proposals/0002-root-signature-in-clang.md
```diff
@@ -936,6 +936,199 @@ Operands:
...
+Given a shader defined from a DXIL Container, perform validations with the
+root signature part (RTS0) of the other DXIL Container
+
+Usage:
+
+```
+  -dumpbin <Shader DXIL Container> -verifyrootsignatures <Root Signature DXIL Container>
+```
+
+Note: `dumpbin` is required to load a binary instead of performing compilation
```
tex3d, line 1100: This is great, but I wonder if we can't just have a `-dumpbin` section we refer to from each of the other options that can apply to a loaded DXIL container using this option.

proposals/0002-root-signature-in-clang.md
```diff
@@ -936,6 +936,199 @@ Operands:
...
+
+#### Option `-verifyrootsignature`
+
+Given a shader defined from a DXIL Container, perform validations with the
+root signature part (RTS0) of the other DXIL Container
+
+Usage:
+
+```
+  -dumpbin <Shader DXIL Container> -verifyrootsignatures <Root Signature DXIL Container>
```
tex3d, line 1097:
```suggestion
  -dumpbin <Shader DXIL Container> -verifyrootsignature <Root Signature DXIL Container>
```

proposals/0002-root-signature-in-clang.md
```diff
@@ -936,6 +936,199 @@ Operands:
...
+```
+
+Behaviour:
+
+ - Behaviour is the same as `-extractrootsignature`. Except that it will still
+output the complete DXIL Container as normal, and, produce an additional DXIL
+container with just the RTS0 part.
+
+Design Notes: Not planned in Clang. For similar motivations as
+`extractrootsignature`.
```
tex3d, line 1130: This option is often used along with the `-Qstrip-rootsignature` option to produce a shader container without the root signature and a separate serialized root signature file to be used separately with the runtime.

So, we aren't planning on providing any ability to extract a root signature into a separate container when compiling a shader?  Nor are we planning on providing the ability to extract a root signature from a shader container that has the RTS0 part?

I think these are both useful for developers.  The alternative is to compile a root signature directly to a root signature target.  But that requires knowing which root signature define goes with a particular entry point for the equivalent functionality (and having used a define in the first place when specifying the attribute, which isn't required).

PR #330, opened 2025-08-20
Publish a statically rendered version of this repo to github pages

damyanp:
This uses [Hugo](https://gohugo.io/) to generate the site.  A github workflow is added to generate and publish the pages.

Update the various proposals so that appropriate metadata is in front matter rather than markdown.


website/.hugo_build.lock
damyanp, line 1: This probably shouldn't be checked in.

website/content/proposals/_content.gotmpl
damyanp, line 1: I may be over indexing on this, but I _really_ don't like having to add front matter to the proposal .md files themselves.

99% of the time we're reading them from the .md rather than the rendered site and so we want the markdown to render well on its own.

website/hugo.toml
```diff
@@ -0,0 +1,47 @@
...
+
+[params]
+  menu_style = "slide-menu"
+  
+[markup]
+  [markup.goldmark]
+    [markup.goldmark.renderer]
+      unsafe = true
+
+ignoreLogs = ['warning-goldmark-raw-html']
```
damyanp, line 14: This sounds scary to me, but it allows us to use HTML tags in our documents without generating warnings, which can be reasonable escape hatch when markdown isn't sufficient.

Icohedron (commented):
nit: on mobile the navigation menu is places all the way at the bottom of the page. So if there is a page with a lot of content, it is very annoying to scroll to the bottom just to access navigation.

website/hugo.toml
Icohedron: For line breaks, HTML is not required. Other ways to line break in GitHub-flavored Markdown are:
- Ending a line with two spaces
- Ending a line with a backslash

Break this\
line!

Or break this  
line!

damyanp:
Rendered version here: https://damyanp.github.io/wg-hlsl/

Although once this PR completes I'm not sure I'll keep that one around.

.github/workflows/hugo.yml
```diff
@@ -0,0 +1,75 @@
+# Sample workflow for building and deploying a Hugo site to GitHub Pages
+name: Deploy Hugo site to Pages
+
+on:
+  # Runs on pushes targeting the default branch
+  push:
+    branches: ["damyanp/hugo"]
```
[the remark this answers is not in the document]
damyanp: It should be main (there's a fix for that)

llvm-beanz (approved):

PR #337, opened 2025-09-04
[NNNN] New proposal for feature profiles

llvm-beanz:
This proposal suggests a different approch for handling target feature profiles which is both more flexible and more meaningful across APIs. The goal is to redefine the concept of target profiles as DXC represented them to extend more clearly across the whole ecosystem.

### DIFF
--- a/proposals/NNNN-feature-profiles.md
+++ b/proposals/NNNN-feature-profiles.md
@@ -146,7 +146,7 @@ SM 6.0 and earlier.
 clang-dxc -T cs_Android15 ...
 ```
 
-Builds a compute shader for Android 15, targeting SPIRV 1.2 and allowing the
+Builds a compute shader for Android 15, targeting SPIRV 1.3 and allowing the
 required features supported by the Android 15 profile.
 
 ```


proposals/NNNN-feature-profiles.md
```diff
@@ -0,0 +1,166 @@
...
+  status: Under Consideration
+  sponsors:
+    - llvm-beanz: Chris Bieneman
+---
+
+## Introduction
+
+This proposal seeks to introduce a new more formalized way of specifying feature
+profiles as a set of features to allow the compiler to utilize when generating
+outputs. This proposal pairs stricter availability checks with explicit opt-in
```
damyanp, line 15: "Feature Profiles" is a new concept being added here?  The current first sentence feels a bit long and clunky.  Here's a stab at another way of wording it:

```suggestion
This proposal introduces feature profiles: a more formalized way to specify a set of features that the compiler can utilize when generating outputs. This proposal pairs stricter availability checks with explicit opt-in
```

proposals/NNNN-feature-profiles.md
```diff
@@ -0,0 +1,166 @@
...
+
+## Proposed solution
+
+The core of this proposal revolves around a new definition for "target profiles"
+that the shader compiler will translate to available feature sets.
+
+The target profile flag supported by DXC-compatible drivers as `-T` will be
+extended to support a new wider set of profile values. The current profiles will
+remain unchanged, which will imply a shader-model feature set and _all_ optional
+features.
```
damyanp, line 40: Since all optional features are included, I wonder if we should have a way to override and remove features.

I think this is implied below with the `+NativeLowPrecision`, example, suggesting that there's a matching `-NativeLowPrecision` option available.  If that's the case it might be nice to explicitly call it out.

damyanp (approved):
A couple of comments, but this looks ready to be merged and discussed further to me.

proposals/NNNN-feature-profiles.md
llvm-beanz: Oh, I should have added an example for that. The existing Clang model is `+` to enable `-` to disable. Will update.

bogner (approved):
This is in good shape to be merged and have further consideration in tree.

PR #345, opened 2025-09-22
[0023] Propose two-handle model for counter variables

s-perron:
This commit updates the proposal for handling counter variables in RWStructuredBuffer.

The key changes are:
- Propose a two-handle model (`__handle` and `__counter_handle`) to explicitly represent the main buffer and its counter.
- Introduce a `[[hlsl::is_counter]]` attribute to differentiate the counter handle at the type level.
- Propose new static methods (`__createFromBindingWithCounter` and `__createFromImplicitBindingWithCounter`) for initialization.
- Introduce new built-in functions (`__builtin_hlsl_resource_counterhandlefromimplicitbinding` and `__builtin_hlsl_resource_counterhandlefromexplicitbinding`) to create the counter handle.
- Detail how this approach will be handled in both the DirectX and SPIR-V backends.


proposals/0023-typed-buffer-counters.md
```diff
@@ -24,39 +24,167 @@ explicit that they are separate resources.
...
+modified to include two separate handles: one for main storage (`bufferHandle`)
+and one for the counter (`counterHandle`). The `bufferHandle` is initialized as
+usual via a call to `llvm.dx.resource.handlefrom*binding` or
+`llvm.spv.resource.handlefrom*binding`, depending on the target.
+
+To establish an explicit link between the resource and its counter, we propose
+introducing two new backend-specific intrinsics:
+
+*   `llvm.{dx|spv}.resource.counterhandlefromimplicitbinding(ResourceHandle,
+    order_id, space_id)`
```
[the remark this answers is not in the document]
s-perron: I just updated the design for handling array. Sine the size and index will have to match the size and index for the "main resource", we can get that information from the call defining that handle. Let me know if anything else seems to be missing.

proposals/0023-typed-buffer-counters.md
```diff
@@ -24,39 +24,211 @@ explicit that they are separate resources.
...
+    adding a new member to `HLSLResourceBindingAttr` to hold an explicit
+    counter binding, separate from the implicit one.
+4.  The original `HLSLCounterBindingAttr` will be discarded after its
+    information has been merged.
+
+This approach ensures that all downstream components, such as CodeGen, only
+need to inspect a single attribute (`HLSLResourceBindingAttr`) to get all
+necessary binding information for a resource and its associated counter. This
+provides a clean user-facing syntax while maintaining a simple and unified
+internal representation.
```
hekota, line 176: I think we will need to keep the `HLSLCounterBindingAttr` in the AST because the AST needs to preserve all source code constructs for rewriter scenarios. Which reminds me that we currently discard the 'HLSLVkBindingAttr` as well if we are not compiling for SPIR-V and we should leave it in.

I understand that dealing with 3 attributes when figuring out the binding is not ideal, especially since we need to handle them in multiple places. We could simplify that by having a shared helper class like this::

```
struct ResourceBindingAttrs {
  HLSLResourceBindingAttr *DxBinding;
  HLSLRVkBindingAttr *VkBinding;
  HLSLRVkCounterBindingAttr *VkCounterBinding;
  bool IsSpirv;

  ResourceBindingAttrs(Decl *D, bool IsSpirv) : IsSpirv(IsSpirv) {
    DxBinding = D->getAttr<HLSLResourceBindingAttr>();
    VkBinding = IsSpirv ? D->getAttr<HLSLRVkBindingAttr>() : nullptr;
    VkCounterBinding = IsSpirv ? D->getAttr<HLSLRVkCounterBindingAttr>() : nullptr;
  }
  
  bool hasExplicitBinding() const { return (DxBinding && DxBinding->hasRegisterSlot()) || VkBinding; }
  bool getBindingSlot() const { ... } ;
  bool getBindingSpace() const;
  bool getImplicitOrderID() const;
 
  bool hasExplicitCounterBinding() const;
  bool getCounterBindingSlot() const;
  bool getCounterImplicitOrderID() const;
  ...
}
```

proposals/0023-typed-buffer-counters.md
```diff
@@ -24,39 +24,211 @@ explicit that they are separate resources.
...
+
+### Initialization and Binding
+
+The core of this design lies in how these two handles are initialized. In Sema,
+new static methods will be added to the resource class to initialize them.
+
+1.  **New Static Methods:** Two new static methods will be added to
+    counter-enabled resource classes:
+    *   `__createFromBindingWithCounter`
+    *   `__createFromImplicitBindingWithCounter`
```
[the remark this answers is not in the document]
s-perron: When I was first writing this I was planning on not handling the explicit binding for counters. Then I partially added it because I needed to attribute for the array initialization. I believe I have gone through and added the explicit binding design. this also helped me realize I was not clear enough on what they function do. I've made it a little more verbose.

proposals/0023-typed-buffer-counters.md
```diff
@@ -24,39 +24,230 @@ explicit that they are separate resources.
...
+4.  **Counter Operations:** Methods like `IncrementCounter` and
+    `DecrementCounter` will be modified in `HLSLBuiltinTypeDeclBuilder.cpp` to
+    operate on the `__counter_handle` member instead of the `__handle` member,
+    directing the atomic operations to the correct resource.
+
+### Explicit Counter Binding with `[[vk::counter_binding]]`
+
+To allow for explicit control over counter bindings, a new attribute,
+`[[vk::counter_binding(binding)]]`, will be introduced. Internally, this will be
+a new attribute HLSLCounterBindingAttr that will be modeled after the
```
hekota, line 111:
```suggestion
a new attribute HLSLVkCounterBindingAttr that will be modeled after the

File: proposals/0023-typed-buffer-counters.md [hunk 5]
```diff
@@ -24,39 +24,230 @@ explicit that they are separate resources.
...
+
+### Array Handling
+
+For arrays of resources, the counter binding information is stored in the
+`HLSLCounterBindingAttr` of the array declaration itself. When Sema acts on the
+variable declaration, it will add an `HLSLCounterBindingAttr` with an implicit
+binding if the attribute does not already exist. When an array element is
+initialized with a call to a `__createHandle...` function during CodeGen, the
+appropriate `...With*Counter` version of the create function will be called for
+resources that have the `HLSLCounterBindingAttr`.
```
hekota, line 189:
```suggestion
For arrays of resources, the counter binding information is stored in the
`HLSLVkCounterBindingAttr` of the array declaration itself. When Sema acts on the
variable declaration, it will add an `HLSLVkCounterBindingAttr` with an implicit
binding if the attribute does not already exist. When an array element is
initialized with a call to a `__createHandle...` function during CodeGen, the
appropriate `...With*Counter` version of the create function will be called for
resources that have the `HLSLVkCounterBindingAttr`.
```

hekota (approved):
LGTM! I think the class for `[[vk::counter_binding()]]` attribute should be named `HLSLVkCounterBinding` to match the existing `HLSLVkBinding`.

proposals/0023-typed-buffer-counters.md
hekota: I think it would be better to have all of the implicit binding order_ids stored in one place - on the `HLSLResourceBindingAttr`. So it would store both implicit order ID for the resource and for the counter (if the resources has a counter). Then if there is a `HLSLVkCounterBindingAttr` that clearly means explicit binding for the counter, which is the same behavior we have for `HLSLVkBinding`.

PR #349, opened 2025-09-25
Mapping of GetDimensions overload to builtin functions and LLVM intrinsics

### LINKED ISSUE #346: Figure out mapping of GetDimensions overloads to builtin functions and LLVM intrinsic
This task is to figure out how to map the many `GetDimensions` overloads to Clang builtin functions and LLVM intrinsics for both buffer resources and for textures. The priority is buffer resources, but we should have a general idea of how this is going to work for textures as well. 

There are 54 cases of `GetDimensions` we need to support (listed [here](https://github.com/llvm/llvm-project/issues/112982#issuecomment-2492908100)). Based on the different output arguments we should be able to reduce this to a much lower number of builtin function that will get translated to one or more LLVM intrinsics.

For DXIL the `GetDimensions` calls end up being translated to a single DXIL op `getDimensions`, but it is TBD whether we also want to have a single `llvm.dx.resource.getdimensions` intrinsic that would directly map to the DXIL op, or whether there is a benefit of having a more strongly typed set of intrinsics until we lower to the DXIL op.

SPIR-V will most likely need multiple LLVM intrinsics because the `GetDimensions` calls get translated to different op codes based on the resource type and overload used, such as `OpArrayLength`, `OpImageQuerySize`, `OpImageQuerySizeLod` or `OpImageQueryLevel`.

hekota:
This proposal describes how to map the many `GetDimensions` overloads for buffer and texture resources to Clang built-in functions and LLVM intrinsics. There are 54 distinct `GetDimensions` cases to support and we can significantly reduce the number of required built-in functions by grouping overloads with similar argument patterns. Each built-in function can then be translated to one or more LLVM intrinsics depending on target platform.

Closes #346

### DIFF
--- a/proposals/NNNN-resources-get-dimensions.md
+++ b/proposals/NNNN-resources-get-dimensions.md
@@ -94,9 +94,9 @@ or `float`.
 
 | Resource class | Overloads   | SPIR-V op |
 |----------------|-------------|-----------|
-|Buffer<br/>RWBuffer|GetDimensions(out uint width)|OpImageQuerySize|
-|ByteAddressBuffer<br/>RWByteAddressBuffer|GetDimensions(out uint width)|OpArrayLength|
-|StructuredBuffer<br/>RWStructuredBuffer<br/>AppendStructuredBuffer<br/>ConsumeStructuredBuffer|GetDimensions(out uint count, out uint stride)|OpArrayLength|
+|Buffer<br/>RWBuffer|GetDimensions(out uint dim)|OpImageQuerySize|
+|ByteAddressBuffer<br/>RWByteAddressBuffer|GetDimensions(out uint dim)|OpArrayLength|
+|StructuredBuffer<br/>RWStructuredBuffer<br/>AppendStructuredBuffer<br/>ConsumeStructuredBuffer|GetDimensions(out uint numStructs, out uint stride)|OpArrayLength|
 
 Built-in function for overloads that just have a single `width` argument will look like
 this:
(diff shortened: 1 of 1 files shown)


proposals/NNNN-resources-get-dimensions.md
```diff
@@ -0,0 +1,217 @@
...
+
+### Clang builtin functions
+
+There are 54 `GetDimensions` member function overloads across all resource
+classes. We need to add a number of builtin functions to Clang that will be used
+to implement these. It will most likely amount to one builtin function per
+unique argument list, altough the same builtin function could be used for
+`GetDimensions` overloads that different only in `uint` vs. `float` argument
+types that are otherwise doing the same thing. This should greatly reduce the
+number of needed builtins function.
```
Icohedron, line 34: I rewrote this paragraph to read better in my head. Feel free to use it!
```suggestion
There are 54 `GetDimensions` member function overloads across all resource
classes. We need to add a number of built-in functions to Clang to implement 
these. This will most likely amount to one built-in function per unique argument
list. However, `GetDimensions` overloads that differ only by the types of their
arguments (namely, `uint` vs. `float`) can be consolidated under the same built-in
function, thereby greatly reducing the number of needed built-in functions.
```

Otherwise, you just misspelt the word "altough".

Also note: you used "built-in" on line 16 but "builtin" everywhere else. You could find and replace "builtin" with "built-in" to be more consistent.

proposals/NNNN-resources-get-dimensions.md
```diff
@@ -0,0 +1,217 @@
...
+`%dx.types.Dimensions = type { i32, i32, i32, i32 }`.
+ 
+Values in this struct correspond to the requested resource dimension values. The
+exact mapping depends on the resource type and is documented
+[here](https://github.com/microsoft/DirectXShaderCompiler/blob/main/docs/DXIL.rst#getdimensions).
+
+The LLVM intrinsic that maps to the DXIL op could look something like this:
+
+```
+{i32, i32, i32, i32} llvm.dx.resource.getdimensions(target("dx.*",..) resource_handle, i32 mip_level)
```
Icohedron, line 55: Missing an `@` in the intrinsic definition?
```suggestion
{i32, i32, i32, i32} @llvm.dx.resource.getdimensions(target("dx.*",..) resource_handle, i32 mip_level)
```

Icohedron (approved):
Looks good. I just had a few nits.

hekota:
> This LGTM. The only other thing to consider is if we should use intrinsic function for a builtin function like https://github.com/llvm/llvm-project/blob/main/llvm/docs/SPIRVUsage.rst#instructions-as-function-calls.
> 
> If you intend to have the intrinsic to be just a single SPIR-V instruction, then the builtin could be an option.

I was imagining there would be 5 LLVM intrinsics for SPIR-V that would map to the 5 SPIR-V operations used to implement `GetDimensions`, but ultimately I'll defer that decision to you.

Issue #175, opened 2025-02-18
Document DXC support of resources inside user-defined structs


hekota:
Resources and arrays of resources can be declared inside user defined structs in constant buffers.

https://godbolt.org/z/afTe1eTzx

It looks like DXC only allows access to resources arrays using a constant expression and not dynamically.
Document supported cases and add test(s) to offload-test-suite.

Related to #127.

PR #369, opened 2025-12-04
Document DXC handling of resources in global non-static structs

hekota:
This initial proposal documents the way DXC handles resources and resource arrays inside user-defined structs and classes which are declared as non-static at the shader global scope (the default `cbuffer` context `$Globals`).

It covers:
- single resource in a struct
- resource array in a struct
- arrays of structs that contain resources
- multiple resource kinds within the same struct
- resources in base classes
- static vs. dynamic indexing
- implicit binding of resources and resource arrays in structs
- error cases

Part 1 of #175

### DIFF
--- a/proposals/NNNN-resources-in-structs.md
+++ b/proposals/NNNN-resources-in-structs.md
@@ -20,8 +20,13 @@ because multiple instances of the same struct could have conflicting bindings.
 
 ## Current Behavior in DXC
 
-This section documents how DXC handles resources declared as members of structs
-and theirs bindings.
+This section documents how DXC handles resources that are members of non-static
+structs instances declared at the shader global scope, i.e. in the default
+`cbuffer` context `$Globals`, and how it handles their register binding.
+
+Additional information about how structs with resources can be used as local
+variables, function parameters or static global decls, and their initialization,
+assignment, and access behaviors will be added later on.
 
 ### Struct with a Single Resource
 
@@ -381,9 +386,32 @@ https://godbolt.org/z/Mo8Paoq7G
 <source>:5:7: error: register space cannot be specified on global constants.
 ```
 
+### Summary
+
+- DXC supports resources as members of structs, generating global resources
+  named after both the struct instance and member.
+
+- Register binding for these resources is determined by explicit `register`
+  annotations on the struct instance; if there are no binding annotations,
+  resources are implicitly bound in the order they are first referenced, which
+  can lead to unpredictable assignments.
+
+- Resource arrays within structs are treated as collections of individual
+  resources, with only accessed elements consuming registers. This differs from
+  global arrays which reserve the entire range.
+
+- Dynamic (non-constant) indexing into resource arrays or arrays of structs with
+  resources is not supported and results in compilation errors.
+
+- Structs may contain multiple resource types, including those inherited or
+  nested, with binding rules applied per resource type.
+
+- DXC validates register ranges for global resources but does not check for
+  overflow in struct members, which may result in silently overflow.
+
 ## Motivation
 
-We need to support this in Clang.
+We need to support resources in structs in Clang.
 
 ## Proposed solution
 


proposals/NNNN-resources-in-structs.md
```diff
@@ -0,0 +1,394 @@
+---
+title: "[NNNN] - Resources in Structures"
+params:
+  authors:
+    - github_username: Helena Kotas
```
spall: should this say hekota instead and your name above?

proposals/NNNN-resources-in-structs.md
```diff
@@ -0,0 +1,394 @@
...
+; ------------------------------ ---------- ------- ----------- ------- -------------- ------
+; d.2.Buf                               UAV     f32         buf      U0             u7     1
+; d.5.Buf                               UAV     f32         buf      U1            u10     1
+; e.7.Buf                               UAV     f32         buf      U2             u0     1
+; e.1.Buf                               UAV     f32         buf      U3             u1     1
+```
+
+### Dynamic Indexing of Struct Arrays with Resources
+
+If a resource is a members of a struct that is in an array, DXC does not support
```
spall: 'is a member of'

proposals/NNNN-resources-in-structs.md
```diff
@@ -0,0 +1,394 @@
...
+; g.F.SrvBuf1                       texture  struct         r/o      T0             t3     1
+; g.SrvBuf2                         texture  struct         r/o      T1             t4     1
+; g.UavBufs2.3                          UAV     f32         buf      U0             u9     1
+; g.UavBufs2.1                          UAV     f32         buf      U1             u7     1
+; g.e.UavBuf1                           UAV     f32         buf      U2             u5     1
+```
+
+#### Example 7
+
+If we remove the `register` annotations are the `G` instance in [Example
```
spall: are -> on

proposals/NNNN-resources-in-structs.md
```diff
@@ -0,0 +1,394 @@
...
+[numthreads(4,1,1)]
+void main() {
+  k.Buf[0] = 0;
+}
+```
+https://godbolt.org/z/Mo8Paoq7G
+```
+<source>:5:7: error: register space cannot be specified on global constants.
+```
+
```
spall: Does this apply to RWStructuredBuffer as well? Is this a possible case and should it be considered?  Can a resource in a struct contain another resource? 
```c++
struct A {
  RWBuffer<float> Buf;
};

struct B {
  RWStructuredBuffer<A> Buf2;
};

B b1 : register(u0);
A a1 : register(u5);
A a2;

[numthreads(4,1,1)]
void main() {
  a1.Buf[0] = a2.Buf[0];
}
```

spall (commented):
mostly looks good to me wrt to the writing.  I'm less confident on if there are missing cases and if it is complete in that regard.

proposals/NNNN-resources-in-structs.md
hekota: No, this is not supported by DXC or Clang. The type in `RWStructuredBuffer` must not be intangible, and anything that contains a resource is considered intangible.

https://godbolt.org/z/j8dfKK7bc

spall (approved):
lgtm

proposals/NNNN-resources-in-structs.md
```diff
@@ -0,0 +1,394 @@
...
+    - hekota: Helena Kotas
+---
+
+## Introduction
+
+HLSL resources are runtime-bound data objects supplied to shader programs as
+input, output, or both. While they are typically declared as global variables of
+specific resource types at the shader's global scope, resources can also be
+members of user-defined structures or classes. For brevity, this document refers
+to user-defined structures and classes simply as _structs_.
```
tex3d: Can we add a qualifier about this document being scoped to register binding behavior for non-static global decls of struct types containing resources?

There would be more to cover for the general concept of resources in structs for HLSL otherwise, such as local, parameter, and static global decls, initializer, assignment, and access behaviors, and so on.

PR #376, opened 2026-01-21
[0041] New proposal for `testing-maximal-reconvergence`

luciechoi:
This proposal suggests a different approach for having comprehensive test coverage for maximal reconvergence feature in the Clang compiler.

### DIFF
--- a/proposals/0041-testing-maximal-reconvergence.md
+++ b/proposals/0041-testing-maximal-reconvergence.md
@@ -1,12 +1,12 @@
 ---
-title: "0039 - Testing Maximal Reconvergence"
-- draft: true
+title: "0041 - Testing Maximal Reconvergence"
 params:
   authors:
     - luciechoi: Lucie Choi
   sponsors:
     - s-perron: Steven Perron
     - Keenuts: Nathan Gauër
+    - bogner: Justin Bogner
   status: Under Consideration
 ---
 
@@ -49,9 +49,27 @@ This is the place that needs extensive testing. In the example below, a compiler
 may reorder the code (e.g SimplifyCFG pass) so that statements are moved
 inside the branches, producing incorrect results.
 
-| Before Optmization | After Optimization |
-| --- | --- |
-| <pre><code>if (non_uniform_cond) {<br>   doA(); <br>   Out[...] = waveOperations();<br>} else {<br>   doB(); <br>   Out[...] = waveOperations(); <br>}<br></code></pre> |  <pre><code>if (non_uniform_cond) {<br>   doA(); <br>} else {<br>   doB(); <br>} <br> // Invalid transformation. <br> Out[...] = waveOperations(); </code></pre> |
+##### Before Optimization
+```cpp
+if (non_uniform_cond) {
+   doA(); 
+   Out[...] = waveOperations();
+} else {
+   doB(); 
+   Out[...] = waveOperations(); 
+}
+```
+
+##### After Optimization
+```cpp
+if (non_uniform_cond) {
+   doA(); 
+} else {
+   doB(); 
+} 
+// Invalid transformation. 
+Out[...] = waveOperations(); 
+```
 
 This kind of optimization should be prevented. In DXC, spirv-opt is used to
 optimize when targeting Vulkan. It is aware of HLSL's
@@ -113,13 +131,13 @@ This is an [example](https://github.com/llvm/offload-test-suite/pull/685) of the
 test generator and the generated
 [tests](https://github.com/llvm/offload-test-suite/pull/620).
 
-#### 1. Random shaders
+#### 1. (Pseudo) Random shaders
 
 Random control flow will be produced by a fixed-seed RNG and hard-coded
 probabilities. For example, they will determine whether the next instruction
-will be a loop, if, switch, etc, and with what conditions. For the random number
-generator, we will port one from
-[llvm::RandomNumberGenerator](https://github.com/llvm/llvm-project/blob/8e335d533682b46289058958456c521df0c8fe32/llvm/include/llvm/Support/RandomNumberGenerator.h#L33C1-L38C42),
+will be a loop, if, switch, etc, and with what conditions. For the pseudo-random
+number generator, we will port one from
+[llvm::RandomNumberGenerator](https://github.com/llvm/llvm-project/blob/8e335d533682b46289058958456c521df0c8fe32/llvm/include/llvm/Support/RandomNumberGenerator.h#L33C1-L38C42), 
 which is deterministic and operating system independent.
 
 These random instructions are represented in a custom intermediate
@@ -178,9 +196,10 @@ pipeline.
 
 #### CMake Target 
 
-We will implement a cmake target `check-hlsl-{platform}-reconvergence`, similar
-to the existing targets. Running this will generate the physical tests and run
-them.
+We will implement cmake targets `check-hlsl-{platform}-reconvergence`, similar
+to the existing targets. Running this command will generate the physical tests
+and execute them. We will separate cmake targets for writing the tests so that
+the tests will not be regenerated every time the tests are run.
 
 #### Github Workflow
 
@@ -193,7 +212,7 @@ New steps will be added to the existing workflow at the end:
 - **Run Reconvergence Tests**
 
 This way, the execution of existing HLSL tests and the reconvergence tests are
-separated.
+separated, and it will be easiser to report and investigate issues.
 
 We don't plan to store the physical test files in the repo. Developers can still
 save, run, and inspect the tests locally by running the target in their machine.
@@ -205,7 +224,28 @@ We will segment the output buffer and verification into multiple buffers and
 checks or implment an environment variable to filter out some logs.
 
 If any test fails, it will fail the workflow, so it's noticeable in the badge.
-`XFail` instructions will be added appropriately to suppress failures.
+
+`XFail` instructions will be added appropriately to suppress failures. Since it
+is undesirable to change the code of the C++ random test generator every time
+failure happens, the test generator may read a structured text file that
+contains a list of failing tests and their environments. This way, only this
+single file will be updated upon any changes in the compilers, and the algorithm
+for generating the tests remains intact.
+
+*reconvergence-failing-tests.txt*
+```yaml
+reconvergence-test_2_16_7_13_3.test
+# Some comment
+# XFAIL: Clang && Vulkan
+
+# Some comment
+# XFAIL: ...
+
+reconvergence-test_5_32_7_13_1.test
+# Some comment
+# XFAIL: ...
+
+```
 
 ### Latency
 


proposals/0039-testing-maximal-reconvergence.md
```diff
@@ -0,0 +1,94 @@
...
+  status: Under Consideration
+---
+
+* PRs: [Testing in offload-test-suite (Draft)](https://github.com/llvm/offload-test-suite/pull/620)
+* Issues:
+  [Implementation in Clang](https://github.com/llvm/llvm-project/issues/136930)
+
+## Introduction
+
+This proposal seeks to add comprehensive conformance tests for the maximal reconvergence feature in HLSL compilers (DXC and Clang).
```
s-perron: The term "maximal reconvergence" is not an HLSL term. It is a SPIR-V term. We should say something more like:


```suggestion
This proposal seeks to add comprehensive conformance tests that HLSL compilers (DXC and Clang) do not violate the optimization restrictions in section 1.6.3 of the HLSL specification.
```

proposals/0039-testing-maximal-reconvergence.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Motivation
+
+Graphics compilers often perform aggressive optimizations that can unexpectedly alter the convergence behavior of threads within a wave. This is a critical issue for shaders containing operations dependent on control flow, such as wave intrinsics, as invalid transformations can lead to wrong or indeterminate results.
+
+Maximal reconvergence is a set of compiler guarantees designed to prevent these unintended changes, ensuring that divergent threads reconverge at expected merge points and that wave operations execute in lockstep where intended.
+
+This feature is fully implemented in DirectXCompiler and LLVM, and testing for correct convergence behavior is critical for reliability. Currently, only a few unit tests exist. We need to extend this coverage to include complex and highly divergent cases.
```
s-perron: I would drop the first sentence. The state of the implementation is irrelevant to the motivation.

proposals/0039-testing-maximal-reconvergence.md
```diff
@@ -0,0 +1,94 @@
...
+
+### Test Generation and Simulation
+The shaders will be generated when the test pipeline starts. Since each GPU has different subgroup sizes, each machine will have a version for every power-of-2 wave size between 4 and 32 (e.g., 4, 8, 16, 32). The tests that do not match the subgroup size of the running GPU will be skipped (e.g. through `# UNSUPPORTED: !SubgroupSizeX` directive).
+
+### Translation
+
+Logic from [Vulkan CTS GLSL generation](https://github.com/KhronosGroup/VK-GL-CTS/blob/main/external/vulkancts/modules/vulkan/reconvergence/vktReconvergenceTests.cpp) will be ported to produce HLSL. This includes translating intrinsics such as `subgroupElect()` to `WaveIsFirstLane()` and `subgroupBallot()` to `WaveActiveBallot()`, etc.
+
+### Execution Pipeline
+Only the code for the **random shader generator** will reside in the offload-test-suite repository.
```
s-perron:
```suggestion
Only the code for the **random test generator** will reside in the offload-test-suite repository.
```

proposals/0039-testing-maximal-reconvergence.md
```diff
@@ -0,0 +1,94 @@
...
+
+### Latency
+
+The entire Vulkan-CTS test (~1500 shaders) takes ~10 seconds to complete, so the test generation + execution time should be similar and should not significantly affect the current pipeline duration. We may also choose to start with smaller iterations (~100 shaders).
+
+### Reporting
+
+Results of the reconvergence tests will be aggregated. Failing shaders will be logged separately or made available via YAML artifacts to avoid diluting logs with excessive data.
```
s-perron: does this mean that the tests that pass will be deleted? where will the failing test be saved?

proposals/0039-testing-maximal-reconvergence.md
```diff
@@ -0,0 +1,94 @@
...
+## Detailed design
+
+### Test Generation and Simulation
+The shaders will be generated when the test pipeline starts. Since each GPU has different subgroup sizes, each machine will have a version for every power-of-2 wave size between 4 and 32 (e.g., 4, 8, 16, 32). The tests that do not match the subgroup size of the running GPU will be skipped (e.g. through `# UNSUPPORTED: !SubgroupSizeX` directive).
+
+### Translation
+
+Logic from [Vulkan CTS GLSL generation](https://github.com/KhronosGroup/VK-GL-CTS/blob/main/external/vulkancts/modules/vulkan/reconvergence/vktReconvergenceTests.cpp) will be ported to produce HLSL. This includes translating intrinsics such as `subgroupElect()` to `WaveIsFirstLane()` and `subgroupBallot()` to `WaveActiveBallot()`, etc.
+
+### Execution Pipeline
```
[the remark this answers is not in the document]
luciechoi: Thanks for pointing out. I've added an example workflow. 

Here is the new draft PR with workflow changes. https://github.com/llvm/offload-test-suite/pull/685

proposals/0039-testing-maximal-reconvergence.md
```diff
@@ -0,0 +1,229 @@
...
+results. Historically, there is only an [informal
+definition](https://github.com/microsoft/directxshadercompiler/wiki/wave-intrinsics#operation)
+of which threads should be active at any point in execution of the shader:
+"<i>implementations must enforce that the number of active lanes exactly
+corresponds to the programmer’s view of flow control</i>". 
+
+When lowering HLSL to SPIR-V, we must make sure the output matches this
+expectation. To do so, there are 2 areas that need to be looked at:
+
+#### 1. Adding `SPV_KHR_maximal_reconvergence` extension and `MaximallyReconvergesKHR` capability.
```
s-perron: Mention that this is Vulkan specific.

proposals/0039-testing-maximal-reconvergence.md
```diff
@@ -0,0 +1,229 @@
...
+
+#### 2. Ensuring the frontend compilers themselves do not alter the state during optimizations.
+
+This is the place that needs extensive testing. In the below example, a compiler
+may reorder the code (e.g loop unswitch pass) so that statements are moved
+inside the branches, producing incorrect results.
+
+| Before Optmization | After Optimization |
+| --- | --- |
+| <pre><code>if (non_uniform_cond) {<br>   doA(); <br>} else {<br>   doB(); <br>}<br>// Expected converged. <br>Out[...] = waveOperations(); </code></pre>| <pre><code>if (non_uniform_cond) {<br>   doA(); <br>   // Invalid transformation <br>   Out[...] = waveOperations();<br>} else {<br>   doB(); <br>   // Invalid transformation <br>   Out[...] = waveOperations(); <br>}<br></code></pre> |
```
s-perron: This is an example of an invalid transformation. However is that the transformation that is actually done? Usually it goes the other way. This is called Tail merging and it is part of SimplifyCFG.

The `e.g.` above makes it seem like you were going to give an example of loop unswitching.

proposals/0039-testing-maximal-reconvergence.md
```diff
@@ -0,0 +1,229 @@
...
+## Proposed solution
+
+We propose implementing a comprehensive test suite in the offload-test-suite
+repository that mirrors the logic of the Vulkan Conformance Testing Suite
+(Vulkan-CTS). This involves generating shaders with random control flows (mixes
+of if/switch statements, loops, and nesting) and verifying the results.
+
+### Shader Generation
+
+Approximately _N_ number of  HLSL shaders will be generated. These shaders use
```
s-perron: Are you going to replace N with a real number? Otherwise, I don't think you need to mention the number of shader. Focus on how they will be generated.

proposals/0039-testing-maximal-reconvergence.md
```diff
@@ -0,0 +1,229 @@
...
+We will generate a set of yaml test files for the offload-test-suite. For each
+shader and subgroup size (4, 8, 16, 32), a test file will be generated that
+executes the shader and verifies that the results match the CPU simulation.
+
+## Detailed design
+
+[This](https://github.com/llvm/offload-test-suite/pull/685) is an example of the
+proposed design.
+
+### Test Generation and Simulation
```
s-perron: You might want to fix up the order here. The test generation include the simulation. Here is an example of how I would lay it out. Fill in the text. I just have a few points on what should be included.

### Test Generation

#### Random shaders

// Move the "Transation" section here.
Explicitly call out the intermediate form for the shader, and describe it a bit.

#### Expected results

// Mention how you will pick the size of the buffer, which is determine by a "dry run" in vulkan.
// Mention that you will get the expected results for each wave size doing a CPU simulation
Mention that the CPU simulation will be done on the intermediate form and not the final SPIR-V shader.

#### Final test file

// Explain how the find test case will be generated. This can be short as it is simply taking the info from the other step. Say where the new tests will be stored.

proposals/0039-testing-maximal-reconvergence.md
```diff
@@ -0,0 +1,229 @@
...
+- Dump GPU Info
+- Run HLSL Tests
+- **Generate Random Reconvergence Tests**
+- **Run Reconvergence Tests**
+
+This way, the execution of existing HLSL tests and the reconvergence tests are
+separated.
+
+```yaml
+# .github/workflows/build-and-test-callable.yaml
```
s-perron: A github workflow? Will I be able to run them locally? I think we should have a cmake target to be able to run the tests locally.

proposals/0039-testing-maximal-reconvergence.md
```diff
@@ -0,0 +1,229 @@
...
+```
+
+We don't plan to store the physical test files in the repo. Developers can still
+run the tests locally by running the test generator to output the tests in their
+machine.
+
+### Reporting
+
+We may implment an environment variable `OFFLOADTEST_SUPPRESS_DIFF` to filter
+out some logs, since for example, diffs will be massive for a failing test.
```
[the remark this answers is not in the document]
s-perron: I looked at the tests closer. Not all tests use filecheck. The ones you created did not.

Something else we can do it so make use arrays of resources. See the "array-global.test" test. Instead of having a single output buffer, the output buffer is an array, where each thread writes to its own buffer.

RWStructuredBuffer<uint4> OutputB : register(u1);

becomes


RWStructuredBuffer<uint4> OutputB[NUM_THREADS] : register(u1);

Then 

OutputB[(outLoc++)*invocationStride + gIndex].x = 0x10002;

becomes

OutputB[gIndex][(outLoc++)].x = 0x10002;

Then you can have a CHECK line for each thread. The size of each line will be much smaller. I one test it would be only 100 values per line, and it if failed it would give a diff pointing to the incorrect value. 

We could try to figure out a way to  modify the indices. Instead of using `outLoc++`, we can use a formula with some constants so that we easily know which line and which iterations of any containing loop was suppose to have written to that location with the given loop iterations. This last part might be a bit forward looking. Not needed now. These are the types of things we can do to make the test easier to debug.

proposals/0039-testing-maximal-reconvergence.md
```diff
@@ -0,0 +1,254 @@
+---
+title: "0039 - Testing Maximal Reconvergence"
```
bogner: Careful, a proposal 39 was committed while this was in review. Looks like you'll need to update to 41

proposals/0039-testing-maximal-reconvergence.md
```diff
@@ -0,0 +1,254 @@
...
+
+#### 2. Ensuring the frontend compilers themselves do not alter the state during optimizations.
+
+This is the place that needs extensive testing. In the example below, a compiler
+may reorder the code (e.g SimplifyCFG pass) so that statements are moved
+inside the branches, producing incorrect results.
+
+| Before Optmization | After Optimization |
+| --- | --- |
+| <pre><code>if (non_uniform_cond) {<br>   doA(); <br>   Out[...] = waveOperations();<br>} else {<br>   doB(); <br>   Out[...] = waveOperations(); <br>}<br></code></pre> |  <pre><code>if (non_uniform_cond) {<br>   doA(); <br>} else {<br>   doB(); <br>} <br> // Invalid transformation. <br> Out[...] = waveOperations(); </code></pre> |
```
bogner: This is really hard to read as text, and previewing it on github I think it's ignoring the `<br>`s, so it's hard to read that way too. Probably better to just do a before and after codeblock consecutively instead of trying to render these side by side.

proposals/0039-testing-maximal-reconvergence.md
```diff
@@ -0,0 +1,254 @@
...
+New steps will be added to the existing workflow at the end:
+
+- Build DXC
+- Build LLVM
+- Dump GPU Info
+- Run HLSL Tests
+- **Run Reconvergence Tests**
+
+This way, the execution of existing HLSL tests and the reconvergence tests are
+separated.
```
bogner: I assume this is just for ease of reporting / seeing what needs to be investigated? Might be worth explicitly stating why separating the tests matters.

bogner (approved):
This is good enough shape to go in, and any further work on the proposal can go in tree. Let's get it merged.

It would be good to get issues filed for followups, including ones about going into more detail about the XFAILs and any tooling we may or may not add for debugging failures.

proposals/0039-testing-maximal-reconvergence.md
luciechoi: Separated.

luciechoi:
> This is good enough shape to go in, and any further work on the proposal can go in tree. Let's get it merged.
> 
> It would be good to get issues filed for followups, including ones about going into more detail about the XFAILs and any tooling we may or may not add for debugging failures.

Thank you for your review!

PR #384, opened 2026-02-19
[0037] Update texture design

s-perron:
This updates the design for the textures based on issue found during
implementation:

1. There is no good default value for the `clamp`, so we add a `*clamp`
   intrinsic for the various Sampling intrinsics for the cases with a
   clamp.
2. We need to use some template magic to get the correct return type for
   the Gather instructions. This is documented in is own section.

### DIFF
--- a/proposals/0037-texture-and-sampler-types.md
+++ b/proposals/0037-texture-and-sampler-types.md
@@ -239,6 +239,41 @@ public:
 }
 ```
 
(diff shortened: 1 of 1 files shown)


hekota (commented):
Most of the lines changed are just formatting, only about 10% is new content. Could you please split this PR into two - one for the "NFC" formatting change and the other for the design doc update?

Keenuts (approved):

PR #398, opened 2026-03-17
[0044] Vertex and Pixel Shader Workstreams

inbelic:
Adds a proposal to outline the workstreams required to allow for general pixel/vertex shaders and texture support

### DIFF
--- a/proposals/0044-vertex-pixel-shader-directx-workstreams.md
+++ b/proposals/0044-vertex-pixel-shader-directx-workstreams.md
@@ -14,7 +14,7 @@ Clang. This includes semantic and signature support, texture/sampler types and
 their methods, and the extensions to the offload test suite required to
 validate them.
 
-The work is organized into six workstreams:
+The work is organized into four workstreams:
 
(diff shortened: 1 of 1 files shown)


proposals/NNNN-hlsl-texture-support-directx.md
```diff
@@ -0,0 +1,308 @@
...
+2. Textured Cube (Core Texture/Sampler Operations): Defines the Texture/Sampler
+types and their methods, defines their lowering to DXIL and updates to allow
+for testing in the offload test suite.
+
+## Motivation
+
+Texture resources and samplers are required for graphics programming in HLSL.
+Clang does not currently support either when targeting DirectX and there was
+not a previous document to concretely outline what features are missing and
+need to be implemented.
```
bogner: Not that the motivation for this is terribly surprising, but this talks about textures and samplers but doesn't mention semantics and signatures at all. Might be better to simplify to say that we need these four things addressed in order to reasonably say we support pixel and vertex shaders.

PR #399, opened 2026-03-17
Update HLSL Intrinsic TableGen proposal with feedback, and set status to Approved

Icohedron:
This PR updates the HLSL intrinsic TableGen proposal [0043] with some feedback from today's design meeting. 

- Made the type of the `Body` field `code` instead of `string`. There
is no difference aside from `code` being more clear that `Body` is
intended for code and not some arbitrary string. 
(See https://discourse.llvm.org/t/tablegen-code-type-to-be-eliminated/56996/8)

- Rewrote descriptions of the `Body` field to make it more clear that
it is intended for single-line implementations of inline functions.
Multi-line implementations should instead live in a helper function
called by `DetailFunc`.

- Added a section with an example of the `Doc` field being used to emit
a Doxygen comment with the first overload.

There apparently is not a `boolean` type for TableGen, only `bit` types with accept 0 or 1. 
So `VaryingScalar` remains 0 or 1.

### DIFF
--- a/proposals/0043-hlsl-intrinsic-tablegen.md
+++ b/proposals/0043-hlsl-intrinsic-tablegen.md
@@ -1,9 +1,9 @@
 ---
-title: "[NNNN] - HLSL Intrinsic TableGen"
+title: "[0043] - HLSL Intrinsic TableGen"
 params:
   authors:
     - icohedron: Deric Cheung
-  status: Under Consideration
+  status: Approved
 ---
 
 ## Introduction


bob80905:
LGTM, I just wonder if an answer to any of the questions you posed here:
https://github.com/llvm/llvm-project/pull/186218#discussion_r2928193800
would change anything in our plans for 16 bit types / macros for such declarations.

Icohedron:
> LGTM, I just wonder if an answer to any of the questions you posed here: [llvm/llvm-project#186218 (comment)](https://github.com/llvm/llvm-project/pull/186218#discussion_r2928193800) would change anything in our plans for 16 bit types / macros for such declarations.

None of the HLSL intrinsics (at least so far in Clang) introduce 16-bit type support later than SM 6.2. But if there were such an HLSL intrinsic that didn't support 16-bit types until a later SM, its defs could be broken up into two. Something like the following:

```tablegen
def hlsl_foo : HLSLOneArgBuiltin<"foo", "__builtin_hlsl_foo"> {
  let VaryingTypes = Non16BitTypes;
  let Availability = SM6_x;
}

def hlsl_foo_16bit : HLSLOneArgBuiltin<"foo", "__builtin_hlsl_foo"> {
  let VaryingTypes = All16BitTypes;
  let Availability = SM6_y;
}
```

That way there are two sets of overloads generated for `foo` but with different Shader Model availability / requirements based on the varying element type.

PR #413, opened 2026-04-10
[0046] Add proposal for HLSL ConstantBuffer<T>

s-perron:
This proposal details the implementation of the `ConstantBuffer<T>`
resource type in Clang and LLVM for HLSL.

Unlike the legacy `cbuffer`, `ConstantBuffer<T>` behaves as a standard
type, supporting instantiation, arrays, function parameters, and
assignments. It acts as a drop-in replacement for `T` via an implicit
conversion operator to `const T &`. The proposal details the Sema member
lookup interception required and the corresponding CodeGen which targets
the appropriate constant address spaces via resource intrinsics.

### DIFF
--- a/proposals/0046-constantbuffer-t.md
+++ b/proposals/0046-constantbuffer-t.md
@@ -1,5 +1,5 @@
 ---
-title: "[NNNN] - HLSL ConstantBuffer<T> Implementation"
+title: "[0046] - HLSL ConstantBuffer<T> Implementation"
 params:
   authors:
     - s-perron: Steven Perron
--- a/proposals/NNNN-constantbuffer-t.md
+++ b/proposals/NNNN-constantbuffer-t.md
@@ -28,9 +28,12 @@ standard, and it is used.
 ## Proposed solution
 
 We propose implementing `ConstantBuffer<T>` as a built-in template class that
-provides an implicit conversion to `T &`. To ensure a seamless developer
-experience, `Sema` will be modified to automatically inject this conversion when
-accessing members of a `ConstantBuffer`.
+provides an implicit conversion to `const hlsl_constant T &`. To ensure a
+seamless developer experience, `Sema` will be modified to automatically inject
+this conversion when accessing members of a `ConstantBuffer`.
+
+Places with HLSL specific handling for aggregates will also be updated to
+convert the `ConstantBuffer<T>` to `T` when necessary.
 
 ### Frontend (Clang AST/Sema)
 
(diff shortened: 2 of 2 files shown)


s-perron:
@hekota @bogner @llvm-beanz Hey, here is a draft design for `ConstantBuffer<T>`. Let me know if you have other ideas or reason this will not work.

I still have to figure out the implicit conversion to `T`. @spall's work to remove the constructors puts that part in flux.

proposals/NNNN-constantbuffer-t.md
```diff
@@ -0,0 +1,348 @@
...
+## Introduction
+
+This proposal details the implementation of the `ConstantBuffer<T>` resource
+type in Clang and LLVM for HLSL. Unlike the legacy `cbuffer` keyword, where each
+member of the `cbuffer` becomes its own global variable, `ConstantBuffer<T>`
+behaves as a standard type, supporting instantiation, arrays, function
+parameters, and assignments. The `ConstantBuffer<T>` type acts more like other
+resource type than it does a `cbuffer`. The unique aspect of the
+`ConstantBuffer<T>` type is that it can be used as a drop in replacement for
+`T`, as if ConstantBuffer<T> inherited from `T`. However, it is not really
```
hekota:
```suggestion
`T`, as if `ConstantBuffer<T>` inherited from `T`. However, it is not really
```

proposals/NNNN-constantbuffer-t.md
```diff
@@ -0,0 +1,348 @@
...
+2.  **Implicit Conversion Operator:** Define an implicit conversion operator
+    `operator const hlsl_constant T &() const` within the `ConstantBuffer<T>`
+    template.
+3.  **Sema Member Lookup Interception:** Modify `Sema::LookupMemberExpr` (in
+    `SemaExprMember.cpp`) to detect member accesses on `ConstantBuffer<T>`. If
+    detected, Sema will inject a call to the implicit conversion operator,
+    effectively transforming `cb.field` into
+    `((const hlsl_constant T &)cb).field`.
+4.  **Sema Constraints:** Enforce that `T` must be a user-defined struct or
+    class, and reject primitive types, vectors, arrays, or matrices as `T`.
```
hekota: This could be done via [concept constraint](https://github.com/llvm/llvm-project/blob/2fe82e377f890d11e5915526f97ef7dce8a2d1d4/clang/lib/Sema/HLSLExternalSemaSource.cpp#L418) similar to the way the element type for Typed or Structured buffers is validated.

proposals/0046-constantbuffer-t.md
```diff
@@ -0,0 +1,348 @@
...
+1.  **Handle Types:** Emit instances of `ConstantBuffer<T>` using the
+    appropriate target types.
+2.  **Conversion Operator Implementation:** The conversion operator is
+    implemented using a new overload of `__builtin_hlsl_resource_getpointer`
+    that does not take an index.
+3.  **Intrinsic Redirection:** In CodeGen, this builtin lowers to a new target
+    intrinsic `llvm.[dx|spv].resource.getbasepointer` intrinsic. Unlike the
+    version used for buffers, this overload does not take an offset index, as
+    `ConstantBuffer` always represents a single instance of `T` at the start of
+    the buffer.
```
hekota, line 81: I wonder if we could use the same intrinsic and just leave the index be 0.

proposals/NNNN-constantbuffer-t.md
```diff
@@ -0,0 +1,348 @@
...
+    `-MemberExpr '<bound member function type>' .operator const hlsl_constant S &
+      `-ImplicitCastExpr 'const hlsl::ConstantBuffer<S>' lvalue <NoOp>
+        `-DeclRefExpr 'cb'
+```
+
+#### 2. Local Assignment
+
+TODO: This needs to be updated based on the implementation in
+https://github.com/llvm/llvm-project/pull/190089. That PR will disable implicit
+conversions.
```
hekota:
```suggestion
TODO: This needs to be updated based on the implementation in
https://github.com/llvm/llvm-project/pull/190089. That PR will disable implicit
copy constructors.
```

hekota (commented):
Looks good, I just wonder whether we need to add the new llvm intrinsic.

We are also going to need to decide if we are going to be supporting resources inside `T`. DXC seems to support it unless `ConstantBuffer<T>` is in an array: https://godbolt.org/z/11dTdhzh1

proposals/NNNN-constantbuffer-t.md
s-perron: That is what I have in the PoC that I have. I'll make the proposal more specific on that point.

proposals/0046-constantbuffer-t.md
s-perron: I tried that. For that to work semantically, for that to work for SPIR-V, we would have to wrap `T` in another struct when defining the `Vulkan.buffer` type because the 0 has a meaning to index into the type.

I can't do this when lowering the handle type to a SPIR-V target type. I can't distinguish between a cbuffer and a ConstantBuffer handle, but only the latter requires the types to be wrapped in a struct.

The other option is to try to do it in HLSLExternalSemaSource and friends. I attempted that, but the changes quickly caused a lot of special cases in code for the other resources.

I found the cleanest solution to be to avoid using the index.

s-perron:
> We are also going to need to decide if we are going to be supporting resources inside `T`. DXC seems to support it unless `ConstantBuffer<T>` is in an array: https://godbolt.org/z/11dTdhzh1

That is an awkward one. It never worked for SPIR-V: https://godbolt.org/z/Mqvq4cd6Y.

I would like to disallow it, but we can talk about this more. I'll update the proposal to disallows resources in ConstantBuffers, and then we can have a discussion in the appropriate meeting.

proposals/NNNN-constantbuffer-t.md
Keenuts: What about methods? Should we start enforcing those have a const `this`?
This somehow is accepted but should IMO not: https://godbolt.org/z/avE8zTKb7

s-perron: As I wrote the implementation, your example would be rejected because the member expression would add a cast to `const hlsl_constant& T`. Since the call to foo is not `const` it will create an error.

However that is a good example that I need to add. If DXC allowed call to non-const member function, should we allow them too? At what point do we issue an error? Note that your example fails when generating DXIL. I'm not sure why it works for spir-v. We probably have a bad optimization removing the invalid store because the spir-v is invalid.

Keenuts: I think we had a couple issues open in DXC for this weirdness,IMO the fact dxc allows this is an oversight, and we should not support this. Maybe something we can specify in hlsl 202x

proposals/0046-constantbuffer-t.md
hekota: That makes sense. It just was not clear to me why we need a new intrinsic. Could you please add a short note to the proposal about this?

hekota:
> I would like to disallow it, but we can talk about this more. I'll update the proposal to disallows resources in ConstantBuffers, and then we can have a discussion in the appropriate meeting.

I would also be in favor of disallowing resources in `ConstantBuffer<T>`, and I believe @tex3d mentioned that some time ago as well.

proposals/NNNN-constantbuffer-t.md
s-perron: I left this as a open question. We should get more people to chime in on. With a good compiler error it might not be too hard for devs to add `const` where needed if we do require const.

tex3d:
> > I would like to disallow it, but we can talk about this more. I'll update the proposal to disallows resources in ConstantBuffers, and then we can have a discussion in the appropriate meeting.
> 
> I would also be in favor of disallowing resources in `ConstantBuffer<T>`, and I believe @tex3d mentioned that some time ago as well.

You definitely weren't ever supposed to be allowed to put a resource in `T` for `ConstantBuffer<T>`. I don't know how that was allowed in DXC for a non-array decl, but it's definitely a bug.

proposals/NNNN-constantbuffer-t.md
```diff
@@ -0,0 +1,412 @@
+---
+title: "[NNNN] - HLSL ConstantBuffer<T> Implementation"
+params:
+  authors: Steven Perron
+    - github_username: s-perron
+  status: Under Consideration
+  sponsors: Steven Perron
+    - github_username: s-perron
```
bogner: The format for authors/sponsors in this header is `- <username>: <human readable name>`

```suggestion
params:
  authors:
    - s-perron: Steven Perron
  status: Under Consideration
  sponsors:
    - s-perron: Steven Perron
```

proposals/0046-constantbuffer-t.md
```diff
@@ -0,0 +1,412 @@
...
```
bogner, line 81: I wonder if in the future we should move away from `getpointer` entirely and use `getbasepointer` in combination with the structured/logical GEP. We can consider that at some later time though.

bogner (approved):
This all seems fairly reasonable to me. We'll still need to work out the constructor aspects and weigh it against the type-alias alternative proposal, but I think we can give it a number and get it merged as under consideration.

The main thing I worry about in this design is the "overloaded dot operator" aspect. The implicit conversion we're adding to Sema can't be represented in either HLSL or C++, so we're committing to ConstantBuffer<T> always having special handling as a type here. This differs from what we've done with the other resources, where at least in theory they could be implemented entirely in the language proper at some point in the future.

proposals/0046-constantbuffer-t.md
s-perron: I was wondering the same thing.

s-perron:
> The main thing I worry about in this design is the "overloaded dot operator" aspect. The implicit conversion we're adding to Sema can't be represented in either HLSL or C++, so we're committing to ConstantBuffer always having special handling as a type here. 

While looking at this, I had wished `ConstantBuffer<T>` was defined to use `operator->` to access the members of `T`. Then I could have just overloaded that. I don't know if we would consider this for 202y.

Issue #180, opened 2025-02-19
Survey the set of cases and edge cases we need to handle for locally scoped resource objects


bogner:
There are various things that can be done with a resource variables - they can be passed to functions in parameters, defined in local arrays, assigned, etc.

```c++
RWBuffer<float> Buf;

[numthreads(1,1,1)]
void main() {
    RWBuffer<float> T = Buf;
    T[0] = 0;
}
```

We need to come up with a set of test cases and acceptance criteria to say we handle these correctly in clang.

bob80905:
This should be closed with the merge of #1090
The offload test suite tests are done, but there are still llvm tests to be added, which will be tracked here: https://github.com/llvm/llvm-project/issues/214537
And there are doc changes to be made, which will be tracked here:
https://github.com/llvm/wg-hlsl/issues/458

PR #414, opened 2026-04-14
[NNNN] Document local resource behavior

bob80905:
This PR lays out a plan for test-driven documentation of local resource behavior.
Addresses https://github.com/llvm/wg-hlsl/issues/180

Assisted by: Github Copilot


proposals/NNNN-local-resources-behavior.md
```diff
@@ -0,0 +1,341 @@
...
+| Copy between locals (`a = g; b = a`) | Clean | Clean |
+| Self-assignment (`buf = buf`) | Clean | Clean |
+| Return initialized local | Clean | Clean |
+| Return uninitialized local | Clean | Clean |
+| Parenthesized ternary expression init | Clean | Clean |
+| Aggregate init of struct with resource | Clean | Clean |
+| Two resources in a single declaration | Clean | Clean |
+| Use of uninitialized local resource | Clean | Clean (sema), **Error (codegen)**: DXIL Op Lowering assertion on uninitialized resource Store |
+| Default-init store (unbound handle) | Error (codegen): *"local resource not guaranteed to map to unique global resource"* | Clean (sema), **Error (codegen)**: DXIL Op Lowering assertion on unbound handle |
+| Conditional init (`if(cond) buf = g;`) | Error (codegen): *"local resource not guaranteed to map to unique global resource"* | **Clean** |
```
inbelic: nit: Some of the Cleans are randomly bold. Or is there a reason these particular ones are?

proposals/NNNN-local-resources-behavior.md
```diff
@@ -0,0 +1,341 @@
...
+
+The observed behavior for local resource variables is documented below,
+organized by category. Each pattern includes the observed behavior for
+both DXC and Clang.
+
+### Basic Local Resource Operations
+
+| Pattern | DXC | Clang |
+|---------|-----|-------|
+| Alias from global (`buf = gBuf0`) | Clean | Clean |
```
inbelic: It is never defined what 'Clean' means. Does it mean compiled without errors/warnings? If something is compiled without errors does that actually means it is the correct behaviour?

proposals/NNNN-local-resources-behavior.md
```diff
@@ -0,0 +1,341 @@
...
+| Assign wrong type | Error: type mismatch | Error: *"no viable overloaded '='"* |
+| `const` reassignment | Error: *"cannot assign to const"* | Error: *"cannot assign to variable with const-qualified type"* |
+| `volatile` resource method call | **Clean** (silently accepts `volatile`) | Error: *"no matching member function"* |
+| `static const` resource method call | **ICE** | Error: `Load`/`Store` not `const`-qualified |
+
+### Invalid Declarations
+
+| Pattern | DXC | Clang |
+|---------|-----|-------|
+| Resource param with default, followed by param without | Error | Error |
```
inbelic: No error listed here

proposals/NNNN-local-resources-behavior.md
```diff
@@ -0,0 +1,341 @@
...
+### Texture2D vs RWByteAddressBuffer
+
+Some behaviors differ between resource types. DXC ICEs with
+`static Texture2D` but accepts `static RWByteAddressBuffer`. The test suite
+uses `RWByteAddressBuffer` because Clang does not yet support `Texture2D`.
+
+## Offload Test Suite
+
+Tests that produce a valid compiled output from at least one compiler are
+candidates for the
```
inbelic: I don't think we should use this as the basis. There might be cases that both compilers produce an incorrect error/crash and need to be corrected. Maybe that isn't the case and we can explicitly state that.

edit: this is probably more applicable to the below inclusion criteria seciton

proposals/NNNN-local-resources-behavior.md
inbelic: I guess the proposal is more to list all the interesting cases and not to explicitly highlight what is incorrect/correct so maybe the second part is less relevant

proposals/NNNN-local-resources-behavior.md
bob80905: No reason, I'll remove the bolds.

proposals/NNNN-local-resources-behavior.md
bob80905: Yes, it is intended to mean compiled with 0 errors and 0 warnings, I'll add a definition above just in case.
And yes, the doc is not making any statements on the correct / intended behavior of local resources, it is simply documenting the current behavior of local resources across a variety of edge cases on both compilers.

proposals/NNNN-local-resources-behavior.md
inbelic: Resolved offline.

We shouldn't be adding a test that isn't intended to run to the offload test suite. The case when both compilers are currently producing errors but shouldn't be can be audited at a later point.

proposals/NNNN-local-resources-behavior.md
```diff
@@ -0,0 +1,340 @@
...
+### Invalid Declarations
+
+| Pattern | DXC | Clang |
+|---------|-----|-------|
+| Resource param with default, followed by param without | Error: *"missing default argument on parameter"* | Error: *"missing default argument on parameter"* |
+| Resource type as `RWStructuredBuffer` element (intangible) | Error: *"is an object and cannot be used as a type parameter"* | Error: *"constraints not satisfied for class template"* |
+| Brace (zero) init `= {}` | Error: empty initializer list | Error: empty initializer list |
+| Compile-time out-of-bounds array index | Error: *"array index N is out of bounds"* | Warning (`-Warray-bounds`) |
+
+### Ternary Conditional Resource Assignment (CodeGen)
```
inbelic, line 236: I would have expected these to be an error in clang as well. Although not at the semahlsl level if that is what is being checked to determine whether it is a warning/error.

proposals/NNNN-local-resources-behavior.md
```diff
@@ -0,0 +1,340 @@
...
+end-to-end correctness beyond what static compilation checks can verify.
+
+### Inclusion criteria
+
+A test is added to the offload test suite when **at least one compiler
+produces a compiled output** — that is, compilation succeeds with zero
+errors. Tests that emit only warnings (e.g. `-Whlsl-explicit-binding`)
+still produce compiled output and qualify. Tests are placed in
+subdirectories based on which compiler(s) succeed and where the other
+compiler fails (sema vs codegen).
```
inbelic: I guess one thing here, has there been a check to what test cases are already covered?

bob80905: No there hasn't, though I think it would be somewhat reasonable for there to be duplicates.
In the same code, we could, for example, test the AST in SemaHLSL to ensure that resource AST nodes are properly present, while the same HLSL can be used to verify local resource behavior. I know @hekota  has some similar looking tests to the ones proposed, though the intent of the test is completely different from this.

proposals/NNNN-local-resources-behavior.md
bob80905: Yep, but IIRC, this will be a key difference between clang and DXC, so I believe the intent is for clang not to reject this.

inbelic: The intent is that clang should reject it, just not during sema, this is because it is not easy to determine whether or not the used resource is unique at that AST level. Instead we generate a warning and then error later during lowering to dxil.

There is a potential future where we do support passing handles to resources around, but that isn't something that is planned.

proposals/NNNN-local-resources-behavior.md
inbelic: Imo we definitely shouldn't have exact duplicate tests. If they use the same source but are testing different things that is okay, they aren't duplicates. Ideally I'd think it might be nice to combine those cases in one file with multiple run lines (if the directory makes sense for both). There is a cost to having redundant tests.

proposals/NNNN-local-resources-behavior.md
```diff
@@ -0,0 +1,361 @@
...
+| Reassign to same global in conditional branch | Clean | Clean |
+| Return initialized local | Clean | Clean |
+| Return uninitialized local | Clean | Clean |
+| Parenthesized ternary expression init | Clean | Clean |
+| Aggregate init of struct with resource | Clean | Clean |
+| Two resources in a single declaration | Clean | Clean |
+| Use of uninitialized local resource | Clean | Clean (sema), Error (codegen): DXIL Op Lowering assertion on uninitialized resource Store |
+| Default-init store (unbound handle) | Error (codegen): *"local resource not guaranteed to map to unique global resource"* | Clean (sema), Error (codegen): DXIL Op Lowering assertion on unbound handle |
+| Conditional init (`if(cond) buf = g;`) | Error (codegen): *"local resource not guaranteed to map to unique global resource"* | Clean |
+| Comma expression init (`(gBuf0, gBuf1)`) | Clean | Warning: *"left operand of comma operator has no effect"* |
```
bogner: I find these tables kind of unwieldy - they're very hard to read in the markdown source and are pretty noisy when rendered as well. We're also missing a crucial piece of information here, which is "what is the expected behaviour?", and it isn't clear if "Clean" is actually what we want in a lot of these cases.

What if we formatted these something like so, where we say what's expected and use a checkmark to indicate that the compiler matches the expectation (some useful examples cherry picked):

---
*Alias from global (`buf = gBuf0`)*

* Expected: Allowed, creates a local variable that references the global
  resource.
* DXC: :white_check_mark:
* Clang: :white_check_mark:

*Self-assignment of uninitialized resource (`Out = Out`)*

* Expected: Error (use of uninitialized resource)
* DXC: :white_check_mark: Error (codegen): `local resource not guaranteed to map to unique global resource`
* Clang: :x: Compiles cleanly - statement is ignored

*Default-init store (unbound handle)*

* Expected: Error for unbound resource use
* DXC: :white_check_mark: Error (codegen): `local resource not guaranteed to map to unique global resource`
* Clang: :x: Crash in codegen (DXIL Op Lowering assertion on unbound handle)

*Comma expression init (`(gBuf0, gBuf1)`)*

* Expected: Allowed, value is last value of the comma expression
* DXC: :white_check_mark:
* Clang: :white_check_mark: - note: warns `left operand of comma operator has no effect`
---

proposals/NNNN-local-resources-behavior.md
```diff
@@ -0,0 +1,361 @@
...
+
+> **Clean** means compilation with 0 warnings and 0 errors.
+
+| Pattern | DXC | Clang |
+|---------|-----|-------|
+| Alias from global (`buf = gBuf0`) | Clean | Clean |
+| Alias chain (`a = g; b = a; c = b`) | Clean | Clean |
+| Copy between locals (`a = g; b = a`) | Clean | Clean |
+| Self-assignment (`buf = buf`) | Clean | Clean |
+| Self-assignment of uninitialized resource (`Out = Out`) | Error (codegen): *"local resource not guaranteed to map to unique global resource"* | Clean |
```
bogner: We want an error here, right?

proposals/NNNN-local-resources-behavior.md
```diff
@@ -0,0 +1,361 @@
...
+| Reassign to same global in conditional branch | Clean | Clean |
+| Return initialized local | Clean | Clean |
+| Return uninitialized local | Clean | Clean |
+| Parenthesized ternary expression init | Clean | Clean |
+| Aggregate init of struct with resource | Clean | Clean |
+| Two resources in a single declaration | Clean | Clean |
+| Use of uninitialized local resource | Clean | Clean (sema), Error (codegen): DXIL Op Lowering assertion on uninitialized resource Store |
```
bogner: I'm surprised that DXC is "clean" here - shouldn't this be an error? There's no way for this to do something sensible.

proposals/NNNN-local-resources-behavior.md
```diff
@@ -0,0 +1,361 @@
...
+[offload test suite](https://github.com/llvm/offload-test-suite) under
+`test/Feature/LocalResources/`. The offload test suite executes shaders
+at runtime against real GPU hardware and software rasterizers, validating
+end-to-end correctness beyond what static compilation checks can verify.
+
+### Inclusion criteria
+
+A test is added to the offload test suite when **at least one compiler
+produces a compiled output** — that is, compilation succeeds with zero
+errors. Tests that emit only warnings (e.g. `-Whlsl-explicit-binding`)
```
bogner: This doesn't make sense to me - tests should only be in the offload test suite if there is an expected valid/correct behaviour. This may be different than at least one compiler succeeding at compiling something, as there may be cases where the compiler does an obviously wrong thing but succeeds at compiling, and there is no sensible "right" thing to do.

proposals/NNNN-local-resources-behavior.md
```diff
@@ -0,0 +1,361 @@
...
+
+The **57 tests** that produce a compiled output from both compilers are
+placed directly in `test/Feature/LocalResources/`. This includes tests
+that are fully clean on both compilers, tests where Clang
+emits `-Whlsl-explicit-binding` warnings but still produces compiled
+output (reassignment patterns across control flow, switches, and loops),
+and 1 test (`local_resource_comma_init.hlsl`) where Clang warns about a
+discarded comma operand but both compilers produce output.
+
+Three subdirectories capture tests that compile on only one compiler:
```
bogner: Subdirectories per compiler doesn't sound right - shouldn't we simply XFAIL the tests for the compiler that doesn't succeed? We should be filing an issue for any of these tests that is indicating a bug, in any case.

proposals/NNNN-local-resources-behavior.md
```diff
@@ -0,0 +1,361 @@
...
+  or error in Clang at the sema stage, regardless of DXC behavior.
+- **`offload-test-suite/test/Feature/LocalResources/`** — Tests where
+  both compilers produce a compiled output (warnings are acceptable).
+- **`offload-test-suite/.../ClangPass-DXCCodegenError/`** — Tests where
+  Clang produces a compiled output but DXC fails to produce output due
+  to a codegen-stage error or ICE.
+- **`offload-test-suite/.../DXCPass-ClangSemaError/`** — Tests where DXC
+  produces a compiled output but Clang fails at sema.
+- **`offload-test-suite/.../DXCPass-ClangCodegenError/`** — Tests where DXC
+  produces a compiled output but Clang crashes during DXIL codegen.
```
bogner: I think we need the categories to be more like:
- clang/test/SemaHLSL/Resources/Local-Resources/ for tests that emit warnings and errors in clang's Sema
- llvm/test/CodeGen/DirectX/<somewhere> for errors that can only be resolved late (some of the ambiguous resource cases come to mind) and thus can't be handled in Sema. This includes cases where Sema warns but the backend has a hard error.
- offload-test-suite/test/Feature/LocalResources/ for offload tests of correct behaviour
- Probably some tests for correct codegen in clang/test/CodeGenHLSL and llvm/test/CodeGen/DirectX as we implement and fix bugs

proposals/NNNN-local-resources-behavior.md
bob80905: I personally find the tables readable, though depending our resolution to the point below I would be willing to adjust the table format to the format you've written. However, I disagree at this point with including expected behavior. This will open up a huge can of worms on what the compiler ought to do and ought not to do, for 70+ tests, and will likely be very unproductive. We absolutely should intend to clearly delineate compiler behavior for each and every test proposed here, but the purpose of this PR is to *document* what each compiler does, and possibly in some cases, allow that documentation to drive our decisions on what our expectations ought to be. 

In short, I think it is alright in this stage to not make a claim on expected compiler behavior, and to directly document behavior. Future work can take care of making ought-type decisions, but this proposal is strictly about documenting local resource behavior.
For further context, see the comment @inbelic made above:
image

proposals/NNNN-local-resources-behavior.md
bob80905: I would think that we would want clang to emit an error here, yes. Though at this point, I don't believe that question is within the scope of this proposal.

proposals/NNNN-local-resources-behavior.md
bob80905: My opinion is that this ought to be an error, though as I commented above, I think this is out of the scope of the proposal.

proposals/NNNN-local-resources-behavior.md
bob80905: Suppose there exists a shader that ought to be compiled into a clean output, but it has undefined runtime behavior.
Does this shader exist? Either the compiler ought to emit an error about undefined runtime behavior, in which case it wouldn't be placed in the offload test suite, or the compiler ought to compile cleanly and we'd get different  runtime behavior based on IHV.
Suppose 2 IHV drivers, A and B, run the shader, and A crashes, while B runs fine.
I would think in this case, we'd effectively utilize the offload test suite to catch the fact that A crashes, and maybe file a driver bug on A? 
I am not quite sure what we'd do in this case, but I would think there would be value in having added it to the offload test suite, to identify that peculiarity in behavior.

But the other option you seem to be suggesting: we keep the test in the llvm repo (so we can't test compilation on DXC, unless we add a copy of the test to DXC, which we probably want to avoid). And we just test that the shader compiles. 
The added difficulty here is that we will need to make a judgement call for each test on whether the test's use case is valid / correct. This is something I want to avoid in this proposal if possible.

Given these concerns, does adding these tests to the offload test suite make sense?

proposals/NNNN-local-resources-behavior.md
bob80905: While normally we would XFAIL tests that fail on specific compilers and file a bug, in this specific context I think it is not always the case that we would want to do this. 
Some shaders may exist that ought to behave one way and compile cleanly on DXC, but should fail on clang.
We wouldn't want to associate a bug on clang with this shader.
Same thing in the inverse case.
So, my reasoning behind this was, what if we organized the tests so that we could quickly identify / locate HLSL local resource use cases that are compiler-specific?

It would be nice to say "I want to know all the local resource use cases that are specific to DXC and pass on the DXC compiler, but the clang compiler rejects it."
These subdirectories allow for this organization, and prevent us from needing to add an associated bug.
I think this class of tests are somewhat different than the other tests, since some are actually not expected to behave identically across compilers.
So the idea is, tests that do behave identically (note, **not ought**, but **do**), are placed together, for organization, and tests that behave differently are categorized.

What do you think?

My vision here was that this set of tests initially documents test behavior only, but over time, transforms into "ought", what we want and expect the tests to behave like. So I think there is value in isolating tests where they ought / we want them to behave differently based on compiler.

proposals/NNNN-local-resources-behavior.md
bob80905: My comment in response to your "Subdirectories per compiler " comment above addresses this comment.

bogner:
> I disagree at this point with including expected behavior. This will open up a huge can of worms on what the compiler ought to do and ought not to do, for 70+ tests, and will likely be very unproductive. We absolutely should intend to clearly delineate compiler behavior for each and every test proposed here, but the purpose of this PR is to document what each compiler does, and possibly in some cases, allow that documentation to drive our decisions on what our expectations ought to be.

I understand your concern about potentially having to figure out expectations for every single case here in order to make progress, but I don't think pretending that we have no expectations whatsoever is doing us any good. Functionally, this PR is saying that our expectations are that both compilers "compile cleanly" for every thing listed here, but I don't really see how that's a useful metric. It also makes it harder to talk about what the compilers are actually doing - if it compiles cleanly but generates garbage code it's a different thing than if it compiles the operation successfully, and if we don't differentiate in this document we basically have to do the entire investigation again for each operation as we go.

I think a lot of the cases here have very obvious expected behaviour, and at the very least we should be listing those and comparing the compiler's behaviour to that. This tells us that we need a test, and easily points out whether or not we're going to need to file issues and XFAIL that test. This makes it much easier to differentiate between "this is valid HLSL" and "the compiler should emit an error here" in ways that still give us the space to recognize that a compiler crashing doesn't conform to "the compiler should emit an error".

However, we should also consider that "the behaviour is undefined", or "we don't have consensus or a clear picture what the behaviour should be" are valid. This would also make it clear where we need to file issues on the HLSL spec work to go and do the work to decide what we want to do, including in cases where we want to deliberately differ between Clang and DXC.

> In short, I think it is alright in this stage to not make a claim on expected compiler behavior, and to directly document behavior. Future work can take care of making ought-type decisions, but this proposal is strictly about documenting local resource behavior.

If this is all we do here, the only reasonable next step is to immediately do the work to make the claims on expected compiler behaviour. I do not believe this is sufficient for us to decide which tests we need, and so we'll end up needing to do as I suggest above anyway, but instead of it being done in a comprehensive survey the conversation will be scattered among the PRs that add the tests themselves.

> Suppose there exists a shader that ought to be compiled into a clean output, but it has undefined runtime behavior.
Does this shader exist? Either the compiler ought to emit an error about undefined runtime behavior, in which case it wouldn't be placed in the offload test suite, or the compiler ought to compile cleanly and we'd get different runtime behavior based on IHV.
Suppose 2 IHV drivers, A and B, run the shader, and A crashes, while B runs fine.
I would think in this case, we'd effectively utilize the offload test suite to catch the fact that A crashes, and maybe file a driver bug on A?
I am not quite sure what we'd do in this case, but I would think there would be value in having added it to the offload test suite, to identify that peculiarity in behavior.

In this case we would be identifying an HLSL spec issue. There are essentially three options:
- The spec defines the behaviour, in which case there's a correct result and any driver that doesn't have that result is incorrect.
- The spec disallows the behaviour, in which case the compilers should reject the shader and emit an error.
- The spec classifies the behaviour as undefined or implementation defined. This should generally be avoided, but in this case we can't write a meaningful runtime test.

> Some shaders may exist that ought to behave one way and compile cleanly on DXC, but should fail on clang.
We wouldn't want to associate a bug on clang with this shader.
Same thing in the inverse case.

I largely disagree. This is why documenting our expectations is important. Neither DXC nor Clang are the arbiter for what's correct, the HLSL spec is. For both HLSL 202X and 202Y, we should have clear definitions of what is correct and what isn't. If Clang or DXC disagrees with those definitions, then that's a bug in Clang or DXC (whether we intend to fix it or not). If those definitions are unclear or missing, that's an issue on the spec itself, and needs to be fixed. For cases where the behaviour in 202X differs from HLSL 2021 or earlier, or where 202Y differs from 202X, any tests need to specify the language version, and can only be meaningfully run on compilers that support that language version.

bob80905:
Though the clientstate godbolt URLs are long, they encode the source within the URL, so the links work without any dependence on godbolt's storage, the URLs are reproducible from the same source, they can be built offline, and there's no need to use godbolt's API or use POST.
These were some of the reasons to use the longer URLs as opposed to generating short links that require a one-time API call per test, trusting godbolt's persistence / having a URL that might change in the future.

bogner:
> Though the clientstate godbolt URLs are long...

Can you at least use out of line links so that there's at least a hope of trying to read the tables as text? That is:
```md
Here's [a link]

[a link]: https://example.com
```

proposals/NNNN-local-resources-behavior.md
```diff
@@ -0,0 +1,432 @@
...
+`DxilCondenseResources` / equivalent.
+
+### Basic Local Resource Operations
+
+> ✅ = currently matches the "ought" claim. ❌ = does not match (actual
+> behavior in parentheses).
+
+| Test | Behavior | Ought Compile | Ought Runtime | Clang | DXC |
+|------|----------|---------------|---------------|-------|-----|
+| [`local_resource_alias_global`][local_resource_alias_global] | Initialize a local from a single global, then Store through the local | Clean | Store reaches the global's binding | ✅ | ✅ |
```
bogner: There's nothing that disallows "`" in markdown link refs, so you could just write this as:
```suggestion
| [`local_resource_alias_global`] | Initialize a local from a single global, then Store through the local | Clean | Store reaches the global's binding | ✅ | ✅ |
```

Then later we simply do:
```
[`local_resource_alias_global`]: ...
```

proposals/NNNN-local-resources-behavior.md
```diff
@@ -0,0 +1,432 @@
...
+| [`local_resource_self_assign_uninitialized`][local_resource_self_assign_uninitialized] | `Out = Out` where `Out` is an uninitialized `out` parameter | Error † (use of uninitialized resource handle) | N/A | ❌ Clang ICEs at codegen (DXIL Op Lowering pass) | ❌ Codegen error: *"local resource not guaranteed to map to unique global resource"* |
+| [`local_resource_reassign_same_global`][local_resource_reassign_same_global] | `if(c) buf=g; else buf=g;` (both branches assign the same global) | Clean | Store reaches `g` | ✅ | ✅ |
+| [`return_local_resource_initialized`][return_local_resource_initialized] | Return a local that aliases a global | Clean | Caller observes the global | ✅ | ✅ |
+| [`return_local_resource_uninitialized`][return_local_resource_uninitialized] | Return an uninitialized local from a function and Store through it | Error † (use of uninitialized resource handle) | N/A | ❌ Clean (sema); ICE at codegen (DXIL Prepare Module pass) | ✅ Codegen error: *"local resource not guaranteed to map to unique global resource"* |
```
bogner: Is this meant to be :x: for dxc? The other cases where you say the error should be "use of uninitialized resource handle" but the DXC error is this unique global resource one are marked :x:

proposals/NNNN-local-resources-behavior.md
```diff
@@ -0,0 +1,432 @@
...
+| [`local_resource_comma_init`][local_resource_comma_init] | `buf = (g0, g1)` — comma expression initializer | Warning † (left operand discarded) | Store reaches `g1` | ✅ Warning: *"left operand of comma operator has no effect"* | ✅ Warning: *"comma expression used where a constructor list may have been intended"* |
+
+### Parameter Passing
+
+| Test | Behavior | Ought Compile | Ought Runtime | Clang | DXC |
+|------|----------|---------------|---------------|-------|-----|
+| [`local_resource_out_param`][local_resource_out_param] | Resource passed as `out` parameter, assigned in callee, read by caller | Clean | Caller observes the assigned resource | ✅ | ✅ |
+| [`local_resource_inout_param`][local_resource_inout_param] | Resource passed as `inout` parameter; mutation in callee is visible to caller | Clean | Caller observes mutated state | ✅ | ✅ |
+| [`local_resource_const_param`][local_resource_const_param] | Resource passed as `const` parameter; `Store` invoked inside callee | Error † (Store mutates state through a `const`-qualified handle) | N/A | ✅ Error: *"no matching member function"* (`Store` not `const`-qualified) | ❌ Clean (silently accepts Store on a `const` resource) |
```
bogner: Is this accurate? Both compilers accept this and do the store in the godbolt link. I think that `const` applies to the resource object itself here, not the buffer it points to. That is, this is like `int *const` rather than `const int *`.

proposals/NNNN-local-resources-behavior.md
```diff
@@ -0,0 +1,432 @@
...
+| [`local_resource_array`][local_resource_array] | Plain local array of resources initialized from globals | Clean | Store reaches the indexed global | ✅ | ✅ |
+| [`local_resource_array_copy`][local_resource_array_copy] | Copy one local resource array to another, then Store through the copy | Clean | Stores reach the original globals | ❌ Warning (`-Whlsl-explicit-binding`): *"assignment of 'Aux' to local resource 'src' is not to the same unique global resource"* | ✅ |
+| [`local_resource_array_dynamic_index`][local_resource_array_dynamic_index] | Dynamic (runtime) index into a local resource array | Clean | Store reaches the dynamically-indexed global | ❌ Clean (sema); codegen error: *"Load of … is not a global resource handle"* | ✅ |
+| [`local_resource_array_partial_init`][local_resource_array_partial_init] | `{g0}` for a 2-element resource array (second element default-initialized) | Error † (every array element must be bound to a global) | N/A | ❌ Clean (sema); codegen error: *"Load of … is not a global resource handle"* | ❌ Clean |
```
bogner: Does this have the right link? I see:
```hlsl
RWByteAddressBuffer Out : register(u0);

[numthreads(1,1,1)]
void main(uint3 tid : SV_DispatchThreadID) {
    RWByteAddressBuffer arr[4];
    arr[0] = Out;
    arr[1] = Out;
    arr[tid.x & 3].Store(0, 42);
}
```

proposals/NNNN-local-resources-behavior.md
```diff
@@ -0,0 +1,432 @@
...
+| [`local_resource_array_size_one`][local_resource_array_size_one] | Resource array of size 1 | Clean | Store reaches the global | ✅ | ✅ |
+| [`struct_resource_member_reassign`][struct_resource_member_reassign] | Reassign a struct's resource member to a different global | Warning † (binding-ambiguous reassignment) | Store reaches the most-recently-assigned global | ❌ Clean (warning does not fire across struct member access) | ❌ Clean (no diagnostic) |
```
bogner: Note that the example here isn't ambiguous, given that it unconditionally reassigns the member:
```hlsl
RWByteAddressBuffer gBuf0 : register(u0);
RWByteAddressBuffer gBuf1 : register(u1);

struct ResHolder { RWByteAddressBuffer buf; };

[numthreads(1,1,1)]
void main(uint tid : SV_GroupThreadID) {
    ResHolder h;
    h.buf = gBuf0;
    h.buf = gBuf1;
    h.buf.Store(0, 42);
}
```

Since `h.buf` can only *ever* be `gBuf1` when it's accessed this code is silly but harmless.

I think you're trying to describe a case more like this, which is ambiguous and must be an error:
```
RWByteAddressBuffer gBuf0 : register(u0);
RWByteAddressBuffer gBuf1 : register(u1);

struct ResHolder { RWByteAddressBuffer buf; };

[numthreads(1,1,1)]
void main(uint tid : SV_GroupThreadID) {
    ResHolder h;
    h.buf = gBuf0;
    if (tid.x)
      h.buf = gBuf1;
    h.buf.Store(0, 42);
}
```

proposals/NNNN-local-resources-behavior.md
```diff
@@ -0,0 +1,432 @@
...
+| [`local_resource_struct_method`][local_resource_struct_method] | Struct with a member function that operates on a resource member | Clean | Store reaches the global via the member function | ✅ | ✅ |
+
+### Control Flow
+
+| Test | Behavior | Ought Compile | Ought Runtime | Clang | DXC |
+|------|----------|---------------|---------------|-------|-----|
+| [`local_resource_shadow_inner_scope`][local_resource_shadow_inner_scope] | Local resource shadowed by another local in an inner block | Clean | Inner-scope Store reaches inner global; outer Store reaches outer global | ✅ | ✅ |
+| [`local_resource_block_lifetime`][local_resource_block_lifetime] | Resource assigned in inner block, used in outer scope | Clean | Store reaches the assigned global | ✅ | ✅ |
+| [`local_resource_nested_blocks_reassign`][local_resource_nested_blocks_reassign] | Reassigned to a different global across nested blocks | Warning † (binding-ambiguous reassignment) | Store reaches the most-recently-assigned global | ✅ Warning (`-Whlsl-explicit-binding`) | ❌ Clean |
```
bogner: Again, this example is straight line code, so there isn't really an ambiguity. I'm actually a bit surprised that clang's warning fires here.

proposals/NNNN-local-resources-behavior.md
```diff
@@ -0,0 +1,432 @@
...
+| [`local_resource_early_return_reassign`][local_resource_early_return_reassign] | Reassigned to a different global on an early-return path | Pending discussion on [#427](https://github.com/llvm/wg-hlsl/issues/427); current Clang choice is Warning † (binding-ambiguous reassignment) | Store reaches the assigned global per taken path | ✅ Warning (`-Whlsl-explicit-binding`) | ❌ Clean |
+| [`local_resource_unreachable_reassign`][local_resource_unreachable_reassign] | Reassigned in provably unreachable code | Pending discussion on [#427](https://github.com/llvm/wg-hlsl/issues/427); current Clang choice is Warning † (binding-ambiguous reassignment; possibly also dead-code warning) | Store reaches the original global (unreachable code does not execute) | ✅ Warning (`-Whlsl-explicit-binding`) | ❌ Clean |
+| [`local_resource_switch_reassign`][local_resource_switch_reassign] | Reassigned to a different global inside switch cases | Warning † (binding-ambiguous reassignment) | Store reaches the per-case-assigned global | ✅ Warning (`-Whlsl-explicit-binding`) | ❌ Clean |
```
bogner: Note that this is only clean on DXC because the switch condition is a constant. I think the runtime result here needs to differentiate between whether we can resolve the ambiguity or not - if we only call this function with constants, then yes "Store reaches the per-case-assigned global", but if we call this with a runtime value it needs to error. It currently correctly errors on DXC and has the error-then-ICE behaviour on clang: https://godbolt.org/z/xTrcfb5ro

bogner (commented):
Generally looking good but I have a couple of comments we should sort out before getting this in.

proposals/NNNN-local-resources-behavior.md
bob80905: I've split it into 2 separate tests so that the store is done both within function scope as a parameter, and separately in main as a local const resource.

proposals/NNNN-local-resources-behavior.md
bob80905: Bad description, I have made a test to reflect this description and updated the description, and created a description for this existing test.

proposals/NNNN-local-resources-behavior.md
bob80905: I've left this unambiguous case, updated the test name to indicate as such, and added a variant of this test that is ambiguous.

proposals/NNNN-local-resources-behavior.md
bob80905: Same as above, named this unambiguous, and added an ambiguous test case.

proposals/NNNN-local-resources-behavior.md
bob80905: Yep, as above I made an ambiguous variant.

bogner (approved):

hekota (approved):

PR #420, opened 2026-04-30
[0037] Update proposal to use `resource_load_level`, and add proposal for multisampled texture loads with `resource_load_ms`

Icohedron:
This PR updates proposal 0037 to reflect the actual implementation of the texture load intrinsic used by Load and mips.operator[][] (https://github.com/llvm/llvm-project/pull/185708)  (`__builtin_hlsl_resource_load` -> `__builtin_hlsl_resource_load_level`).

This PR also proposes defining a new load intrinsic for multisampled textures: `__builtin_hlsl_resource_load_ms` which works similarly.

### DIFF
--- a/proposals/0037-texture-and-sampler-types.md
+++ b/proposals/0037-texture-and-sampler-types.md
@@ -176,7 +176,7 @@ public:
 
   // Load / Access
   T Load(int3 location) {
-    return __builtin_hlsl_resource_load_level(Handle, location.xy, location.z);
+    return __builtin_hlsl_resource_load_level(Handle, location);
   }
   // ... other Load overloads ...
 
@@ -190,7 +190,7 @@ public:
     __hlsl_resource_t Handle;
     uint MipLevel;
     T operator[](int2 Loc) {
-      return __builtin_hlsl_resource_load_level(Handle, Loc, MipLevel);
+      return __builtin_hlsl_resource_load_level(Handle, uint3(Loc, MipLevel));
     }
   };
 
@@ -542,7 +542,7 @@ target-specific LLVM intrinsics in Clang codegen. The naming convention follows
 
 The `__builtin_hlsl_resource_load_level` builtin handles non-multisampled
 texture loads. The mip level is packed into the coordinate vector (e.g., `int3`
-for a 2D texture = xy + mip), and `Offset` defaults to 0 if not provided. For
+for a 2D texture `(x, y, mip)`), and `Offset` defaults to 0 if not provided. For
 multisampled textures, the `__builtin_hlsl_resource_load_ms` builtin is used,
 which takes an explicit sample index as a separate parameter.
 
--- a/proposals/0037-texture-and-sampler-types.md
+++ b/proposals/0037-texture-and-sampler-types.md
@@ -176,7 +176,7 @@ public:
 
   // Load / Access
   T Load(int3 location) {
-    return __builtin_hlsl_resource_load(Handle, location.xy, location.z);
+    return __builtin_hlsl_resource_load_level(Handle, location.xy, location.z);
   }
   // ... other Load overloads ...
 
@@ -190,7 +190,7 @@ public:
     __hlsl_resource_t Handle;
     uint MipLevel;
     T operator[](int2 Loc) {
-      return __builtin_hlsl_resource_load(Handle, Loc, MipLevel);
+      return __builtin_hlsl_resource_load_level(Handle, Loc, MipLevel);
     }
   };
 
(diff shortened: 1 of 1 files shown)


proposals/0037-texture-and-sampler-types.md
```diff
@@ -581,7 +586,8 @@ operations.
 | `llvm.dx.resource.samplelevel`                                       | `dx.op.sampleLevel` (62)                  |
 | `llvm.dx.resource.samplecmp`<br>`llvm.dx.resource.samplecmp.clamp`   | `dx.op.sampleCmp` (64)                    |
 | `llvm.dx.resource.samplecmplevelzero`                                | `dx.op.sampleCmpLevelZero` (65)           |
-| `llvm.dx.resource.load.texture`                                      | `dx.op.textureLoad` (66)                  |
+| `llvm.dx.resource.load.level`                                        | `dx.op.textureLoad` (66)                  |
+| `llvm.dx.resource.load.ms`                                           | `dx.op.textureLoad` (66)                  |
```
bob80905, line 590: Would these need to have different op codes?

Icohedron: Nope. The DXIL op is the same between the texture loads.
We just have two different HLSL intrinsics to better match the HLSL interface.

You can also see here that DXC lowers both .mips and .sample to the same DXIL op: https://hlsl.godbolt.org/z/nndvoW6r5

proposals/0037-texture-and-sampler-types.md
```diff
@@ -520,7 +522,8 @@ used.
 - `T __builtin_hlsl_resource_sample_level(Handle, Sampler, Coord, float LOD, int2 Offset = 0)`
 - `float __builtin_hlsl_resource_sample_cmp(Handle, Sampler, Coord, float CompareValue, int2 Offset = 0)`
 - `float __builtin_hlsl_resource_sample_cmp_level_zero(Handle, Sampler, Coord, float CompareValue, int2 Offset = 0)`
-- `T __builtin_hlsl_resource_load(Handle, Coord, int MipLevelOrSampleIndex = 0, int2 Offset = 0)`
+- `T __builtin_hlsl_resource_load_level(Handle, int3 CoordWithMip, int2 Offset = 0)`
+- `T __builtin_hlsl_resource_load_ms(Handle, Coord, int SampleIndex, int2 Offset = 0)`
```
bob80905, line 526: Does Coord need a type specifier here, like int3?

proposals/0037-texture-and-sampler-types.md
```diff
@@ -537,10 +540,11 @@ The HLSL builtins used in the record type implementation are lowered to
 target-specific LLVM intrinsics in Clang codegen. The naming convention follows
 [0014 - Consistent Naming for DX Intrinsics](0014-consistent-naming-for-dx-intrinsics.md).
 
-The `__builtin_hlsl_resource_load` builtin will support default values for
-optional operands. Specifically, `SampleIndex` will default to 0 for
-non-multisampled textures, and `Offset` will default to 0 if not provided. This
-allows a single builtin to handle the various `Load` overloads.
+The `__builtin_hlsl_resource_load_level` builtin handles non-multisampled
+texture loads. The mip level is packed into the coordinate vector (e.g., `int3`
+for a 2D texture = xy + mip), and `Offset` defaults to 0 if not provided. For
```
bob80905: So if I understand correctly, the argument CoordWithMip is an int3, where each element for a 2D texture is: `x, y, mip`? or would the user pass `x, y, xy + mip`?

bob80905 (commented):
couple more questions

proposals/0037-texture-and-sampler-types.md
Icohedron: I followed the same pattern as the other definitions and left it unspecified.

proposals/0037-texture-and-sampler-types.md
Icohedron: It is `x, y, mip`

Icohedron: https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-to-load#parameters

The first arg of `.Load` is `Location` which the msft doc says is
> [in] The texture coordinates; the last component specifies the mipmap level. 

and the type of `Location` for `Texture2D` is an `int3`

bob80905 (approved):
LGTM, I would just change the phrasing form xy + mip to x, y, mip as a nit, but it makes sense to me now.

s-perron (approved):

PR #421, opened 2026-05-07
Add minutes for HLSL Working Group meeting on 2026-05-07

farzonl:


meetings/2026-05-07.md
```diff
@@ -0,0 +1,8 @@
+---
+title: 2026-05-07 - HLSL Working Group Minutes
+---
+
+* Discussion topics
+* <place_holder>
```
farzonl:
```suggestion
* Many llvm intrinsic in spir-v  are not legalized for matrix types larger than a 2x2.
* <place_holder>
* <place_holder>
```

farzonl: examples include cases of transpose of bool matrix and now atan.

meetings/2026-05-07.md
```diff
@@ -0,0 +1,8 @@
...
+* <place_holder>
```
s-perron:
```suggestion
* ConstantBuffer<T> and constructor work
```

meetings/2026-05-07.md
```diff
@@ -0,0 +1,10 @@
+---
+title: 2026-05-07 - HLSL Working Group Minutes
+---
+
+* Discussion topics
+* Many llvm intrinsic in spir-v  are not legalized for matrix types larger than a 2x2.
+* <place_holder>
+* <place_holder>
+* ConstantBuffer<T> and constructor work
+* <place_holder>
```
bogner:
```suggestion
## Discussion Topics

* SPIR-V matrix legalization: Many intrinsics are not legalized for matrix
  types larger than a 2x2
* ConstantBuffer<T> and constructor work

## Meeting Notes

### SPIR-V matrix legalization

Steven and Farzon discussed how to get legalization working across all of the
matrix intrinsics, and whether we need to handle each manually. A few options
were identified:

* Use the approach we've used for constant buffers and represent matrices in
  the frontend using arrays-of-vectors.
* Generically legalize using scalarization.
* Build a library of legalization helpers and implement legalization op by op.

There was some concern that the frontend approach feels like it might be
papering over the problem, but the only real downside would be a potential
compile time cost due to having more code during optimization. Generic
legalization using scalarization will not be optimal, so it would probably just
be a stopgap. Legalization helpers might work, but due to the legalizer's love
of powers of two it may end up suboptimal for things like 3x3 matrices, whereas
the frontend approach would be able to handle this trivially.

### ConstantBuffer<T> and constructors

Steven and Helena discussed coordinating getting the ConstantBuffer<T> and
constructor PRs in.

Issues with ConstantBuffer<T> and memcpy were discussed, notably that small
memcpys can introduce types out of thing air. This is likely not an issue long
term, since the upstream LLVM byte type work should avoid this in the future.
```

bogner (approved):

PR #425, opened 2026-05-26
[0047] Add DirectX metadata schema for input/output semantic signatures

### LINKED ISSUE #143: Design Signature metadata for IR
This issue tracks the work required to write out a metadata schema, as described here: https://github.com/llvm/wg-hlsl/blob/main/proposals/0044-vertex-pixel-shader-directx-workstreams.md#metadata-schema-for-signature-information.

AC:
- [ ] Create a new proposal towards wg-hlsl that describes the metadata schema and generation strategy
- [ ] Update https://github.com/llvm/llvm-project/issues/57928 with an implementation specification

inbelic:
Create a reference for implementation of how we can representation input and output signatures as a named metadata node in llvm ir. As well as detail the strategy for generating it from code generation such that it can retain all required information for serialization into ISG, PSV parts, etc in the backend.

Assisted by: Claude Opus 4.7

Resolves: https://github.com/llvm/wg-hlsl/issues/143


proposals/NNNN-semantic-signatures-metadata-schema.md
```diff
@@ -0,0 +1,246 @@
+---
+title: "[NNNN] - Semantic Signature Metadata"
+params:
+  authors:
+    - inbelic: Finn Plummer
+  status: Under Consideration
```
inbelic:
```suggestion
  status: Accepted
```

proposals/0047-semantic-signatures-metadata-schema.md
```diff
@@ -0,0 +1,246 @@
...
+|  3 | i32             | semantic kind; `Arbitrary` (0) for user-defined semantics, the corresponding `SV_*` value otherwise. See [`SEMANTIC_KIND`](https://github.com/llvm/llvm-project/blob/main/llvm/include/llvm/BinaryFormat/DXContainerConstants.def) |
+|  4 | metadata node   | reference to a [semantic indices](#semantic-indices) node |
+|  5 | i32             | interpolation mode; see [`INTERPOLATION_MODE`](https://github.com/llvm/llvm-project/blob/main/llvm/include/llvm/BinaryFormat/DXContainerConstants.def) |
+|  6 | i32             | number of consecutive register rows occupied |
+|  7 | i8              | number of components per row (1–4) |
+|  8 | i32             | starting register row; `-1` (`0xFFFFFFFF`) if unallocated |
+|  9 | i8              | starting component column; `-1` (`0xFF`) if unallocated, otherwise 0–3 |
+| 10 | i8              | 4-bit bitmask of components that are always read (input) or may be written (output). |
+| 11 | i8              | 4-bit bitmask of components that are dynamically indexed |
+| 12 | i32             | GS output stream index; 0 for non-GS stages |
```
inbelic, line 79: IIUC these (10-12) are often not relevant, we could store these as a separate name, value metadata node pair list when they are used. This would match how dxc represents this internally. However, it makes parsing more awkward.

proposals/NNNN-semantic-signatures-metadata-schema.md
```diff
@@ -0,0 +1,246 @@
...
+A signature element describes a single packed range of signature rows with
+its semantic name and indices, register placement, component masks, and
+stage-specific attributes. Each element must retain all relevant information to
+serialize into `ISG1`, `OSG1` and `PSV0`.
+
+| #  | Type            | Description |
+|----|-----------------|-------------|
+|  0 | i32             | dense 0-based index within the entry function signature list; matches the operand of `llvm.dx.load.input` / `llvm.dx.store.output` |
+|  1 | metadata string | the semantic name (e.g. `!"TEXCOORD"`, `!"SV_Position"`) |
+|  2 | i32             | component type; see [`llvm::dxil::ElementType`](https://github.com/llvm/llvm-project/blob/main/llvm/include/llvm/Support/DXILABI.h). Equivalent to DXC's [`DXIL::ComponentType`](https://github.com/microsoft/DirectXShaderCompiler/blob/main/include/dxc/DXIL/DxilConstants.h); the two enums share value assignments through `PackedU8x32 = 18`. The SM 6.9 additions (`I8`, `U8`, `F8_E4M3FN`, `F8_E5M2`) will need to be mirrored into `ElementType`. |
```
inbelic:
```suggestion
|  2 | i32             | component type; see [`llvm::dxil::ElementType`](https://github.com/llvm/llvm-project/blob/main/llvm/include/llvm/Support/DXILABI.h). |
```

proposals/NNNN-semantic-signatures-metadata-schema.md
```diff
@@ -0,0 +1,246 @@
...
+  accesses used a non-`poison` dynamic index operand).
+
+### Emitting the Metadata Table
+
+At the end of `emitEntryFunction`, with the two vectors fully populated and
+packed, the metadata is materialized in one pass:
+
+1. For each `SemanticSignatureElement`, build the operand-4 semantic-indices
+   node (`!{ i32 ... }`) and the 13-operand element node described in
+   [Signature Element](#signature-element).
```
inbelic:
```suggestion
1. For each `SemanticSignatureElement`, build the semantic-indices
   node (`!{ i32 ... }`) and the signature element node described in
   [Signature Element](#signature-element).
```

proposals/NNNN-semantic-signatures-metadata-schema.md
```diff
@@ -0,0 +1,246 @@
...
+
+A `SemanticSignatureElement` struct will be defined in
+`llvm/Frontend/HLSL/SemanticSignatures.h` that will contain all the operands as
+above, as well as, helpers to determine the extra container field information
+(eg. is allocated). It will also define provide functions to convert to/from
+the metadata representation and the packing algorithm will is defined here.
+
+### Accumulate `SemanticSignatureElement`s
+
+`emitEntryFunction` will then have a a vector of `SemanticSignatureElement`s for
```
inbelic:
```suggestion
`emitEntryFunction` will then have a vector of `SemanticSignatureElement`s for
```

alsepkow:
Your Resolves link just links to the PR

proposals/NNNN-semantic-signatures-metadata-schema.md
```diff
@@ -0,0 +1,246 @@
...
+  authors:
+    - inbelic: Finn Plummer
+  status: Accepted
+---
+
+## Introduction
+
+This proposal defines the LLVM IR module-level metadata schema for HLSL shader
+input and output signatures when targeting DirectX, and describes how that
+metadata is generated by Clang's HLSL code generation.c
```
bob80905: nit: typo "generation .c"

proposals/NNNN-semantic-signatures-metadata-schema.md
```diff
@@ -0,0 +1,246 @@
...
+in-memory signature elements. Finally we can emit the in-memory representation
+to the metadata nodes as described above. This is described in more detail
+below.
+
+### In-Memory Representation
+
+A `SemanticSignatureElement` struct will be defined in
+`llvm/Frontend/HLSL/SemanticSignatures.h` that will contain all the operands as
+above, as well as, helpers to determine the extra container field information
+(eg. is allocated). It will also define provide functions to convert to/from
```
bob80905: nit: "It will also define provide functions ..."

alsepkow:
```suggestion
(eg. is allocated). It will also provide functions to convert to/from
```

proposals/NNNN-semantic-signatures-metadata-schema.md
```diff
@@ -0,0 +1,246 @@
...
+the metadata representation and the packing algorithm will is defined here.
```
alsepkow:
```
the metadata representation and the packing algorithm will is defined here.
```

Image

proposals/0047-semantic-signatures-metadata-schema.md
```diff
@@ -0,0 +1,246 @@
...
+
+## Metadata Schema
+
+#### Named Signature Table
+
+```LLVM
+!dx.semantic.signatures = !{ !1, !2 }
+```
+
+A named metadata node, `dx.semantic.signatures`, is used to identify the
```
bob80905, line 23: I believe that library shaders may have multiple entry-point function definitions, and each with their own semantic signatures. I could be wrong.
But in this specific case, would there be a way to distinguish between the entry points referenced by this metadata node? like
`dx.semantic.signatures.1`, `dx.semantic.signatures.2`, etc?
Or are library functions out of scope for this proposal?

bob80905 (commented):
Some comments while reviewing with the one and only @alsepkow

proposals/0047-semantic-signatures-metadata-schema.md
```diff
@@ -0,0 +1,246 @@
...
+!1 = !{i32 0, !"POSITION", i32 9, i32 0, !50, i32 0, i32 1, i8 4, i32 0, i8 0, i8 0, i8 0, i32 0}
+!2 = !{i32 1, !"TEXCOORD", i32 9, i32 0, !51, i32 0, i32 2, i8 4, i32 1, i8 0, i8 0, i8 0, i32 0}
+
+; vs_main outputs
+!3 = !{i32 0, !"SV_Position", i32 9, i32 3, !50, i32 0, i32 1, i8 4, i32 0, i8 0, i8 0, i8 0, i32 0}
+
+; ps_main inputs
+!4 = !{i32 0, !"SV_Position", i32 9, i32 3, !50, i32 2, i32 1, i8 4, i32 0, i8 0, i8 0, i8 0, i32 0}
+!5 = !{i32 1, !"TEXCOORD",    i32 9, i32 0, !50, i32 2, i32 1, i8 4, i32 1, i8 0, i8 0, i8 0, i32 0}
+!6 = !{i32 2, !"TEXCOORD",    i32 9, i32 0, !52, i32 2, i32 1, i8 4, i32 2, i8 0, i8 7, i8 0, i32 0}
```
alsepkow, line 155: !6 has DynIndexMask = i8 7 but the HLSL is o.extra = float4(uv1.xyz, 1). Which is a static swizzle? Not dynamic indexing. DynIndexMask should be 0?

proposals/NNNN-semantic-signatures-metadata-schema.md
```diff
@@ -0,0 +1,246 @@
...
+  `-1` (`0xFF`). The two sentinels are always set together:
+  `StartRow == -1` iff `StartCol == -1`.
+
+- Declared component mask: Computed from field 7 (`Cols`) and field 9
+  (`StartCol`) as `((1 << Cols) - 1) << StartCol`.
+
+- `AlwaysReads` / `NeverWrites` Masks: Field 10 is written directly to the
+  container's `AlwaysReads` mask for input signatures; for output signatures
+  the container's `NeverWrites` mask is the bitwise complement of field 10
+  against the declared component mask above.
```
alsepkow:
```suggestion
  the container's `NeverWrites = ~UsageMask & DeclaredMask`
```

proposals/NNNN-semantic-signatures-metadata-schema.md
```diff
@@ -0,0 +1,246 @@
...
+
+The following downstream container fields are derived from the operands
+above rather than carried in separate metadata operands:
+
+- Allocated status: A signature element is allocated iff field 8
+  (`StartRow`) is not `-1` (`0xFFFFFFFF`) and field 9 (`StartCol`) is not
+  `-1` (`0xFF`). The two sentinels are always set together:
+  `StartRow == -1` iff `StartCol == -1`.
+
+- Declared component mask: Computed from field 7 (`Cols`) and field 9
```
alsepkow:
```suggestion
- Declared component mask (DeclaredMask): Computed from field 7 (`Cols`) and field 9
```

proposals/0047-semantic-signatures-metadata-schema.md
```diff
@@ -0,0 +1,246 @@
...
+params:
+  authors:
+    - inbelic: Finn Plummer
+  status: Accepted
+---
+
+## Introduction
+
+This proposal defines the LLVM IR module-level metadata schema for HLSL shader
+input and output signatures when targeting DirectX, and describes how that
```
alsepkow, line 12: Where does the SPIR-V implementation and/or spec live? Does it exist yet?

proposals/NNNN-semantic-signatures-metadata-schema.md
alsepkow: Is the '.c' a typo?

alsepkow (commented):
A few comments from reviewing with @bob80905

proposals/0047-semantic-signatures-metadata-schema.md
```diff
@@ -0,0 +1,246 @@
...
+```
+
+A signature element describes a single packed range of signature rows with
+its semantic name and indices, register placement, component masks, and
+stage-specific attributes. Each element must retain all relevant information to
+serialize into `ISG1`, `OSG1` and `PSV0`.
+
+| #  | Type            | Description |
+|----|-----------------|-------------|
+|  0 | i32             | dense 0-based index within the entry function signature list; matches the operand of `llvm.dx.load.input` / `llvm.dx.store.output` |
```
joaosaffran, line 67: Q: I fell like this information is duplicated, we can use the metadata note position in the parent to infer this informationn can't we ? 

For example: 

```
!dx.semantic.signatures = !{ !1, !2 }
!1 = !{ ptr @vs_main, !3, !4 }
!3 = !{ !5, !6 }
!5 = !{ !"TEXCOORD", ... }
!6 = !{ !"TEXCOORD", ... }
```

Looking on the list of input signatures, metadata `!3` I can infer that metadata `!5` would be index 0 and `!6` would be index 1.

proposals/NNNN-semantic-signatures-metadata-schema.md
```diff
@@ -0,0 +1,246 @@
...
+!3 = !{ !5, !6 }
+```
+
+A signature element list consists of a list of references to signature
+element nodes.
+
+#### Signature Element
+
+```LLVM
+!1 = !{ i32 0, !"TEXCOORD", i32 9, i32 0, !50, i32 0, i32 1, i8 4, i32 0, i8 0, i8 0, i8 0, i32 0 }
```
joaosaffran: nit: Since the previous codesnippet listed `!5` as a signature element, I think this change will make all snippets link together to make an example of a valid signature metadata representation.

```suggestion
!5 = !{ i32 0, !"TEXCOORD", i32 9, i32 0, !50, i32 0, i32 1, i8 4, i32 0, i8 0, i8 0, i8 0, i32 0 }
```

proposals/0047-semantic-signatures-metadata-schema.md
```diff
@@ -0,0 +1,246 @@
...
+
+A named metadata node, `dx.semantic.signatures`, is used to identify the
+table of per-entry-point semantic signatures. The table itself is a list of
+references to function/signature triples. If no entry point in the module has
+any signature, the named metadata node may be omitted entirely.
+
+#### Function/Signature Triple
+
+```LLVM
+!1 = !{ ptr @vs_main, !3, !4 }
```
joaosaffran, line 31: Q: Are there multiple versions (like: v1, v2, etc) of input/output signatures throughout the different shader models?  Should that information be encoded in the metadata ?

proposals/0047-semantic-signatures-metadata-schema.md
```diff
@@ -0,0 +1,246 @@
...
+|  5 | i32             | interpolation mode; see [`INTERPOLATION_MODE`](https://github.com/llvm/llvm-project/blob/main/llvm/include/llvm/BinaryFormat/DXContainerConstants.def) |
+|  6 | i32             | number of consecutive register rows occupied |
+|  7 | i8              | number of components per row (1–4) |
+|  8 | i32             | starting register row; `-1` (`0xFFFFFFFF`) if unallocated |
+|  9 | i8              | starting component column; `-1` (`0xFF`) if unallocated, otherwise 0–3 |
+| 10 | i8              | 4-bit bitmask of components that are always read (input) or may be written (output). |
+| 11 | i8              | 4-bit bitmask of components that are dynamically indexed |
+| 12 | i32             | GS output stream index; 0 for non-GS stages |
+
+The following downstream container fields are derived from the operands
```
joaosaffran, line 81: Q: Why represent those fields in the metadata if they are going to be derived from other fields? Can't we calculate them in DirectX lowering instead? Or are they derived only in certain conditions?

joaosaffran (commented):
Left a few questions, mostly to help with my own understanding of signatures.

proposals/0047-semantic-signatures-metadata-schema.md
inbelic: Yes. This spec is specific to targeting DirectX.

The spir-v side is here: https://github.com/llvm/wg-hlsl/blob/main/proposals/0031-semantics.md#spir-v

proposals/0047-semantic-signatures-metadata-schema.md
inbelic: Correct. Each metadata node entry in this list will correspond to a single entry point. The first operand of which is a ptr to the entry function it corresponds to.

This should be sufficient for library shaders

proposals/0047-semantic-signatures-metadata-schema.md
inbelic: I think we could technically could. However, this value is what maps the load/store op to the table entry. It is imperative that these do not get out of sync. I would argue that having it as an operand is maybe less likely to be accidentally altered.

proposals/0047-semantic-signatures-metadata-schema.md
inbelic: Correct, the following are not stored as metadata

proposals/0047-semantic-signatures-metadata-schema.md
inbelic: The i8 7 corresponds to parameter 10 which is the elements that are always read/written. The next operand would be the dynamic indexing

proposals/0047-semantic-signatures-metadata-schema.md
inbelic: There is only the ISG1/OSG1 parts in DXIL that don't have different versions. PSV does have multiple versions but we can populate the part independently of this. We don't need version info in the metadata for version specific validations either. So this isn't required

joaosaffran (approved):
LGTM

proposals/0047-semantic-signatures-metadata-schema.md
alsepkow: Should we put DirectX in the title?

proposals/NNNN-semantic-signatures-metadata-schema.md
```diff
@@ -0,0 +1,245 @@
...
+
+## Generating Metadata
+
+The `dx.semantic.signatures` table is produced entirely from within
+`CGHLSLRuntime::emitEntryFunction` in `clang/lib/CodeGen/CGHLSLRuntime.cpp`.
+Briefly, this is done so by traversing through each of the entry function's
+input parameters and return outputs, accumulating an in-memory representation
+of the signature elements and emitting the load/store intrinsic. It uses the
+order the parameters/return outputs are declared in to assign the signature id
+that will associate the semantic in the intrinsic call and the metadata table.
```
alsepkow: nit: 

```suggestion
of the signature elements and emitting the load/store intrinsic. Signature IDs are assigned in declaration order (parameters, then return outputs), giving each semantic a stable ID that appears in both the intrinsic call and the metadata table.
```

alsepkow: Or something similar. Just to clean up the statement a little. I had to do a double take on it when reading.

proposals/NNNN-semantic-signatures-metadata-schema.md
```diff
@@ -0,0 +1,245 @@
...
+to the metadata nodes as described above. This is described in more detail
+below.
+
+### In-Memory Representation
+
+A `SemanticSignatureElement` struct will be defined in
+`llvm/Frontend/HLSL/SemanticSignatures.h` that will contain all the operands as
+above, as well as, helpers to determine the extra container field information
+(eg. is allocated). It will also define functions to convert to/from the
+metadata representation and the packing algorithm is defined here.
```
alsepkow: nit:
```suggestion
A `SemanticSignatureElement` struct, defined in
`llvm/Frontend/HLSL/SemanticSignatures.h`, will contain all the operands as
above as well as helpers to determine the extra container field information
(eg.,  is allocated). The struct will also define functions to convert to and from the
metadata representation, and the header will define the packing algorithm.
```

proposals/NNNN-semantic-signatures-metadata-schema.md
```diff
@@ -0,0 +1,245 @@
...
+
+### Accumulate `SemanticSignatureElement`s
+
+`emitEntryFunction` will then have a vector of `SemanticSignatureElement`s for
+both input and output signatures. The existing traversal visits each leaf
+semantic in declaration order as it generates `llvm.dx.load.input` /
```
alsepkow: Very minor nit, but I think its better to be super explicit and use 'and' , 'or', or whatever instead of slashes in something formal. The intent is very clear in this case as-is though.

alsepkow (commented):
LGTM. Just some nits on tightening some statements. Take them or leave them.

PR #434, opened 2026-06-17
[0048] Local Resources offload test suite updates

bob80905:
This PR updates some tests after an attempt to write all tests for the offload test suite.
It is meant to also prime discussion on a couple tests.
These changes are needed to be sensible: certain tests shouldn't have any ought runtime behavior, and should be updated to reflect as such.


proposals/0048-local-resources-behavior.md
```diff
@@ -130,7 +130,7 @@ miss the codegen-stage error that both compilers raise from
...
+| [`local_resource_struct_member_reassign_ambiguous`] | Conditional reassign of a struct's resource member (`h.buf = gBuf0; if (tid) h.buf = gBuf1;`) — `h.buf` may resolve to either global at the access site | Warning † at sema, Error at codegen if ambiguity survives optimization (binding-ambiguous reassignment) | Store target depends on `tid` (only one path possible per dispatch) | ❌ No sema warning (`-Whlsl-explicit-binding` doesn't fire across struct member access), then codegen error *"Resource access is not guaranteed to map to a unique global resource"* | ✅ Codegen error: *"local resource not guaranteed to map to unique global resource"* |
```
bob80905: I made an NFC here just to point to this test. The Ought-Compile description isn't ideal, and I would like to be able to state definitively whether an error ought to be expected here or not, without the "if ambiguity survives optimization" condition.
This will determine whether this test ought to be in the offload test suite or not.

proposals/0048-local-resources-behavior.md
```diff
@@ -249,7 +249,7 @@ the same binding are unambiguous and should compile clean.
...
 | [`local_resource_conditional_reassign_from_array`] | `if(cond) out = arr[i];` (conditional reassign from unbounded resource array) | Warning † (binding-ambiguous reassignment) | Store reaches either the original global or `arr[i]` per `cond` | ✅ Warning (`-Whlsl-explicit-binding`) | ❌ Codegen error: *"local resource not guaranteed to map to unique global resource"* |
-| [`local_resource_static_ternary_assign`] | Ternary assigning to a `static` local resource (lib target) | Warning † (binding-ambiguous reassignment) | Store reaches the selected global; the binding persists across calls | ✅ Warning (`-Whlsl-explicit-binding`) | ❌ Codegen errors: *"non const static global resource use is disallowed"* + *"local resource not guaranteed to map to unique global resource"* |
+| [`local_resource_static_ternary_assign`] | Ternary assigning to a `static` local resource (lib target) | Warning † (binding-ambiguous reassignment) | N/A (library target) | ✅ Warning (`-Whlsl-explicit-binding`) | ❌ Codegen errors: *"non const static global resource use is disallowed"* + *"local resource not guaranteed to map to unique global resource"* |
```
bogner: This doesn't make sense - why would being a library target make this N/A? OTOH, this test seems kind of irrelevant since `static` local resources aren't really allowed in a lib target anyway - dxc says "error: static global resource use is disallowed for library functions"

https://godbolt.org/z/1ze986no7

proposals/0048-local-resources-behavior.md
bogner: It's hard to come up with a case that doesn't either trivially constant fold (in which case there is no error) or emits the error in all cases. It'd be nice to come up with something that differs between optimized and -O0 to demonstrate what we meant here.

proposals/0048-local-resources-behavior.md
bob80905: Library shaders have no standalone execution model. They are library shaders, intended to be compiled along with other shaders for execution. 
So, given it shouldn't run by itself, we shouldn't impose any "Ought Runtime" statements on it.
Separately, strangely, the godbolt link does not report any dxc compilation errors (the one in the doc, not in your comment)
So I might have to update this, but I still think the N/A should stand.

bogner: Oh sorry I got mixed up - the columns are compile and runtime, not sema/codegen. In that case I agree N/A makes sense, but this shouldn't just be a warning - `static` local resources shouldn't compile at all in library profiles (and DXC does this correctly already).

The reason the godbolt link in the doc doesn't report an error is because it doesn't export any functions, so it doesn't generate any code at all. Since dxc emits this error in codegen, it can only fire if there's anything to codegen. IMO clang should be able to error on this in Sema, though I think this doc merely needs to specify that it does indeed error. Also, as I mentioned above, the ternary is completely irrelevant here, that static local resource itself is the problem.

proposals/0048-local-resources-behavior.md
bob80905: Since we have an unambiguous variant of this test, I think we should change this test to make it absolutely ambiguous, regardless of compiler optimization. This is because we will need to include this test in the offload test suite if it is sufficiently unambiguous, and pass a compiler optimization flag as well, which complicates things.
The point of this PR is just to get the doc into shape to accurately reflect offload test suite tests.
So, I'll make this truly ambiguous by upping numthreads to [4,1,1] instead, then the ought statement changes to a warning at sema and an error at codegen.
Then, we can be clear that this test does not (in any of its compiler optimization variations) belong in the offload test suite.

bogner (approved):

proposals/0048-local-resources-behavior.md
```diff
@@ -364,8 +365,8 @@ Helena Kotas, Justin Bogner
...
+[`local_resource_conditional_reassign_different_global`]: https://godbolt.org/clientstate/eyJzZXNzaW9ucyI6W3siaWQiOjEsImxhbmd1YWdlIjoiaGxzbCIsInNvdXJjZSI6IlJXQnVmZmVyPHVpbnQ%2BIEluIDogcmVnaXN0ZXIodTApO1xuUldTdHJ1Y3R1cmVkQnVmZmVyPHVpbnQ%2BIE91dDAgOiByZWdpc3Rlcih1MSk7XG5SV1N0cnVjdHVyZWRCdWZmZXI8dWludD4gT3V0MSA6IHJlZ2lzdGVyKHUyKTtcblxuY2J1ZmZlciBjIHtcbiAgICBib29sIGNvbmQ7XG59O1xuXG52b2lkIGJyYW5jaGVkX2Fzc2lnbm1lbnQodWludCBpZHgpIHtcbiAgICBSV1N0cnVjdHVyZWRCdWZmZXI8dWludD4gT3V0ID0gT3V0MDtcbiAgICBpZiAoY29uZCkge1xuICAgICAgICBPdXQgPSBPdXQxO1xuICAgIH1cbiAgICBPdXRbaWR4XSA9IEluW2lkeF07XG59XG5cbltudW10aHJlYWRzKDY0LCAxLCAxKV1cbnZvaWQgbWFpbih1aW50MyB0aWQgOiBTVl9EaXNwYXRjaFRocmVhZElEKSB7XG4gICAgYnJhbmNoZWRfYXNzaWdubWVudCh0aWQueCk7XG59XG4iLCJjb21waWxlcnMiOlt7ImlkIjoiZHhjX3RydW5rIiwib3B0aW9ucyI6Ii1UIGNzXzZfMyAtRSBtYWluIiwibGlicyI6W10sInRvb2xzIjpbXX0seyJpZCI6Imhsc2xfY2xhbmdfdHJ1bmsiLCJvcHRpb25zIjoiLVQgY3NfNl8zIC1FIG1haW4iLCJsaWJzIjpbXSwidG9vbHMiOltdfV19XX0%3D
```
alsepkow, line 368: Not a new pattern in this PR. But aren't these links going to become stale as things are implemented? Should they point to a specific version of the compiler instead? 

I'm not convinced that actually matters. But just stood out to me that one day pointing to trunk for hlsl clang on failing conditions will just work instead.

bob80905: IIUC, you're saying these links should be updated to point to specific compiler versions rather than trunk, so that behavior remains constant even as the compilers evolve.
I think it's better to keep it pointing to trunk, the benefits are that this keeps this document living. 
There are currently discrepancies between ought behavior and current behavior. With trunk, when we fix compiler behavior to match ought behavior, the spec will need to be updated to accurately reflect the current status.
Then, we will know to update the spec with the new matching current behavior and ought behavior.
Leaving it on an older version lends the doc to rotting, since the current compile/runtime behavior won't actually be up to date.

bob80905: A more concise way to put it: 
Yes, the spec can be stale / inaccurate if we point it to trunk.
But, if someone is looking at the godbolt link anyways, and notices a discrepancy, that would reveal to us the need to update the doc.
It is more valuable to know the current status of the compilers than a previous version. Otherwise we'd have to maintain and update all the links with the new compiler versions.

alsepkow: I suppose what stood out to me is tracking the status of work in a spec. I'm not necessarily opposed to that though.

bob80905: Yeah, I understand how that's somewhat odd.
The original issue intended to create some sort of "survey" document to describe the current state of the local resources work, with all the behaviors documented.
I think tracking the status is the most effective way to accomplish this, and it's a nice and organized way to see what is tested where.

alsepkow (approved):
lgtm

PR #446, opened 2026-07-01
[xxxx] Add a non power of 2 matrix legalization strategy for SPIRV

farzonl:
The existing matrix legalization strategy is very brittle This change reconsiders what to do and instead proposes a new largest common divisor way of splitting.

### DIFF
--- a/proposals/xxxx-spirv-matrix-legalization.md
+++ b/proposals/xxxx-spirv-matrix-legalization.md
@@ -8,7 +8,7 @@ params:
 
 ## Introduction
 
-HLSL matrix types flatten to wide `<N x T>` vectors (e.g. `bool3x4` → `<12 x i1>`),
+HLSL matrix types flatten to wide `<N x T>` vectors (e.g. `bool3x4` --> `<12 x i1>`),
 but SPIR-V shader/Vulkan targets cap vectors at 4 components. The GlobalISel
 legalizer must split wide loads, stores, extends, and shuffles into legal chunks.
 See https://github.com/llvm/llvm-project/issues/186864.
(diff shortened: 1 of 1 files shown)


bob80905:
⁠Does our decision here also have an effect on how we implement long vectors in clang?
If we have a long vector that's a prime, like 71, and it isn't divisible by 2 3 or 4, then would we switch to a different method?
Or are there any cases where we have large matrices that are prime in size that would fail with this approach?

farzonl:
> ⁠Does our decision here also have an effect on how we implement long vectors in clang? If we have a long vector that's a prime, like 71, and it isn't divisible by 2 3 or 4, then would we switch to a different method? Or are there any cases where we have large matrices that are prime in size that would fail with this approach?

It should have no impact on clang, this is entirely a backend feature. I suspect the spir-v backend will need to go through a similar change that the DXC backend went through so we leave things in vectorized forms that don't need to modify.

proposals/xxxx-spirv-matrix-legalization.md
```diff
@@ -0,0 +1,111 @@
...
+  authors:
+    - farzonl: Farzon Lotfi
+---
+
+## Introduction
+
+HLSL matrix types flatten to wide `<N x T>` vectors (e.g. `bool3x4` --> `<12 x i1>`),
+but SPIR-V shader/Vulkan targets cap vectors at 4 components. The GlobalISel
+legalizer must split wide loads, stores, extends, and shuffles into legal chunks.
+See https://github.com/llvm/llvm-project/issues/186864.
```
farzonl, line 14: To make it easier  to click on this issue: https://github.com/llvm/llvm-project/issues/186864

bogner (approved):

damyanp:
Don't forget to assign a proposal number before merging this PR!

bob80905 (approved):

farzonl:
I think we table this for now and revisit later.

Issue #203, opened 2025-04-01
Document DXC handling of dynamic resources and add tests(s) to offload-test-suite


hekota:
Document how DXC handles dynamic resources.
Create a set of test cases and add them to the offload-test-suite. 

https://microsoft.github.io/DirectX-Specs/d3d/HLSL_SM_6_6_DynamicResources.html

damyanp:
Some more investigation into the size of this feature required before we can estimate it.

PR #460, opened 2026-08-13
Design proposal for dynamic resources

### LINKED ISSUE #204: Dynamic resources design
Create a proposal how to implement dynamic resource in Clang.

https://microsoft.github.io/DirectX-Specs/d3d/HLSL_SM_6_6_DynamicResources.html

Related comment in proposal for resource constructors: https://github.com/llvm/wg-hlsl/pull/197#discussion_r2047684735

Depends on: llvm/wg-hlsl#203

hekota:
Initial design proposal for dynamic resources in Clang.

Fixes #204
Part of #203

Assisted by Clause Opus 5 and GPT-5.6 Sol.

### DIFF
--- a/proposals/NNNN-dynamic-resources.md
+++ b/proposals/NNNN-dynamic-resources.md
@@ -25,7 +25,7 @@ signature descriptor-table mappings.
 
 ## Motivation
 
-Dynamic resources are part of the HLSL language and we need to support it in
+Dynamic resources are part of the HLSL language and we need to support them in
 Clang.
 
 ## Current Behavior in DXC
@@ -34,7 +34,7 @@ This section documents how DXC handles dynamic resources.
 
 ### Example 1 - Assigning to local resource variable
 
-Resource instance aquired by direct indexing is usually assigned to a local
+Resource instance acquired by direct indexing is usually assigned to a local
 resource variable:
 
 ```cpp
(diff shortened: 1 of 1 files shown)


proposals/NNNN-dynamic-resources.md
```diff
@@ -0,0 +1,387 @@
...
+void main() {
+  foo(ResourceDescriptorHeap[Index]);
+}
+```
+https://godbolt.org/z/EjaK4rxhe
+
+### Example 3 - Using directly indexed heap as a resource class
+
+The result of the direct indexing into `ResourceDescriptorHeap` or
+`SamplerDescriptorHeap` cannot be directly used as if it was a resource class:
```
damyanp: I was wondering if this was a limitation we wanted to carry forward, but I think that this is because there's no way to know the type of the resource in these cases.  Is that correct?  The way it's worded here it sounds, on a superficial reading, like an arbitrary restriction in DXC we might want to fix.

damyanp (approved):
Looks great for merging for further discussion.

proposals/NNNN-dynamic-resources.md
hekota: Right, there is no way to know the type of resource that is getting indexed. I will add a note to explain that.

bogner (approved):

PR #464, opened 2026-08-13
[0049] - Resource Instance Property Annotations

### LINKED ISSUE #452: [NNNN] Specify how `globallycoherent` will be retained through llvm ir
This issue tracks the resolution of the TODO note in the [resource docs](https://llvm.org/docs/DirectX/DXILResources.html#resource-type-information-and-properties):

> TODO: There are two fields in the DXIL metadata that are not represented as part of the target type: IsGloballyCoherent and HasCounter.

> Since these are derived from analysis, storing them on the type would mean we need to change the type during the compiler pipeline. That just isn’t practical. It isn’t entirely clear to me that we need to serialize this info into the IR during the compiler pipeline anyway - we can probably get away with an analysis pass that can calculate the information when we need it.

> If analysis is insufficient we’ll need something akin to annotateHandle (but limited to these two properties) or to encode these in the handle creation.

AC:

- [ ] A proposal is accepted of how the todo should be resolved
- [ ] The llvm docs are updated with the new proposal

inbelic:
Resolves https://github.com/llvm/wg-hlsl/issues/452

### DIFF
--- a/proposals/NNNN-resource-property-annotation.md
+++ b/proposals/NNNN-resource-property-annotation.md
@@ -1,0 +1,210 @@
+---
+title: "[NNNN] - Resource Property Annotation"
+params:
+  status: Under Consideration
+  authors:
+    - inbelic: Finn Plummer
+  sponsors:
+    - inbelic: Finn Plummer
+---
+
+* Related proposals: [0006], [0015], [0018], [0022], [0023]
+
+## Introduction
+
+This proposal resolves how the non-structural resource instance properties
+`HasCounter`, `IsGloballyCoherent`, and `IsReorderCoherent` will be
+represented. `HasCounter` is observable from the counter operations already
+present in IR and so it does not require any additional representation. Global
+and reorder coherence express source-level semantics that are not observable
+from the resource operations and so they will be preserved with an explicit
+resource coherence annotation. The per-target lowering of each property is
+described in the sections below.
+
+## Has Counter
+
+Use of a resource counter is able to be retained in IR. Every increment or
+decrement is represented by an `llvm.dx.resource.updatecounter` operation whose
+handle operand identifies the resource instances that require a counter, and
+whose increment operand records the direction of the update.
+
+When lowering to DirectX, resource analysis will visit these operations before
+they are lowered to `dx.op.bufferUpdateCounter` and trace each handle to the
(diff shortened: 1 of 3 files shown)


proposals/NNNN-resource-property-annotation.md
```diff
@@ -0,0 +1,210 @@
+---
+title: "[NNNN] - Resource Property Annotation"
```
inbelic:
```suggestion
title: "[NNNN] - Resource Instance Property Annotations"
```

proposals/NNNN-resource-property-annotation.md
```diff
@@ -0,0 +1,210 @@
...
+
+The most direct representation would be for the loads and stores through a
+coherent resource to carry the memory semantics themselves, which would make
+the property inferable rather than annotated. As described in the Global
+Coherence section, LLVM has does not have such semantics available on `load`
+and `store`, and memory barriers apply to a memory space rather than to a
+specific resource.
+
+This would be possible but would require additions to LLVM's memory semantics,
+which is a substanial change.
```
inbelic: Looking for feedback if this is true? Would adding memory semantics be a possibility?

proposals/NNNN-resource-property-annotation.md
bogner: Should update the filename as well.

proposals/NNNN-resource-property-annotation.md
bogner: I think this option could really be split into two ideas, with fairly different trade offs in how heavyweight they are.

1. Adding semantics to loads/stores. This would probably look like creating some new metadata to attach to loads and stores to describe coherence. This is probably pretty substantial given that loads and stores obviously show up all over the place and we'd need to make sure we were handling the new attribute and interactions with existing attributes correctly.

2. Using barriers. For this, I think we'd probably want to introduce a new type of barrier that's resource specific. This could then use the resource as an argument and whatnot. I guess in practice this looks a lot like the annotations you're proposing here though - we'd still need target specific handling, the only difference would be that the placement of the barrier instruction itself would carry more specific meaning.

bogner (approved):
This is in good shape to get a proposal number and merge. Next, we should discuss this in a design meeting.

PR #468, opened 2026-08-17
Finalize local resources doc

### LINKED ISSUE #458: Synchronize local-resources-behavior.md with LLVM and Offload tests.
After https://github.com/llvm/llvm-project/pull/190037 lands, all the tests for local resources behavior should be in.
This issue tracks making sure any changes that needed to be made to the set of tests in the Offload test suite or the LLVM Project, (i.e., test locations, or test removals/additions, or any other relevant changes), are reflected by:
https://github.com/llvm/wg-hlsl/blob/main/proposals/0048-local-resources-behavior.md.

bob80905:
This PR updates the local resources survey doc so that it matches the current state of the tests in both the offload test suite and the llvm project.
Many tests have been removed / consolidated since the last update, so it is important that this doc accurately reflects the current state of testing coverage.

Fixes: https://github.com/llvm/wg-hlsl/issues/458


proposals/0048-local-resources-behavior.md
```diff
@@ -378,107 +259,20 @@ time a compiler changes. The Ought-driven placement described in
...
...ZlciBCdWYpIHtcbiAgICBCdWYgPSBHQnVmMDtcbn1cblxuW251bXRocmVhZHMoMSwxLDEpXVxudm9pZCBtYWluKHVpbnQgR0kgOiBTVl9Hcm91cEluZGV4KSB7XG5cbiAgUmVzSG9sZGVyIEggPSB7R2V0QnVmKCksIDB9O1xuICBSV0J5dGVBZGRyZXNzQnVmZmVyIEJ1ZiA9ICh0cnVlID8gSC5CdWYgOiBHQnVmMSk7XG4gIGlmIChmYWxzZSlcbiAgICBCdWYgPSBHQnVmMTtcbiAgaWYgKHRydWUpXG4gICAgQnVmID0gQnVmO1xuICBSV0J5dGVBZGRyZXNzQnVmZmVyIEFsaWFzID0gQnVmO1xuICBSV0J5dGVBZGRyZXNzQnVmZmVyIENvbW1hID0gKEdCdWYxLCBBbGlhcyk7XG5cbiAgRm9yd2FyZFN0b3JlKENvbW1hLCAwKTtcblxuICBHZXRCdWYoKS5TdG9yZSg0LCA0Mik7XG5cbiAgUldCeXRlQWRkcmVzc0J1ZmZlciBGcm9tUmV0ID0gR2V0QnVmKCk7XG4gIEZyb21SZXQuU3RvcmUoOCwgNDIpO1xuXG4gIFJXQnl0ZUFkZHJlc3NCdWZmZXIgQm90aFNhbWUgPSAoR0kgIT0gMCkgPyBHQnVmMCA6IEdCdWYwO1xuICBCb3RoU2FtZS5TdG9yZSgxMiwgNDIpO1xuXG4gIFJXQnl0ZUFkZHJlc3NCdWZmZXIgRnJvbU91dDtcbiAgV3JpdGVUaHJvdWdoKEZyb21PdXQpO1xuICBGcm9tT3V0LlN0b3JlKDE2LCA0Mik7XG5cbiAgUmVzSG9sZGVyIE1peGVkO1xuICBNaXhlZC5CdWYgPSBHQnVmMDtcbiAgTWl4ZWQuVmFsdWUgPSA0MjtcbiAgTWl4ZWQuQnVmLlN0b3JlKDIwLCBNaXhlZC5WYWx1ZSk7XG59XG4iLCJjb21waWxlcnMiOlt7ImlkIjoiZHhjX3RydW5rIiwib3B0aW9ucyI6Ii1UIGNzXzZfMyAtRSBtYWluIiwibGlicyI6W10sInRvb2xzIjpbXX0seyJpZCI6Imhsc2xfY2xhbmdfdHJ1bmsiLCJvcHRpb25zIjoiLVQgY3NfNl8zIC1FIG1haW4iLCJsaWJzIjpbXSwidG9vbHMiOltdfV19XX0%3D
```
inbelic, line 317: Is using the long form really required?

proposals/0048-local-resources-behavior.md
```diff
@@ -39,15 +39,17 @@ This proposal aims to:
...
+   current behavior is wrong points back to a tracking compiler bug or
+   spec issue.
+
+> **Note:** Two pre-existing multi-function test files landed via PR
+> #182101 (`local_resource_bindings.hlsl` and
+> `local_resource_bindings_errs.hlsl`) were moved from directly under the
+> SemaHLSL directory into `Resources/Local-Resources/` and renamed to
+> `bindings.hlsl` and `bindings_errs.hlsl`. They are retained as
+> multi-function files rather than split per-pattern, so their original
+> cases stay together.
```
inbelic:
```suggestion
```

proposals/0048-local-resources-behavior.md
```diff
@@ -84,211 +86,86 @@ recorded in the table. This matters for binding-ambiguous patterns: a
...
-| [`local_resource_const_param_store`] | Function takes `const RWByteAddressBuffer buf` and calls `buf.Store(...)` inside the callee body | Clean (same rationale as above — `const` is on the handle, not the buffer) | Store reaches the global passed in | ❌ Error: *"no matching member function for call to 'Store'"* (diagnostic fires at the callee `Store` call site) | ✅ Clean |
+| [`consolidated_assignments`] | Chained aliasing/initialization forms — ternary initializer (constant and same-arm runtime conditions), struct aggregate init from a function return, direct init from a function return, a method called directly on a function return, alias chain, comma initializer, self-assign, dead-branch reassignment, and forwarding through a call chain — all resolving to one global | Clean | Every Store reaches the single global (`gBuf0`); `gBuf1`'s binding is folded away | ✅ | ✅ |
```
inbelic: I am not sure the usefulness of the last two columns now. It doesn't seem like any of the cases have a runtime behaviour difference, so this should be visible from looking at the godbolt link, and then the behaviour will be updated automatically instead of these going stale.

Imo, it might also be more legible to move the behaviour columns into the godbolt links so we can annotate inline etc.

proposals/0048-local-resources-behavior.md
bob80905: It embeds the HLSL source into the URL, so that it is impossible for the link to not work if the godbolt site decides to expire the link or anything. I think it is the best option if we want a doc that is relevant in the long run.

inbelic: I see, all the other proposals use the short form, so maybe those should be updated then?

proposals/0048-local-resources-behavior.md
bob80905: I agree that runtime behavior doesn't need to be described for tests that are only being run in the llvm repo, and not on the offload test suite, so I've reorganized it so the column doesn't exist for non-offload-test-suite tests.
But the purpose of the doc is to drive progress in fixing these issues. I don't think we should expect people to need to click on all the links and see what godbolt says in terms of compilation results and manually match it to the ought column to decide whether or not there is work to do (correct me if that's a strawman). We already have many successes / matches in terms of actual and expected behavior, so I think it would save time and make the doc more useful to store this info, even if it can become stale.
Another thing that helps is that regression is unlikely, so we only need to investigate the mismatches. Removing those columns would remove our knowledge of what is mismatching.

proposals/0048-local-resources-behavior.md
bob80905: It would be nice though I acknowledge that it's a lot of work. But yeah I think that is the better longterm solution.

proposals/0048-local-resources-behavior.md
inbelic: Okay, I won't block on this just a suggestion

proposals/0048-local-resources-behavior.md
inbelic: I see, although maybe this suggests it will be okay to continue to use short: https://github.com/compiler-explorer/compiler-explorer/discussions/4942#discussioncomment-5525574

inbelic (approved):
I do stand by my suggestions but won't block on them

proposals/0048-local-resources-behavior.md
bogner: I don't think we need to bother changing this document, but the short links are intended to be good forever (see [the FAQ](https://github.com/compiler-explorer/compiler-explorer/wiki/FAQ)). The only real reason to bother with long links is if you want them to persist across instances of compiler explorer or if it's important that the links somehow outlast godbolt.org

bogner (approved):

